PR #97, opened 2013-07-18
Add support for WGL_NV_gpu_affinity

bretjackson:
This patch allows users to choose which GPU the opengl rendering commands are sent to using the wgl_nv_affinity extension. The gpu is specified through a new 'GLFW_AFFINITY_GPU' hint. If the extension is unsupported on the current hardware, the hint is just ignored. If the hint is not set, a normal device context is created as usual.
For systems with multiple GPUs, using multiple windows this can greatly speed up rendering efficiency.

It also adds a new Native API call to get the device context.

### DIFF
--- a/docs/window.dox
+++ b/docs/window.dox
@@ -188,6 +188,11 @@ used by the context.  This can be one of `GLFW_NO_RESET_NOTIFICATION` or
 `GLFW_LOSE_CONTEXT_ON_RESET`, or `GLFW_NO_ROBUSTNESS` to not request
 a robustness strategy.
 
+The `GLFW_AFFINITY_GPU` hint specifies which GPU should render OpenGL commands
+for the associated context. This uses the WGL_NV_gpu_affinity extension, and is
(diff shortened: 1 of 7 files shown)


unbored:
It helps me a lot! Thanks. But in the newest version, error occurred when destroying the window. The window is created twice. When the window is created the second time, reallocation of the affinity functions leads to null pointers, so that destruction of the affinity DC failed. So I added a judgement for the null pointers.
For AMD cards, it seems that the commands are dispatched to the GPU where the window initialized. However, glfwCreateWindow command doesn't have parameters for the window position, meaning that the window is initialized in the primary monitor, where only the primary GPU is chosen. glfwSetWindowPos won't change the GPU that commands are dispatched.

zx-li001:
@unbored so you use the old version or find a better way to resolve the errors in newest version?

cxw42:
@unbored I'm also interested.  Would you be willing to share your final code?  Thanks!

PR #98, opened 2013-07-19
Retrieve joystick names from registry

salvy:


elmindreda:
This won't be necessary once DirectInput is used.

Issue #78, opened 2013-07-01
Joystick functions return zero buttons/axes with various controllers


tgoulart:
After calls to `glfwGetJoystickButtons` and `glfwGetJoystickAxes`, `count` is always 0. I have the following code, which I believe is correct:

``` c++
for (int i=GLFW_JOYSTICK_1; i <= GLFW_JOYSTICK_LAST; ++i) {
    if (glfwJoystickPresent(i) == GL_TRUE) {
        int nbuttons, naxes;
        const unsigned char *pressed = glfwGetJoystickButtons(i, &nbuttons);
        const float *axes = glfwGetJoystickAxes(i, &naxes);
        printf("%d %d -- %p %p\n", nbuttons, naxes, pressed, axes);
    }
}
```

It outputs `0 0 -- 0x100c239f0 0x100c264c0`, and this tells me that, even though GLFW didn't catch any errors, something weird is happening internally. I'm not familiar with the HID API, but I can debug this further if you need more information.

Hopefully, this isn't just me being stupid.

System: OS X 10.8.2, AMD Radeon HD 6770M, Xbox 360 wired controller, latest GLFW 3.0.1 from github as of now.

Issue #146, opened 2013-09-27
Some joystick axes are inverted


elmindreda:
[Discussion here](https://sourceforge.net/p/glfw/discussion/247562/thread/c80f3707/).

PR #164, opened 2013-10-27
Convert joystick code to HID interface

proc-addr:
tested with playstation sixaxis controller. should fix issue #78.
also remove y axis inversion (issue #146).


elmindreda:
This has been replaced by #536.

PR #168, opened 2013-11-04
Made glfwConfig.cmake export GLFW_LIBRARIES

barcharcraz:
glfwConfig.cmake.in only created GLFW_LIBRARY, not GLFW_LIBRARIES. This made it hard to use glfw in config mode since you would have had to manually link all its dependencies (x11 and so on).


shaxbee:
#460 submitted by me and @ngvitovitch covers that - we added target_link_libraries and target_include_directories directives and proper glfw export.

elmindreda:
This has been replaced by #460.

PR #192, opened 2013-12-13
Replace CGDisplayIOServicePort with a workaround implementation

mhenr18:
This could be cleaned up and refined a bit by comparing more display
attributes when available, but it gets the job done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ src/glfw_config.h
 src/glfw3.pc
 src/glfwConfig.cmake
 src/glfwConfigVersion.cmake
+install_manifest.txt
 
 # Compiled binaries
 src/libglfw.so
--- a/include/GLFW/glfw3native.h
+++ b/include/GLFW/glfw3native.h
@@ -55,7 +55,8 @@ extern "C" {
  *
  *  The available context API macros are:
  *  * `GLFW_EXPOSE_NATIVE_WGL`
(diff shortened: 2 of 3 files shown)


jaredjones:
Lion is also now unsupported even for security updates from Apple.

elmindreda:
Steam still supports 10.6 and the amount of effort required for GLFW to do so is still just a few isolated ifdefs. If it becomes an issue or when Steam removes support I'll drop those lines. Support for 10.6 isn't holding the code back, as far as I can tell. The code uses several of 10.7 features and at least one symbol added in 10.10.

Supporting Windows XP is far more, far nastier work.

elmindreda:
So, #448 might be the first good argument for dropping 10.6 support.

jaredjones:
I understand the desire to want to keep Snow Leopard support, but no one should be using that OS anymore. It's vulnerable to Heartbleed and about 50 other major vulnerabilities that have been patched on OS X since it was supported. World of Warcraft: The largest MMO on the planet doesn't even support Snow Leopard. :)

elmindreda:
I don't have any particular emotional attachment to any version after 10.3; it's just that the effort involved to keep supporting a version still supported by Steam used to be tiny and thus worth it.

jaredjones:
Why don't we just check the operating system and if it's Snow Leopard it can execute the legacy code, however if it's running a newer version it can run this new implementation?

PR #342, opened 2014-08-28
Set window Background to black

ricardomv:
The background was being set to transparent, for example on wayland you couldn't see the window if only the background was set with an alpha value of 0 . This patch sets the clear color of tests and examples to black (and some to a transparent black). It also removes the first frame that was being drawn when a window is created avoiding a initial blink.

Ricardo

### DIFF
--- a/examples/boing.c
+++ b/examples/boing.c
@@ -201,7 +201,7 @@ void init( void )
    /*
     * Clear background.
     */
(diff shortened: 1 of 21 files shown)


ricardomv:
This was fixed in #757

elmindreda:
Thank you for the follow-up!

PR #366, opened 2014-10-14
Upstream biicode

drodri:
This PR modifies (non-intrusively) the CMakeLists in order to use the www.biicode.com dependency manager.
It also adds a couple of files under the "bii" subfolder.
Included a .travis.yml proposal for initially handling travis-ci builds.

Please check:
- Biicode block: http://www.biicode.com/diego/glfw
- Biicode documentation: http://docs.biicode.com/c++/examples/glfw.html
- Github fork: https://github.com/drodri/glfw
- Travis build: https://travis-ci.org/drodri/glfw

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,5 @@ tests/title
 tests/version
 tests/windows
 
+bii
+bin
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -545,6 +545,11 @@ configure_file("${GLFW_SOURCE_DIR}/src/glfw3.pc.in"
 #--------------------------------------------------------------------
 add_subdirectory(src)
 
+IF(BIICODE)
+    INCLUDE(biicode.cmake)
+    return ()
+ENDIF()
+
 if (GLFW_BUILD_EXAMPLES)
     add_subdirectory(examples)
 endif()
@@ -583,4 +588,3 @@ if (GLFW_INSTALL)
                           "${GLFW_BINARY_DIR}/cmake_uninstall.cmake")
     endif()
 endif()
-
--- a/README.md
+++ b/README.md
@@ -34,6 +34,13 @@ See the
 guide in the GLFW documentation.
 
 
+### Using GLFW with biicode
+
+GLFW is available at [biicode](http://www.biicode.com) C/C++ dependency
+manager, you can find the [block here](http://www.biicode.com/diego/glfw), and
+[examples of use here](http://docs.biicode.com/c++/examples/glfw.html)
+
+
 ## Reporting bugs
 
 Bugs are reported to our [issue tracker](https://github.com/glfw/glfw/issues).
--- a/biicode.cmake
+++ b/biicode.cmake
@@ -1,0 +1,26 @@
+INIT_BIICODE_BLOCK()
+
+#Biicode defines its own targets. Include src CMakeLists to load variables
+#with the adequate source files depending on the system
+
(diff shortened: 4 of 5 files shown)


drodri:
Hi Camilla,

I have updated the PR to account for biicode 2.0. As you can see, now there is only a single "biicode.conf" configuration file instead of many small ".bii" configuration files. Also many other improvements have been done. Biicode also keeps growing towards its goal of becoming open source soon.

I have also added Appveyor Continuous Integration with an appveyor.yml file, which is equivalent to TravisCI, but for Windows. Currently, it builds GLFW with MinGW, but a build matrix can be done and builds with different flavours of Visual Studio can be also automatically done. Also, similar to travis, it allows to check builds on PullRequest, so you can now in advance if some PR correctly builds on those platforms or not. I think they could be a great enhancement for the GLFW project, and I would love to help in the process of integrating them.

Best

drodri:
I have refactored the pull request:
- Removed CI scripts for appveyor and travis. Lets keep 1 PR per feature. I would love to add support for them, but I will do it in new PRs.
- Isolated required cmake configuration in independent "biicode.cmake" file, thus it is only necessary to add 4 lines to the main CMakeLists.txt, and it is not necessary to touch src/CMakeLists.txt
- Removed (as requested) deps/* files from the library. In fact, now biicode uses the full src/CMakeLists.txt as-is, and just wraps the library "glfw". This deps/* files are wrapped in their own library "glfw_deps" only if included by the final consumers. Please note that this is one of the advantage of using biicode, consumers can #include header files (also those in deps) and build their examples very easily, even if copy&pasting the examples within the library, without requiring anything else:
http://www.biicode.com/examples/glfw http://docs.biicode.com/c++/examples/glfw.html

elmindreda:
Thank you. This is clean enough to consider. Doing that now.

drodri:
Hi Camilla,

Recent improvements in biicode (from version 2.6.3) allows much simpler project layouts, testing, etc. So in order to make my repo work, now you could just do:
```
$ git clone https://github.com/drodri/glfw.git
$ cd glfw
$ git checkout upstream_biicode
```
now, initialize biicode project, with new simple layout
```
$ bii init -L 
$ bii build
$ bin\.... //launch the examples you want
```
New "bii test" command will build and run all the test:
```
$ bii test
```
Once everything works, simply publish:
 ```
$ bii publish
```
That branch will publish to my user (if allowed). To publish to another user, just modify the [parent] sections of biicode.conf, with your username, and version -1, or for the initial submission, it could be removed, it will be automatically created from your username:
```
$ bii user myusername
$ bii publish
```
Please let me know any doubt or any further issue.

andrewrk:
This code belongs in a fork or a packaging repository, not the upstream project.

drodri:
Yes, probably, though it was more complex to do so without duplicating much of the existing CMakeLists.txt in the other repository (if a separate packaging-only repo) and that would be more difficult to mantain. A fork could be easier to maintain, but I wanted to contribute to the upstream project with travisCI and appveyor runs, that builds and test the project (even automatically for PR, so you know in advance if submissions are OK). That builds can also automatically publish to biicode after succesful builds & tests, so the whole process is automated, and that can be done if it is in the upstream repository.

Nevertheless, we are working also in an external package-only approach that could also be used without affecting the upstream project, so I will let you know when it is ready, if you want to check it. Thanks!

shibukawa:
http://blog.biicode.com/biicode-just-the-company-post-mortem/index.html

PR #377, opened 2014-11-01
Support for automatic graphics switching

jonathanhardie:
Enables the context to participate in automatic graphics card switching.
This means that GLFW won't force switching to the discrete graphics 
card on any Mac that has dual graphics cards available.  
This will result in lower power consumption for GUI-style apps where
the integrated graphics are more than adequate, but may be ignored
for heavier applications such as games, where the user most likely
wants to take advantage of the more powerful discrete graphics chip.  

For this functionality to actually be active, the Apps' Info.plist
file still needs to be edited to contain: 
<key>NSSupportsAutomaticGraphicsSwitching</key><true/>
otherwise it continues with default behaviour of forced discrete gfx.

Note that switching graphics chipsets also incurs a ~200ms penalty on 
app startup, which is avoided by having this attribute set.
Further info: 
https://developer.apple.com/library/mac/technotes/tn2229/_index.html

### DIFF
--- a/src/nsgl_context.m
+++ b/src/nsgl_context.m
@@ -125,6 +125,25 @@ int _glfwCreateContext(_GLFWwindow* window,
     NSOpenGLPixelFormatAttribute attributes[40];
 
     ADD_ATTR(NSOpenGLPFAClosestPolicy);
+    
+    // Enables the context to participate in automatic graphics card switching.
+    // This means that GLFW won't force switching to the discrete graphics 
+    // card on any Mac that has dual graphics cards available.  
+    // This will result in lower power consumption for GUI-style apps where
+    // the integrated graphics are more than adequate, but may be ignored
+    // for heavier applications such as games, where the user most likely
+    // wants to take advantage of the more powerful discrete graphics chip.  
+    
+    // For this functionality to actually be active, the Apps' Info.plist
+    // file still needs to be edited to contain: 
+    // <key>NSSupportsAutomaticGraphicsSwitching</key><true/>
+    // otherwise it continues with default behaviour of forced discrete gfx.
+	
+    // Note that switching graphics chipsets also incurs a ~200ms penalty on 
+    // app startup, which is avoided by having this attribute set.
+    // Further info: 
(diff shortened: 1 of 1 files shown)


jaredjones:
There isn't a way to programmatically trigger this. Apple does all of this via a Plist. I recommend that this is merged into master. :)

shurcooL:
In that case, merging this seems like the best way to go.

elmindreda:
My only concern with this is that the TN seems to imply that virtual screen (kinda sorta the GPU) can change at runtime *after* context creation if you opt into this.

elmindreda:
Parking this here for easy access.

[Technical Q&A QA1734 - Allowing OpenGL applications to utilize the integrated GPU](https://developer.apple.com/library/mac/qa/qa1734/)

MaraviRasmussen:
Hey guys , 

Anyone can explain step by step ho to achieve this ? i have no experience modifying info.plist , and really want to achieve this , i want to use steam but it seems open gl context still depends on my discreet gpu,

will wait for help, thanks!

kkaefer:
There is a way to control this behavior: `NSOpenGL` is backed by `CGL` on macOS. Instead of adding `NSOpenGLPFAAllowOfflineRenderers`, and adding the plist setting, you can also add `kCGLPFASupportsAutomaticGraphicsSwitching` to the `NSOpenGLPixelFormatAttribute` array. Of course, once Apple decides to change the underlying implementation, this might fail in spectacular ways.

However, there seems to be an officially supported API to accomplish this: [`[NSOpenGLPixelFormat initWithCGLPixelFormatObj:]`](https://developer.apple.com/reference/appkit/nsopenglpixelformat/1436129-initwithcglpixelformatobj?language=objc) takes a `CGLPixelFormatObj`. So we could move the initialization of the pixel format object from `NS*` to `CGL*` (it supports the same enums, just with a different prefix).

kkaefer:
Added a [PR](https://github.com/glfw/glfw/pull/935) that adds a `GLFW_ALLOW_GRAPHICS_SWITCHING` flag that enables graphics switching for a window/framebuffer without having to modify the Info.plist file.

PR #379, opened 2014-11-04
Make glfwPostEmptyEvent thread-safe

pcwalton:
The old code was not using `XSendEvent()` in a thread-safe manner, which could
cause messages to be corrupted. The thread-safe way to use `XSendEvent()` from
multiple threads would be to use `XLockDisplay()` and `XUnlockDisplay()` around
every call to Xlib. However, this would add unnecessary overhead to all X
communication. To avoid that, this patch changes the synchronization mechanism
of `glfwPostEmptyEvent()` on X11 from an X event to a send on a Unix pipe.
Because `glfwWaitEvents()` was already using `select()` for blocking, this
turned out to fit in quite naturally with the existing design and eliminates
the thread safety issues.

### DIFF
--- a/src/x11_init.c
+++ b/src/x11_init.c
@@ -29,10 +29,12 @@
 
 #include <X11/Xresource.h>
 
+#include <fcntl.h>
 #include <stdlib.h>
 #include <string.h>
 #include <limits.h>
 #include <stdio.h>
+#include <unistd.h>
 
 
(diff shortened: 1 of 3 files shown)


elmindreda:
`XSendEvent` [locks the display](https://github.com/mirror/libX11/blob/master/src/SendEvent.c#L57) internally. Closing this.

PR #380, opened 2014-11-06
Experimental Mir support backend

BrandonSchaefer:
This adds the option for GLFW_USE_MIR. Which will then use the Mir backend. There are still a few functions that Mir does not support. When those land new patches will come!

Very few things needed to change outside of the backend it self to land Mir support thanks to jadahl. Whats not supported: Clipboard, cursors, WM functions (such as show/hide, set title, set window pos, etc), Gamma. Everything else should be supported! (Unless I missed something in the unsupported list!).

To test, it'll be easier if you're on the laster ubuntu version (14.10). Though 14.04 should work here as well.

You'll want mir-demos from main. (sudo apt-get install mir-demos)
Move to a tty(1)
sudo mir_demo_server_shell

Move to a different tty(2)
sudo chmod 777 /tmp/mir_server
export MIR_SOCKET=/tmp/mir_server

To test the server is working:
mir_demo_client_eglplasma (q to quit)

From here, you can compile glfw with -DGLFW_USE_MIR=1 and test out the examples!

If you want to build mir from source, just send me an email and I can help with that!

### DIFF
--- a/src/mir_window.c
+++ b/src/mir_window.c
@@ -489,7 +489,9 @@ void _glfwPlatformPostEmptyEvent(void)
 
 void _glfwPlatformGetFramebufferSize(_GLFWwindow* window, int* width, int* height)
 {
+  if (width)
     *width  = window->mir.width;
+  if (height)
     *height = window->mir.height;
 }
 


src/mir_window.c
```diff
@@ -0,0 +1,542 @@
...
+
+static void handleKeyEvent(MirKeyEvent const key, _GLFWwindow* window)
+{
+    int pressed = key.action == mir_key_action_up ?  GLFW_RELEASE : GLFW_PRESS;
+    int mods    = mirModToGLFWMod(key.modifiers);
+
+    long text = _glfwKeySym2Unicode(key.key_code);
+    int plain = !(mods & (GLFW_MOD_CONTROL | GLFW_MOD_ALT));
+
+    _glfwInputKey(window, toGLFWKeyCode(key.scan_code), key.scan_code, pressed, mods);
```
BrandonSchaefer: I need to double check this... as I thought i checked the toGLFWKeyCode was working with the scan_code (with out the extra + 8). Maybe i just need to add it on to the second key.scan_code used?

PR #437, opened 2015-02-02
Joystick test drawing bug

bornskilled200:
joysticks controls was not being drawn visibly


elmindreda:
That is indeed a bug. Thank you!

PR #445, opened 2015-02-18
Added a few tips to the GLFW building and compiling documentation.

siavashserver:
These might save a few minutes of googling for newcomers :)


elmindreda:
Very good point.

PR #467, opened 2015-03-15
Add glfwSetWindowIcons

### LINKED ISSUE #453: Window icon support
I do hope I am not recreating this issue but I did a search and cannot find another.

It is mentioned on the website documentation however that x11 currently supports no window icon, I would however like to keep a track of the status of this to see its progression.

Is it possible to set the window icon on X11 some how? I see that fluxbox may not be supported.

Usefull link:
http://stackoverflow.com/questions/9634451/how-can-i-change-the-taskbar-icon-using-xiconifywindow

linkmauve:
There is currently only an implementation for the X11 backend, all of the other ones are stubbed, as I won’t be able to test a Mir, Win32 or Cocoa implementation.  The Wayland protocol currently doesn’t support that feature, so I added a comment describing that.

How do you usually deal with such PRs, do you merge them and then wait for other PRs to implement the same feature on other backends, or do you stall the PR until people have provided the missing implementations?

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1844,6 +1844,26 @@ GLFWAPI void glfwGetWindowSize(GLFWwindow* window, int* width, int* height);
  */
 GLFWAPI void glfwSetWindowSize(GLFWwindow* window, int width, int height);
 
+/*! @brief Sets the icons representing the specified window.
+ *
+ *  This function sets the icons, as an array of GLFWimage, of the specified
+ *  window.
+ *
+ *  If count is 0, it removes any icon previously set that way.
+ *
+ *  @param[in] window The window whose icons to change.
+ *  @param[in] images Array with the icons to set.
+ *  @param[in] count Number of images in the array.
+ *
+ *  @par Thread Safety
+ *  TODO
+ *
+ *  @since Added in GLFW TODO.
+ *
+ *  @ingroup window
+ */
+GLFWAPI void glfwSetWindowIcons(GLFWwindow* window, const GLFWimage* images, int count);
+
 /*! @brief Retrieves the size of the framebuffer of the specified window.
  *
  *  This function retrieves the size, in pixels, of the framebuffer of the
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -996,6 +996,13 @@ void _glfwPlatformSetWindowSize(_GLFWwindow* window, int width, int height)
         [window->ns.object setContentSize:NSMakeSize(width, height)];
 }
 
+void _glfwPlatformSetWindowIcons(_GLFWwindow* window, const GLFWimage* images, int count)
+{
+    // TODO
+    _glfwInputError(GLFW_PLATFORM_ERROR,
+                    "Cocoa: Set window icons not implemented yet");
+}
+
 void _glfwPlatformGetFramebufferSize(_GLFWwindow* window, int* width, int* height)
 {
     const NSRect contentRect = [window->ns.view frame];
--- a/src/internal.h
+++ b/src/internal.h
@@ -544,6 +544,11 @@ void _glfwPlatformGetWindowSize(_GLFWwindow* window, int* width, int* height);
  */
 void _glfwPlatformSetWindowSize(_GLFWwindow* window, int width, int height);
 
+/*! @copydoc glfwSetWindowIcons
+ *  @ingroup platform
+ */
+void _glfwPlatformSetWindowIcons(_GLFWwindow* window, const GLFWimage* images, int count);
+
 /*! @copydoc glfwGetFramebufferSize
  *  @ingroup platform
  */
--- a/src/mir_window.c
+++ b/src/mir_window.c
@@ -531,6 +531,12 @@ void _glfwPlatformSetWindowSize(_GLFWwindow* window, int width, int height)
                     "Mir: Unsupported function %s", __PRETTY_FUNCTION__);
 }
 
+void _glfwPlatformSetWindowIcons(_GLFWwindow* window, const GLFWimage* images, int count)
+{
+    _glfwInputError(GLFW_PLATFORM_ERROR,
+                    "Mir: Unsupported function %s", __PRETTY_FUNCTION__);
+}
+
 void _glfwPlatformGetWindowSize(_GLFWwindow* window, int* width, int* height)
 {
     if (width)
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -916,6 +916,13 @@ void _glfwPlatformSetWindowSize(_GLFWwindow* window, int width, int height)
     }
 }
 
+void _glfwPlatformSetWindowIcons(_GLFWwindow* window, const GLFWimage* images, int count)
+{
+    // TODO
+    _glfwInputError(GLFW_PLATFORM_ERROR,
+                    "Win32: Set window icons not implemented yet");
+}
+
 void _glfwPlatformGetFramebufferSize(_GLFWwindow* window, int* width, int* height)
 {
     _glfwPlatformGetWindowSize(window, width, height);
--- a/src/window.c
+++ b/src/window.c
@@ -514,6 +514,15 @@ GLFWAPI void glfwSetWindowSize(GLFWwindow* handle, int width, int height)
     _glfwPlatformSetWindowSize(window, width, height);
 }
 
+GLFWAPI void glfwSetWindowIcons(GLFWwindow* handle, const GLFWimage* images, int count)
+{
+    _GLFWwindow* window = (_GLFWwindow*) handle;
+
+    _GLFW_REQUIRE_INIT();
+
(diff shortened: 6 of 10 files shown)


elmindreda:
I usually do the remaining ones myself unless someone beats me to it. Many features I implement on all platforms with no PRs.

This looks like a good start. You should use `GLFWimage` instead of separate parameters.

There is also the question of multiple icon sizes. The proposed implementation in the [windowicon](https://github.com/glfw/glfw/commits/windowicon) branch solves this by passing in an array of images. However, that branch has other issues.

Then there's the issue of window vs. application icons. Some platforms provide one but not the other, except for X11 where it also depends on the chosen desktop system.

I will have more time to look into this once 3.1.1 is out, which it should be this week.

linkmauve:
I will probably restart from this branch, which I didn’t know.

I’m not aware of multiple icons support in X11 using the _NET_WM_ICON atom, maybe using an extension?

About application vs. window, on both X11 and Wayland (and Mir?) the .desktop file provides the former, while the latter (the object of this PR) isn’t supported by Wayland currently.

Take your time, I am not in any kind of hurry. :)

slimsag:
> I’m not aware of multiple icons support in X11 using the _NET_WM_ICON atom, maybe using an extension?

According to [the specification](http://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472568384):

> This is an array of possible icons for the client. This specification does not stipulate what size these icons should be, but individual desktop environments or toolkits may do so. The Window Manager MAY scale any of these icons to an appropriate size.

I assume decent window managers would choose the icon of the closest/best size and then scale it as needed.

linkmauve:
I changed the API to support multiple images (exactly like in that other branch), and fixed the X11 implementation accordingly.
There is still no Win32, Mir or Cocoa implementation.

elmindreda:
That looks really good.

elmindreda:
This is related to #453.

jaredjones:
I would like to contribute to this PR for OS X. @elmindreda how would you prefer for me to submit the work I've done? Another PR (ehh)?

elmindreda:
@jaredjones Make a PR to some branch. I'll be merging them manually anyway.

jaredjones:
Window icons in OS X have a different meaning than they do in other operating systems and window managers. Window icons in OS X symbolize links and urls for saving documents on the file system. This is why you'll notice that many applications do not have icons unless they deal with documents or files. On OS X there is the application icon and that is it. The App Icon is loaded from the plist when the program first runs.
You can get OS X to display a window icon using the file url method, but once again this is misusing the AppKit API because that's not what icons are for.

Suggestion: I wouldn't support icons in GLFW or at least in OS X for that matter. I've personally been setting icons for my programs by utilizing Plists and on Windows I run a .RC file that packages and ICO into my program, by default that adds an icon to the Window.

CODE:
// If the count is set to 0 or no images passed in, remove icon.
    if (images == NULL || count == 0)
    {
        [window->ns.object setRepresentedURL:[NSURL URLWithString:nil]];
    }
    // Reset's the Window Title: This is necessary to add an image icon to the Window
    // This will create the default document icon automatically.
    [window->ns.object setRepresentedURL:[NSURL URLWithString:[window->ns.object title]]];

linkmauve:
On Fri, Mar 20, 2015 at 05:32:54PM -0700, Jared Jones wrote:
> Window icons in OS X have a different meaning than they do in other operating systems and window managers. Window icons in OS X symbolize links and urls for saving documents on the file system. This is why you'll notice that many applications do not have icons unless they deal with documents or files. On OS X there is the application icon and that is it. The App Icon is loaded from the plist when the program first runs.
> You can get OS X to display a window icon using the file url method, but once again this is misusing the AppKit API because that's not what icons are for.

This sounds like the same reasoning as on Wayland (where it’s actually totally impossible currently to do that), that is that the application has an icon, defined in the .desktop, and individual windows use the same.

My usecase is for an emulator, to display the game’s icon instead of the emulator’s icon, since conceptually the user wants to know which game is played, not through which means (kind of like any launcher platform).

> Suggestion: I wouldn't support icons in GLFW or at least in OS X for that matter. I've personally been setting icons for my programs by utilizing Plists and on Windows I run a .RC file that packages and ICO into my program, by default that adds an icon to the Window.

That’s what we already do, but it’s somewhat meaningless for the user imo.

jaredjones:
Yea, it's just weird having to implement something that works for X11 and Windows but for OS X window icons have a completely different meaning.
So what should GLFW do about it? Allow only X11 and Windows to set icons? I mean this sounds perfectly viable to me. I wonder what SDL and SFML do for windowing icons.

elmindreda:
> So what should GLFW do about it? Allow only X11 and Microsoft Windows to set icons?

That's my plan, so far.

rwesterlund:
I've tried out the code and there were a few things I had to change to make it work by comparing with the SDL2 implementation.

1. I'm running a fairly unmodified Ubuntu 14.04 LTS, and `_NET_WM_ICON` does not get returned with the list of supported atom so the reference will be set to `None`. This will later crash when calling `_glfwPlatformSetWindowIcon` with `BadAtom` so there should be a guard against that.

2. I've never done any X11 programming before so I don't know much about getting a supported atom versus creating one, but SDL2 simply calls `XInternAtom` with the parameter `only_if_exists` set to False for any atom it wants so I tried replacing `getSupportedAtom(supportedAtoms, atomCount, "_NET_WM_ICON")` with `XInternAtom(_glfw.x11.display, "_NET_WM_ICON", False)`. This gives me a valid atom to send later.

3. At the end of `_glfwPlatformSetWindowIcon` I had to add a call to `XFlush(_glfw.x11.display)` before the icon would finally show up in the unity taskbar.

Sacryn:
Using this solution with rwesterlunds changes seem to work fine unless you run your application in full screen (providing a GLWFmonitor pointer to glfwCreateWindow). If starting in full screen then changing the icon via glfwSetWindowIcon never bites (Ubuntu still shows a question mark when alt-tabbing).

linkmauve:
Sacryn, you likely want to use a .desktop file instead, which is the standard way to associate a running application with its description on Linux.

I am two months late, but I will test the proposed fixes from SDL2 tonight, thanks rwesterlund. :)

Sacryn:
We do use a .desktop file (auto generated by Steam our case). Which for some reason seem to be ignored. Manually setting the icon via this addition seem to work though. But not when the game launches into full screen mode immediately (which causes the default question mark icon to appear instead).

Spasi:
> I'm running a fairly unmodified Ubuntu 14.04 LTS, and _NET_WM_ICON does not get returned with the list of supported atom so the reference will be set to None. This will later crash when calling _glfwPlatformSetWindowIcon with BadAtom so there should be a guard against that.

We're hitting this bug too with the current implementation (Ubuntu 15.10).

> I've never done any X11 programming before so I don't know much about getting a supported atom versus creating one, but SDL2 simply calls XInternAtom with the parameter only_if_exists set to False for any atom it wants so I tried replacing getSupportedAtom(supportedAtoms, atomCount, "_NET_WM_ICON") with XInternAtom(_glfw.x11.display, "_NET_WM_ICON", False). This gives me a valid atom to send later.

This indeed works; getting a valid atom and the window icon is set correctly after a call to `glfwSetWindowIcon`. I'm not sure if GLFW should do this though. Maybe it's best if it encourages users to follow OS best practices (.desktop files in this case).

elmindreda:
There's no harm in creating them; seems I was just being paranoid. Will fix.

elmindreda:
@Spasi See if 0b6a4313dd6074cb4e2614e14389961c135076d3 helps.

Spasi:
> See if 0b6a431 helps.

Nope. The problem is [here](https://github.com/glfw/glfw/blob/0b6a4313dd6074cb4e2614e14389961c135076d3/src/x11_init.c#L430), the list of `_NET_SUPPORTED` atoms does not include `_NET_WM_ICON` (or `_NET_WM_ICON_NAME`), so `getSupportedAtom` sets them to `None`.

Upgraded to Ubuntu 16.04 just now, same problem.

elmindreda:
Trying again, today with less headache and hopefully more reading comprehension. :flushed:

elmindreda:
@Spasi See if 932a161d44386987ab8af7fa92173044246da872 helps.

Spasi:
> See if 932a161 helps.

No crash and icons work. Thanks!

ws909:
> > So what should GLFW do about it? Allow only X11 and Microsoft Windows to set icons?
> 
> That's my plan, so far.

It's perfectly possible to change the application Dock icon of applications in macOS at runtime. Activity Monitor (by Apple) animates the window icon to present a graph of resource usage. `glfwSetWindowIcons` could be implemented to set the Dock icon instead. Though, if I'm not mistaken, changing the task bar icon is also possible in Windows, though I'm unaware if that actually is the window icon. If not, a `glfwSetApplicationIcons` could be of interest.

Issue #41, opened 2013-06-13
IME input


elmindreda:

elmindreda:
[Input Method Manager](https://msdn.microsoft.com/en-us/library/windows/desktop/dd318641.aspx)

ocornut:
FYI as of current GLFW, using Microsoft IME to input e.g. Japanese, the key pressed during IME inputs seems to reach the application directly via regular WM_KEYDOWN messages.

I don't even know what is causing this, in my simple test application for DirectX9 with a trivial message loop and no specific code ( https://github.com/ocornut/imgui/tree/master/examples/directx9_example ) the IME is intercepting those keys correctly. I am now investigating what is causing the difference to see if I can suggest a fix for GLFW.

ocornut:
OK so what is happening is, in my application I am using wParam which contains virtual key-code and can be filtered by IME. GLFW in translateKey() uses HIWORD(lParam)&0x1FF which is scancode and is never filtered.

So when you press and release a letter key intercepted by IME, the sequence goes something like:
```
WM_KEYDOWN 229
13.36 Key 49 pressed
WM_KEYUP 78
13.43 Key 49 released
```

So to emulate something similar from a GLFW user point of view, it may be enough to add this to translateKey():
```
if (wParam == VK_PROCESSKEY)
	return _GLFW_KEY_INVALID;
```
Which will generate neither pressed neither released events.

Now the problem is that there so many possible ways to interpret the low-data but I believe, and in spite of the fact that there is so much more to IME and this isn't the entire picture, that this behaviour would be the desirable default and very beneficial. The vast majority of applications don't want to react to key-presses when the IME window is displayed, which at least would allow IME-based text inputs into the application.

Application that strives to be 100% IME aware (down to implementing their own IME display), which is extremely extremely rare would have anyway to handle many more messages in great details, which will probably never be the scope of GLFW (but one can imagine that GLFW may in the future provide WM_* hooks to allow the interested user to handle them themselves, and again there's no point in doing that work unless there's a good reason to).

kovidgoyal:
FYI I implemented support for IME via IBUS on X11/wayland in my glfw fork.

qiujiangkun:
Any progress, especially fcitx?

runner-mei:
can fix it soon!

qiujiangkun:
[...]

PR #473, opened 2015-03-21
XIM doesn't work because some events are not filtered with XFilterEvent

mattn:
I'm using  japanese input method on linux. XIM works with filtering some X events. KeyPress/KeyRelease/ClientMessage etc. So the all of events should be filtered with XFilterEvent.


elmindreda:
Even with this is fixed Japanese input won't work unless one has a kana keyboard layout. I'll be adding proper support for pre-edit input method styles later with #41, but that requires adding new API functions.

PR #475, opened 2015-03-24
Fixes for memory leaks

Kvalme:
Some core resources not freed when glfw terminates:
monitors, modes, xkb


linkmauve:
Hello @Kvalme, are you planning on fixing this PR?

Issue #492, opened 2015-04-06
Switching back to a fullscreen application on Yosemite with multiple monitors crashes the application


chanchikwan:
The problem should be reproducible by running the following code on OS X Yosemite with a dual-monitor setup.  The code creates a fullscreen window.  Simply press "Cmd+Tab" twice to switch to a different application and switch back, the code crashes.

	#include <GLFW/glfw3.h>
	
	int main()
	{
		glfwInit();
	
		GLFWmonitor *m = glfwGetPrimaryMonitor();
		GLFWwindow  *w = glfwCreateWindow(640, 480, "buggy", m, NULL);
	
		while(!glfwWindowShouldClose(w))
			glfwPollEvents();
	
		glfwTerminate();
	
		return 0;
	}

elmindreda:
If the object is prone to disappearing it's probably better not to cache it in the first place. The screen of a `GLFWmonitor` isn't used very often anyway. Or, if it is to be cached, it should be updated on `NSApplicationDidChangeScreenParametersNotification`.

PR #493, opened 2015-04-06
Fix bug for fullscreen window on Yosemite with multiple monitors

chanchikwan:
This is a quick-and-dirty fix to issue #492.  It checks if

	window->monitor->ns.screen

points to a valid screen, and update it when necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,8 @@ GLFW bundles a number of dependencies in the `deps/` directory.
 
 ## Changelog
 
+ - Bugfix: Initialization failed on headless systems
+
 
 ## Contact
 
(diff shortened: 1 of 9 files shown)


elmindreda:
This is a valid issue but I don't agree with the solution. Please see my reply to #492 for details.

chanchikwan:
How about this solution?  As you pointed out, the screen of a GLFWmonitor isn't used very often so I just remove it form _GLFWmonitorNS.  A new function

    static NSScreen *getScreen(CGDirectDisplayID displayID)

is added at the beginning of `src/cocoa_window.m` to get the screen from a displayID only when we need it.

elmindreda:
That looks good. Thank you!

On an unrelated note, your editor appears to remove the last line of some source files, which introduces noise to your patches. I've fixed it in these commits while merging, but it may be something to look into.

PR #496, opened 2015-04-10
Reduce some repetition in input.dox

dschleck:


elmindreda:
Thank you for spotting this!

Issue #225, opened 2014-01-28
Segfault on Linux when RandR not present


larsoner:
I have been configuring Travis to run tests, and I hit one that _seems_ to indicate an issue with `glfw` assuming `RANDR` is an enabled extension, see the output of:

https://travis-ci.org/Eric89GXL/vispy/jobs/17791008

In my case, I haven't yet been able to work around it easily by starting `Xvfb` with different options, but I'll keep trying.

Does this indicate an underlying issue with how extensions are handled by the library in Linux, when `RANDR` is not enabled?

PR #498, opened 2015-04-13
FIX: Should be null in case of no monitors

### LINKED ISSUE #497: Segfault on glfwTerminate
As of today, we get a segfault when calling `glfwTerminate` (https://travis-ci.org/vispy/vispy/jobs/58080099#L3551):

```
*** glibc detected *** /home/travis/virtualenv/python2.7_with_system_site_packages/bin/python: free(): invalid pointer: 0x00007fff8841d180 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x7e626)[0x2b0051476626]
/usr/local/lib/libglfw.so.3(+0x7a06)[0x2b005708fa06]
/usr/local/lib/libglfw.so.3(glfwTerminate+0xc8)[0x2b005708dffd]
```

We haven't had this problem before. I think the last time we would have run/tested would have been 1-2 days ago.

I can try to actually turn this into some minimal example, but I figured I'd ping here to see if someone made a change in the last couple of days that might be causing this.

Eric89GXL:
Closes #497.

Saves from a segfault when no monitors are found (e.g., when using Xvfb).

FWIW this would have been caught if some basic tests were run using Travis, as would have #225. My offer to implement some basic Travis support still stands :)

### DIFF
--- a/src/x11_monitor.c
+++ b/src/x11_monitor.c
@@ -201,7 +201,7 @@ void _glfwRestoreVideoMode(_GLFWmonitor* monitor)
 _GLFWmonitor** _glfwPlatformGetMonitors(int* count)
 {
     int i, j, k, found = 0;
-    _GLFWmonitor** monitors;
+    _GLFWmonitor** monitors = NULL;
 
     *count = 0;
 


elmindreda:
Thank you for the offer but I need to learn it properly. Travis CI and Coverity Scan are both on my list for this release cycle.

PR #499, opened 2015-04-15
Implement glfwGetCursorPos

linkmauve:
Pointer motion events were already handled, it was just lacking the actual exposition to the API.

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -86,6 +86,11 @@ static void pointerHandleMotion(void* data,
                         "Wayland: GLFW_CURSOR_DISABLED not supported");
         return;
     }
+    else
+    {
+        window->wl.cursorPosX = wl_fixed_to_double(sx);
+        window->wl.cursorPosY = wl_fixed_to_double(sy);
+    }
 
     _glfwInputCursorMotion(window,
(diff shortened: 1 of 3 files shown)


linkmauve:
Oops, my rebase erased jadahl’s comment (and fixed the main issue).

Anyway, to answer that question, I think pointer events from outside of the window are very useful in some cases, think an OpenLayers-like map application, the user wants to be able to scroll even outside of the window, it makes a frustrating experience if the window is small and input stops at the edges of the window.

elmindreda:
The comment is still there: linkmauve/glfw@be327a324d1bf58b64d3fb649250c010d92a54aa

jadahl:
The wl_*.c part looks fine now. @elmindreda, can you confirm the change to input.c is correct and won't break other platforms?

elmindreda:
@jadahl It's... tolerable. I'd rather it use platform local variables instead of re-introducing shared logic that will do nothing on any other platform.

If Mir works the same as Wayland then I guess this way is all right.

linkmauve:
Addressed your issues, everything stays in the wl_* files now.

elmindreda:
Merged. Thank you!

Issue #500, opened 2015-04-16
glfwCreateWindow does not restore previous context


chanchikwan:
When using multiple windows on OS X Yosemite, creating a fullscreen window after non-fullscreen windows seems to destroy existing OpenGL context.

For example, the following code creates a non-fullscreen window `w0`, a fullscreen window `w1`, and a non-fullscreen window `w2`:
```
#include <GLFW/glfw3.h>
#include <cstdio>

int main()
{
        glfwInit();

        GLFWmonitor *m = glfwGetPrimaryMonitor();

        GLFWwindow *w0 = glfwCreateWindow(640,  480, "0", NULL, NULL);
        printf("%s\n", glGetString(GL_VERSION)); // an OpenGL context is available

        GLFWwindow *w1 = glfwCreateWindow(1440, 900, "1 buggy", m, w0);
        printf("%s\n", glGetString(GL_VERSION)); // the existing OpenGL context is gone.  Is this a bug?

        GLFWwindow *w2 = glfwCreateWindow(640,  480, "2", NULL, w1);
        printf("%s\n", glGetString(GL_VERSION)); // an OpenGL context is available again

        while(!glfwWindowShouldClose(w0) &&
              !glfwWindowShouldClose(w1) &&
              !glfwWindowShouldClose(w2))
                glfwPollEvents();

        glfwTerminate();

        return 0;
}
```
The code outputs:
```
2.1 NVIDIA-10.2.7 310.41.25f01
(null)
2.1 NVIDIA-10.2.7 310.41.25f01
```
After the creation of `w0`, a valid OpenGL context is available as expected and `glGetString(GL_VERSION)` gives the correct OpenGL version string.  However, with the creation of `w1`, the existing OpenGL context gets destroyed (or become inactive).  Calling OpenGL functions may crash the program.  And `glGetString(GL_VERSION)` gives a `null`.  Creating a non-fullscreen window `w2` fixes this problem and a valid OpenGL context becomes available again.

This problem can also be fixed by calling `glfwMakeContextCurrent(w1);` right after the creation of `w1`.  However, is this an expected behavior?  Or is it a bug?

elmindreda:
That seems like a bug. There should be no context current after any of those calls.

elmindreda:
The cause of this issue (and #443) is the NSOpenGLView class trying to be clever. I will likely fix both by going back to a plain NSView.

jaredjones:
Okay great! I'm glad you figured it out.

[...]

PR #504, opened 2015-04-26
Properly restore previous context on OS X Yosemite

chanchikwan:
This is a fix to issue #500.

_glfwPlatformShowWindow(window) and _glfwPlatformUnhideWindow(window)
implicitly make the context of window current on OS X Yosemite, which
undoes _glfwPlatformMakeContextCurrent(previous).  To properly restore
the previous context, we call _glfwPlatformMakeContextCurrent(previous)
just before glfwCreateWindow() returns.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1061,12 +1061,16 @@ void _glfwPlatformShowWindow(_GLFWwindow* window)
     //       should probably not be done every time any window is shown
     [NSApp activateIgnoringOtherApps:YES];
 
+    _GLFWwindow* previous = _glfwPlatformGetCurrentContext();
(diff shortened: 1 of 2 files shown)


elmindreda:
In that case the restore needs to be done inside `_glfwPlatformShowWindow` and `glfwPlatformUnhideWindow`. Otherwise `glfwShowWindow` will still act weirdly.

chanchikwan:
Now I restore the previous context in `_glfwPlatformShowWindow` and `_glfwPlatformUnhideWindow`.  Do you think `glfwShowWindow` will work correctly with this?

elmindreda:
I think so. I'll test it as soon as I have access to the Mac again.

PR #505, opened 2015-05-03
Optional wiring of gl3w into glfw to correct broken linking

huhlig:
Option for building gl header files if unavailable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -529,6 +529,18 @@ write_basic_package_version_file("${GLFW_BINARY_DIR}/src/glfw3ConfigVersion.cmak
                                  VERSION ${GLFW_VERSION_FULL}
                                  COMPATIBILITY SameMajorVersion)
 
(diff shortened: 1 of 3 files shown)


elmindreda:
Libraries like gl3w should be built top of GLFW, not the other way around.

huhlig:
Except you include GL.h inside of glfw which causes issues with include ordering because of c++ sillyness.

elmindreda:
Easily solved in the general case with `GLFW_INCLUDE_NONE`.

In this specific case, a public header for such a library wouldn't necessarily need to include the GLFW header and could safely be [included first](http://www.glfw.org/faq.html#can-i-use-extension-loaders-with-glfw). An stb style header-only version of such a library could include the GLFW header [itself](https://github.com/elmindreda/greg/blob/master/templates/greg.h.in#L98), disabling any later inclusion by the application. Either of these options would prevent the regular OpenGL header from interfering.

Issue #165, opened 2013-10-27
CGDisplayIOServicePort and SetFrontProcess deprecated on 10.9


fatto:
Starting from Mavericks 'CGDisplayIOServicePort' and 'SetFrontProcess' are deprecated
the warnings are from:
- cocoa_monitor.m:48:42: `info = IODisplayCreateInfoDictionary(CGDisplayIOServicePort(displayID)`
- cocoa_window.m:823:5: `SetFrontProcess(&psn);`

PR #508, opened 2015-05-10
Fixed: Deprecated CGDisplayIOServicePort

jaredjones:
*A function has manually been created that deals with grabbing the approriate service_io port.
*This function adds support for collision detection against the VendorID, ProductID, and the SerialNumber. The serial number check is required incase a user has two types of identical monitors.
*Supports OS X 10.4 (Tiger) and Newer (Untested, based on API Docs)

I would like to give special thanks to "@mhernr18" for providing the initial code, which is the majority of what this PR is based on.
Recloses:   #165
Closes PR: #192

### DIFF
--- a/include/GLFW/glfw3native.h
+++ b/include/GLFW/glfw3native.h
@@ -56,6 +56,7 @@ extern "C" {
  *  The available context API macros are:
  *  * `GLFW_EXPOSE_NATIVE_WGL`
  *  * `GLFW_EXPOSE_NATIVE_NSGL`
+ *  * `GLFW_EXPOSE_NATIVE_CGL`
  *  * `GLFW_EXPOSE_NATIVE_GLX`
  *  * `GLFW_EXPOSE_NATIVE_EGL`
  *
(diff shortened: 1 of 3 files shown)


jaredjones:
This is based on the PR provided by @mhenr18. The only difference is that the serial number is also being used to check for the appropriate monitor incase the user is running two identical monitors. This PR also has cleaner code and doesn't have all the whitespace.

elmindreda:
None of those numbers are guaranteed to be present. Since you're already touching IOKit, have a look at `CGDisplayUnitNumber` instead.

jaredjones:
That's a good idea however the "UnitNumber" is not placed inside of the dictionary for the "io_service_t". So it's not possible to compare the unit number while iterating through io ports.

jaredjones:
However if the Vendor/Product/SN are not available then the monitor name likely won't be available either.

elmindreda:
> However if the Vendor/Product/SN are not available then the monitor name likely won't be available either.

That's a good point.

mpflanzer:
Please see my follow up on this: https://github.com/glfw/glfw/pull/511

jaredjones:
Closed in Reference to: #511

Issue #25, opened 2013-05-17
Creating GLFW window objects with existing native handles


martinbis11:
It's really great that we can access native data from the OpenGL context such as glfwGetWin32Window() and glfwGetWGLContext().

It would be nice if that part of the API would be extended to wrap a window that has been created outside of GLFW into GLFW structures for OpenGL context creation, or wrap a context that was created outside of GLFW into GLFW strucutres so that we can use GLFW functions on that context (such as glfwSwapInterval()) for instance.

An example of where this would useful is for screensaver development.  Let's say an OpenGL cross-platform screensaver is to be developed.  On Windows and Linux for instance, writing a screensaver starts by the OS giving you a window in a platform-specific way.  It would be nice to be able to give that OS-specific window handle to GLFW so that from then on, we can use platform-independent GLFW calls to create the context, swap buffers, etc.

TronicLabs:
+1 for window created upon a child window using it's handle

ultrant:
+1 for window created upon a child window using it's handle

galek:
+1

WRodewald:
+1

hsdk123:
It's difficult to believe that this feature has been put off for 3 years - it would be great for this to be incorporated.

TronicLabs:
+3 YEAR LATER
:)

elmindreda:
Sorry it's taken so long. It's been difficult to schedule this partly because a lot of internals needed restructuring. That work is done now.

hsdk123:
@elmindreda Woah, that sounds great! Does this mean we'll see progress on this now? (been waiting for this for 2+ years now)

starseeker:
Would the API for this look something like:

GLFWAPI GLFWwindow* glfwWrapWindowWin32(HWND *win, HGLRC *context, GLFWwindow *share);

in glfw3native.h? (With, obviously, appropriate functions for other backends)

There seem to be three possible scenarios:

1) given context without window, create and provide a wrapping GLFWwindow
2) given a window without an OpenGL context, initialize the context and create the GLFWwindow using the existing OS info (my case - have Tk window, need OpenGL context + cross platform OpenGL management of it.)
3) given a window and context from the OS (i.e. a fully created, working OpenGL window from another toolkit or something like OGRE or OSG) populate the GLFWwindow structure with the necessary information to allow GLFW wrapper functions to manipulate it. (I may have use for this case down the road...)

[...]

PR #509, opened 2015-05-10
Added ability to reuse a context of another window

thekend:
When using multiple windows with OpenGL, there is some options:
- create the windows with different independent GL contexts
- create the windows with different GL contexts sharing resources
- create the windows with the same GL context

At current time the third option is not supported by GLFW, however sometimes it's useful. This can be achieved with a new window hint: the target name can be GLFW_REUSE_CONTEXT. When this is set to true and a non-null share parameter is added to glfwCreateWindow, then the context of the share window is used for the created window instead of creating a new one with sharing resources by GL contexts.

Implementation is only provided for Win32 now.


elmindreda:
You don't reference count the context so it'll get destroyed along with the first window that shares it, or depending on hints, by context recreation. You also look up a new pixel format and risk incompatibility. This design requires you to ignore all framebuffer and context related hints, which is confusing to the user and not currently done.

This is better solved by #25.

thekend:
I agree that the implementation is highly incomplete now (it only answers my own purpose in practice). Maybe the idea should be added to the issues instead.

The problem may be solved better with https://github.com/glfw/glfw/issues/25, however  (if I haven't missed something) this involves creating a native window, which is uncomfortable (at least in my point of view: then why use glfw).

Yes, this solution requires to ignore all framebuffer and context related hints. I don't feel this is very confusing when one wants to do reusing a context of another window.
Then what could be a better design? Another window creation function? Any other idea, directions?

elmindreda:
Exactly what different options will be provided by #25 is influenced by exactly these kinds of questions. Please add your $0.02 to the thread.

Issue #443, opened 2015-02-14
Crash while resizing the application window


salino:
The GLFW application crashes when I resize the window vertically towards to zero. This happens with any example I tried so far.

- OS X Yosemite (10.10.2)
- GLFW master

In cocoa_window.m, line 646ff:

    // From http://cocoadev.com/index.pl?GameKeyboardHandlingAlmost
    // This works around an AppKit bug, where key up events while holding
    // down the command key don't get sent to the key window.
    - (void)sendEvent:(NSEvent *)event
    {
        if ([event type] == NSKeyUp && ([event modifierFlags] & NSCommandKeyMask))
            [[self keyWindow] sendEvent:event];
        else
            [super sendEvent:event]; // <-------------- Thread 1: EXC_BAD_ACCESS
    }

![event](https://cloud.githubusercontent.com/assets/4344802/6201273/7e4115f0-b49f-11e4-8e7b-4e2ca2f47fad.png)
![glfw_error](https://cloud.githubusercontent.com/assets/4344802/6201258/f743843e-b49e-11e4-9677-99f097cc624f.png)

jaredjones:
I can't seem to figure this one out... NSZombieEnabled is turned on and I can't seem to figure out what virtual memory address does not exist that's causing the EXE_BAD_ACCESS. I wouldn't be surprised if the bug is in the Cocoa API somewhere.

shurcooL:
I can confirm this happens.

elmindreda:
If those really are stack frame indices then it's likely that the workaround cited above causes a recursion loop.

jaredjones:
Nope, not a recursion loop. I placed a printf in that method and didn't get any output right before it crashed.

jaredjones:
Left Side: x86 Code of the Crash
Right Side: [NSApplication SendEvent:]; implementation.
![screen shot 2015-04-20 at 2 20 34 pm](https://cloud.githubusercontent.com/assets/1316894/7238503/877e53c0-e768-11e4-8b2b-4a4542cb0ee9.png)

elmindreda:
Please see if the above commit resolves this issue for you.

jaredjones:
It worked. :+1:

salino:
Works for me as well :smile:

PR #510, opened 2015-05-12
OS X Fix: NSWindow no Longer Crashes on Resize

jaredjones:
*After doing research I found that Apple puts an NSWindow to sleep using (AppNap) after the window reaches below 4 pixels in height.
*GLFW will now set a minimum height with the OS X default title bar height taken into consideration; this resolves the issue.

Closes: #443

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -648,6 +648,17 @@ static int translateKey(unsigned int key)
     return YES;
 }
 
+- (float)titleBarHeight
+{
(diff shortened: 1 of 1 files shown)


elmindreda:
Turns out this crash also goes away if one reverts 81bcefe86c1bb9f3bafcc89dbc5c59a4db3c7335, replacing the NSOpenGLView with a regular NSView. Seems Apple is trying to be a little too clever again.

elmindreda:
This approach also fixes #500 (and hence obviates #504).

PR #511, opened 2015-05-12
Fixed problem with not existing serial number for built-in display

mpflanzer:
The check the serial number that jaredjones introduced seems not to work for me. I have the problem that I cannot get a serial number for the built-in display on my MacBook Air (OS X 10.10.3). It is not contained in the IODisplay info dictionary and CGDisplaySerialNumber returns zero.
For an external monitor it works without problems. All information is returned correctly. Since also the system report shows no serial number for the built-in display I guess Apple not included any.

I changed the code for the check such that is is also successful if the serial number does not exist or is zero. There is in any case no way do distinguish to monitors of the same model if they do not have a serial number.

### DIFF
--- a/include/GLFW/glfw3native.h
+++ b/include/GLFW/glfw3native.h
@@ -56,6 +56,7 @@ extern "C" {
  *  The available context API macros are:
  *  * `GLFW_EXPOSE_NATIVE_WGL`
  *  * `GLFW_EXPOSE_NATIVE_NSGL`
+ *  * `GLFW_EXPOSE_NATIVE_CGL`
  *  * `GLFW_EXPOSE_NATIVE_GLX`
  *  * `GLFW_EXPOSE_NATIVE_EGL`
  *
@@ -94,6 +95,8 @@ extern "C" {
  /* WGL is declared by windows.h */
 #elif defined(GLFW_EXPOSE_NATIVE_NSGL)
  /* NSGL is declared by Cocoa.h */
+#elif defined(GLFW_EXPOSE_NATIVE_CGL)
+ #include <OpenGL/OpenGL.h>
 #elif defined(GLFW_EXPOSE_NATIVE_GLX)
  #include <GL/glx.h>
 #elif defined(GLFW_EXPOSE_NATIVE_EGL)
@@ -220,6 +223,14 @@ GLFWAPI id glfwGetCocoaWindow(GLFWwindow* window);
  *  @ingroup native
  */
 GLFWAPI id glfwGetNSGLContext(GLFWwindow* window);
+#endif
+    
+#if defined(GLFW_EXPOSE_NATIVE_CGL)
+    /*! @brief Returns the `CGLContextObj` of the specified window.
+     *  @return The `CGLContextObj` of the specified window.
+     *  @ingroup native
+     */
+    GLFWAPI CGLContextObj glfwGetCGLContext(GLFWwindow* window);
 #endif
 
 #if defined(GLFW_EXPOSE_NATIVE_X11)
--- a/src/cocoa_monitor.m
+++ b/src/cocoa_monitor.m
@@ -55,55 +55,58 @@ static io_service_t IOServicePortFromCGDisplayID(CGDirectDisplayID displayID)
     if (err)
         return 0;
     
-    while ((serv = IOIteratorNext(iter)) != 0)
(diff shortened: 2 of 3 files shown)


jaredjones:
Ah, thanks for fixing that nullptr mistake with the SN.
In my personal opinion it should be fine if we can't distinguish between the two for the sake of simply getting the DisplayName. If the vendor and product both match up then I'm pretty positive that any multiple displays will have the same name.

elmindreda:
@mpflanzer What is the output of the `monitors` test from `master` on that machine?

jaredjones:
I'm curious as well because on my Retina MBP I get a monitor name off of master.
Name: Color LCD (primary)
Current mode: 1440 x 900 x 24 (8 8 8) 59 Hz
Virtual position: 0 0
Physical size: 331 x 206 mm (110 dpi)
Modes:
  0: 640 x 480 x 24 (8 8 8) 59 Hz
  1: 800 x 600 x 24 (8 8 8) 59 Hz
  2: 1024 x 768 x 24 (8 8 8) 59 Hz
  3: 1280 x 800 x 24 (8 8 8) 59 Hz
  4: 1440 x 900 x 24 (8 8 8) 59 Hz (current mode)
  5: 1680 x 1050 x 24 (8 8 8) 59 Hz
  6: 2048 x 1280 x 24 (8 8 8) 59 Hz
  7: 2560 x 1600 x 24 (8 8 8) 59 Hz
  8: 2880 x 1800 x 24 (8 8 8) 59 Hz

mpflanzer:
With the current master from glfw I also correctly get settings for the internal display, like @jaredjones. The alternative from jaredjones to replace the deprecated version results in a null pointer exception for me since there seems to be no serial number for the internal display. My version should fix this issue such that the deprecated version could be replaced.

Thanks for the info, jaredjones. I was not sure how to do it the proper way.

jaredjones:
No she's stating that if the old deprecated version of CGDisplayIOServicePort is actually giving you a display name like: "Color LCD" but yet the replacement version states: "Unknown" then we have a problem.

Can you checkout your latest commit and run the GLFW Test "Monitors" and paste the output in here.

mpflanzer:
Ah sorry. It thought it has been clear that I get the same output with the current master and with my changes.

Output GLFW master:
```
Name: Color LCD (primary)
Current mode: 1440 x 900 x 24 (8 8 8) 59 Hz
Virtual position: 0 0
Physical size: 285 x 178 mm (128 dpi)
Modes:
  0: 640 x 480 x 24 (8 8 8) 59 Hz
  1: 720 x 480 x 24 (8 8 8) 59 Hz
  2: 800 x 500 x 24 (8 8 8) 59 Hz
  3: 800 x 600 x 24 (8 8 8) 59 Hz
  4: 1024 x 640 x 24 (8 8 8) 59 Hz
  5: 1024 x 768 x 24 (8 8 8) 59 Hz
  6: 1152 x 720 x 24 (8 8 8) 59 Hz
  7: 1280 x 800 x 24 (8 8 8) 59 Hz
  8: 1440 x 900 x 24 (8 8 8) 59 Hz (current mode)
```

Output with my changes:
```
Name: Color LCD (primary)
Current mode: 1440 x 900 x 24 (8 8 8) 59 Hz
Virtual position: 0 0
Physical size: 285 x 178 mm (128 dpi)
Modes:
  0: 640 x 480 x 24 (8 8 8) 59 Hz
  1: 720 x 480 x 24 (8 8 8) 59 Hz
  2: 800 x 500 x 24 (8 8 8) 59 Hz
  3: 800 x 600 x 24 (8 8 8) 59 Hz
  4: 1024 x 640 x 24 (8 8 8) 59 Hz
  5: 1024 x 768 x 24 (8 8 8) 59 Hz
  6: 1152 x 720 x 24 (8 8 8) 59 Hz
  7: 1280 x 800 x 24 (8 8 8) 59 Hz
  8: 1440 x 900 x 24 (8 8 8) 59 Hz (current mode)
```

jaredjones:
Thanks man, that's what we wanted to see. :)

jaredjones:
@elmindreda 
Do you have any further issues with this PR?

elmindreda:
@jaredjones Sorry about the late reply. I just need to find time to review. I intend for this to happen before 3.1.2.

PR #532, opened 2015-06-12
Implement touch events on Wayland

linkmauve:
This is totally untested, as I don’t have any touch device around, but should be pretty ok (at least it compiles).

I went with the assumption a negative id wasn’t a valid one, but it isn’t specified so it could be wrong, this is a point to verify.

Also, if glfwSetTouchInput() disables touch input while some contact points are still present, they will stay pressed “forever”.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ skills.
  - Cameron King
  - Peter Knut
  - Eric Larson
+ - Quinten Lansu
  - Robin Leffmann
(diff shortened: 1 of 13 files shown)


jadahl:
I'm verry sorry for taking so much time before reviewing. I think it looks more or less good, so with at least the coding style fix and the bug fix regarding the potential invalid internat stalte I'll merge it.

PR #534, opened 2015-06-18
Removed Deprecated convertBaseToScreen

jaredjones:
In OS X 10.7 Apple deprecated convertBaseToScreen, however it is not up until the OS X 10.11 SDK that the compiler sets a warning. This small change silences the warning in the 10.11 SDK using Xcode 7.


jaredjones:
This breaks 10.6 and older, so runtime checks will need to be added.

elmindreda:
Thank you.

> This breaks 10.6 and older, so runtime checks will need to be added.

At least for 3.1.2. It's looking likely that 10.6 support will be removed in 3.2.

Issue #17, opened 2013-05-02
Joysticks connected after glfwInit not detected


elmindreda:
Joysticks connected after `glfwInit' has completed are not detected and hence cannot be used from GLFW. Resolving this involves using inotify or libudev on Linux and the IOKit on OS X. The proper solution on Linux is libudev, while the easy solution is inotify.

elmindreda:
This has been fixed for Linux in 3.1 with 56202ee7c61bea44ebea13f1b5efe06b6e4fd6e6. OS X remains to be fixed.

PR #536, opened 2015-06-20
Add support for joystick hot swapping

aaronmjacobs:
Fixes issue #17
Tested with Xbox 360, Xbox One, and PS4 controllers, and Cyborg flight stick

### DIFF
--- a/src/iokit_joystick.h
+++ b/src/iokit_joystick.h
@@ -33,17 +33,17 @@
 #include <IOKit/hid/IOHIDKeys.h>
 
 #define _GLFW_PLATFORM_LIBRARY_JOYSTICK_STATE \
-    _GLFWjoystickIOKit iokit_js[GLFW_JOYSTICK_LAST + 1]
+    _GLFWjoystickIOKit iokit_js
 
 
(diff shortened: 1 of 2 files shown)


elmindreda:
This duplicates all buttons and axes 2-3 times. Tested with wired Xbox 360 controller and clone.

aaronmjacobs:
The duplicates are being generated by the calls to CFArrayApplyFunction from within addJoystickElement.

Doing some quick testing, it would appear that when the elements were obtained from the properties given by IORegistryEntryCreateCFProperties (using the kIOHIDElementKey key), the children had to be traversed in order to discover all of the joystick elements. This seems to no longer be the case with IOHIDDeviceCopyMatchingElements - it appears to provide all elements in the returned CFArrayRef, making the traversal of the child elements incorrect.

aaronmjacobs:
Having gone back over Apple's documentation, reviewed other projects' handling of IOHIDElements, and doing further testing, it seems that the correct solution is to not recurse over the children. My pushed changes should fix the duplicate buttons and axes.

elmindreda:
Works beautifully.

Issue #478, opened 2015-03-24
always check for memory allocation failure


andrewrk:
A robust C library checks for allocation failures. I noticed that GLFW does not do this. For example, if you grep for `calloc` I don't see a single one that checks the return value.

elmindreda:
Yep. This is what `GLFW_OUT_OF_MEMORY` is waiting for.

elmindreda:
I keep meaning to do this but then actual bugs get in the way. Heap allocation for the tiny sizes GLFW deal with won't actually fail unless the system is toast. The best reason to do it anyway is to pass static analysis without warnings.

elmindreda:
Thank you for creating an issue for this. There's a bunch of these things that I've neither got round to doing or adding to the tracker.

BrettRToomey:
@elmindreda Shall I go through and cleanup these allocations? What do you want me to do with `GLFW_OUT_OF_MEMORY`?

PR #544, opened 2015-07-01
Support custom dynamic memory functions

anthonytsmith:
This allows users to use their own memory allocation/deallocation functions to be used in GLFW.  This enables users who want extra control over their application's use of memory (i.e. some games or engines) to have that.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -43,6 +43,11 @@ extern "C" {
  *  This is the reference documentation for context related functions.  For more
  *  information, see the @ref context.
  */
+/*! @defgroup memory Dynamic Memory
+ *
+ * This is the reference documentation for memory management related functions 
+ * and types.
+ */
 /*! @defgroup init Initialization, version and errors
  *
  *  This is the reference documentation for initialization and termination of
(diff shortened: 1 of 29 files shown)


elmindreda:
GLFW 3 actually had this functionality at one point during development. You can find it in the log. However, the amount of memory allocated by GLFW is tiny (on the order of a few KB) and almost all of it is allocated either at initialization or during window/context/cursor creation. During a typical frame _no memory is allocated_. Exceptions include the handling of controller connection and drag and drop events, which from the viewpoint of a game run-loop occur about once a century.

Therefore I decided it wasn't worth the added complexity and removed it (0c3b1b5a0ee571289f254945a0c9046ee4e872f7).

anthonytsmith:
My reasoning for adding it is because a lot of the applications I've worked on have allocated all of the memory up-front and proceeded to manage that memory.  Currently, GLFW does not play nice with applications that have that restriction.  However, using some simple regex find/replace and creating the rest of the structures/code, it can be easily added to GLFW should this feature not make it in a second time.

P.S. Thanks for linking to the commit.

elmindreda:
Further discussion depends on having fixed #478.

elmindreda:
Thank you for this PR. It has been superseded by #1947, which will add a very similar mechanism.

PR #545, opened 2015-07-02
Started adding memory allocation checks

BrettRToomey:
Regarding issue #478, I began by checking all instances of `calloc` for a `NULL` return. As I'm new to this project, I figured there's a chance I did something in a way you don't like, which is why I'm sending the request before I finish cleaning everything up.

### DIFF
--- a/src/cocoa_monitor.m
+++ b/src/cocoa_monitor.m
@@ -67,14 +67,14 @@ static char* getDisplayName(CGDirectDisplayID displayID)
     size = CFStringGetMaximumSizeForEncoding(CFStringGetLength(value),
                                              kCFStringEncodingUTF8);
     name = calloc(size + 1, sizeof(char));
-	if (name == NULL)
-	{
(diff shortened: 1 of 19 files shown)


shurcooL:
The formatting looks off. It seems you're using tabs while the rest of the project uses spaces.

BrettRToomey:
Ah, I knew I'd do something like that! My apologies, I'll get right on it!

elmindreda:
The problem with a patch like this is that it'd take me longer to verify than to write it again myself.

elmindreda:
As an example, `strdup` also performs heap allocation but this patch actually calls it in response to allocation failure.

andrewrk:
> The problem with a patch like this is that it'd take me longer to verify than to write it again myself.

Totally agreed. The pull request is a nice gesture but in the end it's kind of a one-person job.

elmindreda:
More than a gesture, I'd say, but unfortunately it's not useful in _this specific_ case.

PR #557, opened 2015-07-16
Add an option to link against libepoxy

linkmauve:
This allows glfw to defer the choice between libGL, libGLESv1_CM or libGLESv2 to runtime, while supporting every function and define from all three of those.


elmindreda:
I'm currently moving that choice to runtime without external dependencies, as well as the choice between native and EGL.

linkmauve:
libepoxy also helps in supporting multiple versions of a same symbol by aliasing the *KHR, *ARB, *EXT, *NV, etc. version of a same symbol.

But if you already have that, I guess libepoxy support is superfluous, or at least not *that* useful.

elmindreda:
It does not, but libepoxy should be on top of GLFW, not the other way around.

linkmauve:
The idea was to prevent libglfw.so from linking against libGL or libGLES*, as it still needs a library providing the various glGetInteger() or glGetString() and libepoxy is exactly that.

I will use libepoxy in Citra to let the user select either GL or GLES, and it is currently doing weird things when using the one that glfw isn’t compiled against.

elmindreda:
Yes, I understand, but GLFW must already know which library is appropriate or it won't be able to create contexts or provide `glfwExtensionSupported` and `glfwGetProcAddress`.

PR #567, opened 2015-07-27
Fix OS X spurious key pressed event when modifier key is held during window creation

### LINKED ISSUE #566: When released, modifier keys held during window creation are reported as pressed
Reproducible on OS X 10.10.4 for GLFW 3.1.1 and the latest in master:

1. Hold down any (up to all) modifier keys
2. Launch `tests/events`
3. Release modifier key(s) after window is created and in focus

Results:

```
xw@Lotus ~/glfw> ./tests/events
Library initialized
Creating windowed mode window 1 (640x480)
Main loop starting
00000000 to 1 at 0.145: Window focused
00000001 to 1 at 0.680: Key 0x0154 Scancode 0x0038 (LEFT SHIFT) (with no mods) was pressed
```

I looked into the Cocoa logic (I haven't tried this on any other platform) and there's a few ways for it break:

1. It looks like the modifiers are initialized to zero upon window creation and not to the user state.
2. The modifier key logic detects changes over the entire `NSEvent modifierFlags` vector, not just the one corresponding to the relevant key in `NSEvent keyCode`. This makes it harder to fix the case of multiple simultaneous modifiers (e.g. CTRL+SHIFT) while the window opens.
3. Left/right modifier handling is done by toggling the corresponding key's state when a `flagsChanged` comes in with no change in flags. This should instead be, "with no change in flags and the corresponding `modifierFlags` bit set," since if the mod bit is clear it doesn't make sense to report a key pressed event.

GHF:
Fixes #566.

- Initialize modifierFlags to user state when opening window
- Match flagsChanged keycode to corresponding bit in modifierFlags to
filter out unrelated events
- Only toggle key state if modifier bits are already set (not just if
they haven't changed)

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -168,6 +168,29 @@ static int translateKey(unsigned int key)
     return _glfw.ns.publicKeys[key];
 }
 
+// Translate a GLFW keycode to a mask for its GLFW modifier flag bit
+//
+static int translateKeyToFlagMask(int key)
+{
+    switch (key)
+    {
+        case GLFW_KEY_LEFT_SHIFT:
+        case GLFW_KEY_RIGHT_SHIFT:
+            return GLFW_MOD_SHIFT;
+        case GLFW_KEY_LEFT_CONTROL:
+        case GLFW_KEY_RIGHT_CONTROL:
+            return GLFW_MOD_CONTROL;
+        case GLFW_KEY_LEFT_ALT:
+        case GLFW_KEY_RIGHT_ALT:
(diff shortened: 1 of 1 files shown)


jaredjones:
Looks good! Great work man. :)

stevetranby:
Any idea when this will be merged into master?

elmindreda:
I cannibalized bits of this PR to remove the modifier flags cache altogether with 9f5658c8ac16c8da4c86fd5c11e3408d731c7a08. Thank you and sorry for taking so long to get to this one.

PR #580, opened 2015-08-13
x11_window.c: select may return EINTR, needs to be retried

Peaker:
When using glfw in Haskell in Linux via the Haskell bindings, the Haskell runtime system uses plenty of signals.

These signals cause glfwWaitForEvents to return immediately, because "select" in x11_window.c returns with EINTR. Then, XPending is called, finds nothing on the X file descriptor, returns 0, and glfw returns to the user without actually waiting for any event.

The fix is simply to loop while select returns EINTR.

### DIFF
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -62,6 +62,7 @@ typedef struct
 //
 void selectDisplayConnection(struct timeval* timeout)
 {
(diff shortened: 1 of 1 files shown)


elmindreda:
Thank you!

PR #583, opened 2015-08-23
Clear all buffer to black in glfwCreateWindow

jayschwa:
This eliminates flashing in OS X when running a barebones GLFW example app that doesn't do any rendering.

Newer users are sometimes confused by the flashing when they run the [GLFW.jl sample code](https://github.com/JuliaGL/GLFW.jl#simple-window-example), which is based off the [C version](http://www.glfw.org/documentation.html).

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -200,10 +200,10 @@ GLFWAPI GLFWwindow* glfwCreateWindow(int width, int height,
         return NULL;
     }
 
-    // Clearing the front buffer to black to avoid garbage pixels left over
-    // from previous uses of our bit of VRAM
+    // Clear the buffers to black to avoid left over garbage pixels in VRAM
     window->Clear(GL_COLOR_BUFFER_BIT);
     _glfwPlatformSwapBuffers(window);
+    window->Clear(GL_COLOR_BUFFER_BIT);
 
     // Restore the previously current context (or NULL)
     _glfwPlatformMakeContextCurrent(previous);


shurcooL:
Wasn't the back buffer already cleared and swapped to front before your change?

elmindreda:
Unable to reproduce on 10.10.5.

jayschwa:
> Wasn't the back buffer already cleared and swapped to front before your change?

Maybe I mixed up nomenclature. The "back" is what is currently being rendered to, and the "front" is what is being displayed to the screen, right? Either way, only one of the buffers has been cleared, leaving the other with junk. In a real program, this doesn't matter since something is going to be rendered, but it is evident in the simplest sample apps on certain machines.

> Unable to reproduce on 10.10.5.

I'm running 10.10.4 on a MacBook Pro (Retina, 15-inch, Early 2013) with NVIDIA GeForce GT 650M 1024 MB. This is the only machine I've used where VRAM isn't automatically cleared (I presume by the driver) before GLFW gets it. Never seen this on Windows or Linux.

elmindreda:
Ah, I see. You're not rendering but you're still swapping buffers?

elmindreda:
I'd suggest not swapping buffers in your first example. There's no inherent need to swap if you're not rendering anything and it might make more sense to introduce the function as "we've rendered our first very own things now and here is how we display them" instead of "here is a call you need to make because reasons". (I'm paraphrasing for effect, not criticising the comment in your example.)

jayschwa:
Fair enough. FYI, the template is basically a one-to-one copy of what is in the glfw.org documentation.

I guess the only other reason for doing this would be abstract "hygiene" reasons, even though it's really something the driver ought to do.

elmindreda:
> FYI, the template is basically a one-to-one copy of what is in the glfw.org documentation.

Guilty as charged! I never considered that `/* Render here */` would be a no-op.

> I guess the only other reason for doing this would be abstract "hygiene" reasons, even though it's really something the driver ought to do.

Yeah, it's a [serious security issue](http://www.ieee-security.org/TC/SP2014/papers/StealingWebpagesRenderedonYourBrowserbyExploitingGPUVulnerabilities.pdf). That said, I'll look into fixing it on our end. Of course, there are drivers that triple-buffer, etc.

elmindreda:
The [documentation page](http://www.glfw.org/documentation.html) example was updated for 3.2. This should eliminate this issue when copying the example verbatim.

elmindreda:
I don't believe this can be solved in a sane fashion by GLFW when the application does not update the window contents in response to window damage (i.e. the refresh callback).

Issue #427, opened 2015-01-21
Additional Cursors


imaginationcoin:
Please can you add definitions and support for the default cursors:

resize diagonal top left
resize diagonal bottom right
resize diagonal top right
resize diagonal bottom left
alt select
unavailable
hand grab
help select
hand grab
handwriting

elmindreda:
Those ones weren't added because they're not available on all platforms and at that point you're better of making your own.

imaginationcoin:
horizontal and vertical resize are standard, so the two diagonal resize, i think a couple of them are missing along with just a few other standard cursors. I would like the ability to offer the standard system cursors as well as custom ones.

elmindreda:
Here are the standard cursors available on each platform:

* [OS X](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSCursor_Class/index.html)
* [Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/ms648391.aspx)
* [X11](http://tronche.com/gui/x/xlib/appendix/b/)

ocornut:
I came here asking for the same question and found this. The lack of cursor on OS X is pretty unfortunate :( Perhaps is that it wouldn't hurt if glfw supported the definition for some of those extra cursors, with the documented fact that they won't be used on some platform. (the comment can be next to the definition).

(As for X11, It looks pretty clear that X11 cursors predate the jurassic era so perhaps they shouldn't be treated as the bottleneck.)

elmindreda:
I may squeeze in a few extra standard cursors by following the example set by SDL2. Since OS X has the most limited set but also doesn't (obviously) support cursor themes, I'll also consider adding custom images matching the existing standard cursors.

andrewrk:
Alternate approach: an API where you can set standard cursors including the ones in the top post here, but get an error code if the platform does not support the selected standard cursor. Then the application code could notice the error and set a custom cursor. That way GLFW does not have to include any art.

elmindreda:
@andrewrk That is the solution I want to avoid, as it is the least user friendly.

andrewrk:
I understand. Be careful though - trading power for user-friendliness in a C API is a dangerous road.

PlutoniumHeart:
I wouldn't throw an error, just ignore it in OS X and don't do anything, people use GLFW on OS X probably already know better than asking for things that don't exist any way.

wjakob:
[...]

PR #592, opened 2015-08-31
Add support for a wait/hourglass cursor

httpdigest:
This adds support for the wait/hourglass cursor with a Windows implementation using ICL_WAIT and using XC_watch on X11.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -684,6 +684,11 @@ extern "C" {
  *  The vertical resize arrow shape.
  */
 #define GLFW_VRESIZE_CURSOR         0x00036006
+/*! @brief The wait/hourglass shape.
+ *
+ *  The wait/hourglass shape.
+ */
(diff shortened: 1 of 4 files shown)


elmindreda:
This overlaps with #427.

httpdigest:
Oh, sorry. You are right.

httpdigest:
So, how to continue? Should we close this PR? When will the other PR make it into master?
It'd be really nice to have at least the hourglass cursor supported on Windows and Linux.

elmindreda:
Both issues pose the same basic question: should cursors be supported that are not provided by every platform? Once standard cursors are optional, they're less useful.

This issue poses an additional question: should a cursor be supported that encourages single-threaded applications?

elmindreda:
Since cursor themes aren't really a thing on OS X, the ones that are supported by the other platforms could be emulated with matching images, if someone volunteers to draw them. That would be doubly useful as OS X has the smallest set of built-in cursors.

PR #593, opened 2015-09-02
Travis Ci(unix and osx build) and AppVeyor(windows) badges in Readme.md

boddicheg:
Add badges like this:

[Build Status](https://travis-ci.org/boddicheg/glfw) [Build Status](https://ci.appveyor.com/project/boddicheg/glfw)

to Readme.md with current build status on osx, linux and windows.
If this pull request will be accepted, after every commit in master branch this badges will be updated according current build status.

[Example](https://github.com/boddicheg/glfw/tree/multi-platform)

P.S: both badges are clickable
Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,12 @@ language: cpp
 
 os:
 - linux
-- osx
 
 sudo: true
+
+before_install:
+- sh install-deps.sh
 
 install:
 - cmake .
 - make
-- sudo make install
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ os:
 
 sudo: true
 
-before_install:
+before_script:
 - sh install-deps.sh
 
-install:
+script:
 - cmake .
 - make
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: cpp
+
+os:
+- linux
+- osx
+
+sudo: true
+
+install:
+- cmake .
+- make
+- sudo make install
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: cpp
 
 os:
 - linux
+- osx
 
 sudo: true
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # GLFW
 [![Build Status](https://travis-ci.org/boddicheg/glfw.svg?branch=multi-platform)](https://travis-ci.org/boddicheg/glfw)
-
+[![Build Status](https://ci.appveyor.com/api/projects/status/github/boddicheg/glfw?branch=multi-platform&svg=true)](https://ci.appveyor.com/project/boddicheg/glfw)
 ## Introduction
 
 GLFW is a free, Open Source, multi-platform library for OpenGL and OpenGL ES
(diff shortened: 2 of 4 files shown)


shurcooL:
CI is pretty useful, nice job on making it work.

It's fairly basic in that it only tests that the code builds successfully (there are no automatic tests). But it's still useful to prove that the CMake setup and such is resulting in something than can build successfully.

install-deps.sh
```diff
@@ -0,0 +1,6 @@
+wget http://www.cmake.org/files/v2.8/cmake-2.8.12.tar.gz
+tar xzf cmake-2.8.12.tar.gz
+cd cmake-2.8.12
+cmake .
+make
+sudo make install
```
linkmauve: You probably don’t need to be root to install this package, just set its prefix to something like `~/.local`.

In Citra we download the binary instead, to make the build shorter, maybe you could take inspiration from this script: https://github.com/citra-emu/citra/blob/master/.travis-deps.sh

README.md
```diff
@@ -1,5 +1,6 @@
 # GLFW
-
+[![Build Status](https://travis-ci.org/boddicheg/glfw.svg?branch=multi-platform)](https://travis-ci.org/boddicheg/glfw)
+[![Build Status](https://ci.appveyor.com/api/projects/status/github/boddicheg/glfw?branch=multi-platform&svg=true)](https://ci.appveyor.com/project/boddicheg/glfw)
```
linkmauve: You should leave two blank lines here, to make this file more readable.

.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: cpp
+
+os:
+- linux
+- osx
+
+sudo: true
+
+before_script:
+- sh install-deps.sh
```
linkmauve: If you put the executable bit on this file, and a shebang, you don’t have to hardcode the interpreter to use.

.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+
+script:
+- cmake .
```
linkmauve: Why don’t you use the same method as appveyor, that is to build from a `build` directory?

linkmauve:
It would be even more useful to also run the tests, but I don’t think either of the providers you chose support that.

Aside from my few remarks, LGTM.

elmindreda:
The union of the changeset looks good but please squash it.

elmindreda:
Travis and Appveyor CI has been added.

shurcooL:
Nice!

I want to clarify what the travis CI test checks for. Is it only checking that the source code builds successfully without any errors? Does it also run any tests? Does it check against any runtime problems (like `glfwCreateWindow` failing to create a window)?

boddicheg:
@shurcooL 

> Is it only checking that the source code builds successfully without any errors?

Yep

>Does it also run any tests?

Nope, but if you add some script to ```script``` path to  .travis.yml file like this:

```
cmake tests/ 
```

so then it will.

> Does it check against any runtime problems (like glfwCreateWindow failing to create a window)?

I think it does but i am not sure.

elmindreda:
@shurcooL This tests compilation and linking only. No GLFW code is run, but this is still very helpful.

The current set of tests aren't suited for offline testing, but with this merged it's much more attractive to write ones that are. I don't believe most of GLFW can be usefully tested offline, but the bits that can should be.

Issue #232, opened 2014-02-10
Replace winmm with XInput and DirectInput


elmindreda:
[Start here](http://msdn.microsoft.com/en-us/library/windows/desktop/ee417014.aspx).

Sheado:
Hi,
Does this issue also cover the shared trigger axis problem for the Xbox controller? Or should a new issue be opened for that specifically?

elmindreda:
@Sheado Yes, it does.

theagentd:
Has this been implemented? We're having the shared trigger axis problem over here... >_<

Sonickidnextgen:
Hi, I just noticed this issue myself. I've been using GLFW for quite some time, in side projects and through [Monkey](https://github.com/blitz-research/monkey). If this is still an issue, I'd be happy to add XInput support. I've already written several dynamic-linking setups, similar to [what Dolphin does](https://github.com/dolphin-emu/dolphin/blob/49ea15b9502d98f08c47429ce93827073cefdaad/Source/Core/InputCommon/ControllerInterface/XInput/XInput.cpp#L77). Only, you know, it doesn't tend to break *as much*. I'm not very familiar with DirectInput, however. That being said, with both WinMM and XInput, what would be a benefit of supporting DirectInput? Also, what would be the best way to structure a change like this? Conditional support when compiling? Support WinMM as a fall-back?

Either way, I'm quite familiar with C and very familiar with C++. [This project](https://github.com/Regal-Internet-Brothers/IOSync/blob/1b70968e1fd85f031a9a159d2195899aeee73521/IOSync/src/devices/native/winnt/Real_XInput_Wrapper.cpp#L103), and [this project](https://github.com/Regal-Internet-Brothers/xinput/blob/bdd1590eea3bb8cefdcb58854c152f7915cb51e3/native/xinput.cpp#L303) both use function-pointers, 'LoadLibrary', and 'GetProcAddress'. They also support 'XInputGetStateEx' when available, meaning the guide button (Button 10) works.

If you're not interested, I understand. I'm also not the first guy to go to, where git is concerned.

theagentd:
Has there been any progress on this? It's really something we need to fix Xbox controllers and performance issues with the current controller polling.

elmindreda:
@theagentd I'm about halfway done and it should make it into 3.2.

theagentd:
Thanks for the update! Do you have a more precise ETA?

elmindreda:
@theagentd I don't right now.

[...]

Issue #57, opened 2013-06-17
Support for haptic feedback (vibration, rumble)


elmindreda:

thekitchensink:
Any update on this?

elmindreda:
@thekitchensink Not yet. I will post here when there is.

kevinmoran:
Just gonna throw my vote of support in here, this is pretty the only feature I'm missing in GLFW.

elmindreda:
Will start on this at some point after #900.

elmindreda:
All necessary joystick API transitions have been made. The way is clear to implement this now.

elmindreda:
#900 is done.

kevinmoran:
Awesome, looking forward to this feature but no rush

elmindreda:
Does anyone following this issue need a more complex interface than [XInputSetState](https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.reference.xinputsetstate.aspx) and if so why?

sir-pinecone:
Hey @elmindreda, do you happen to have an ETA on the haptic feedback support? I need to add this to my game soon so I'm just trying to decide if I should wait for your patch or code something myself. I already wrote some first pass code for this in a different project which I can probably retrofit into a fork of glfw. Thanks

elmindreda:
@sir-pinecone No ETA, sorry. Go ahead and roll your own for now.

JuKu:
@elmindreda Is there an update?

elmindreda:
Not right now. Finishing up 3.3, which has been repeatedly delayed due to Life happening.

XenoAmess:
@elmindreda Is there an update?

Perksey:
Please don't push this back again. Would love to see this for 3.4! Even if it's just a basic `glfwSetVibration` method.

WarrenMarshall:
Another upvote for this. This is literally the last thing I'm using SDL for ... would love to cut SDL loose.

XenoAmess:
> Another upvote for this. This is literally the last thing I'm using SDL for ... would love to cut SDL loose.

Sounds exactly the same situations you are facing to me.

carmiker:
I'll add another vote for this feature being added.

SuperFromND:
[...]

PR #615, opened 2015-10-17
Add possiblity to use XInput to handle josyticks

tikhomirov:
This is partial fix for #232. XInput documentation says:

> Use of legacy DirectInput is not recommended, and DirectInput is not available for Windows Store apps.

I thought that maybe it would be better to implement only XInput part and keep existing WinMM code as a fallback.If you still think that DirectInput required, please let me know and I'll start working on it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ option(GLFW_DOCUMENT_INTERNALS "Include internals in documentation" OFF)
 
 if (WIN32)
     option(GLFW_USE_HYBRID_HPG "Force use of high-performance GPU on hybrid systems" OFF)
+    option(GLFW_USE_XINPUT "Use XInput to handle josticks" ON)
 endif()
 
 if (APPLE)
@@ -209,6 +210,10 @@ if (_GLFW_WIN32)
 
     if (GLFW_USE_HYBRID_HPG)
         set(_GLFW_USE_HYBRID_HPG 1)
+    endif()
+
+    if (GLFW_USE_XINPUT)
+      set(_GLFW_USE_XINPUT 1)
     endif()
 endif()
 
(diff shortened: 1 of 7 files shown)


Niko113355:
I'm familier with this issue and tried to address it before.
The problem is implementing the fallback. I failed to find an easy way to identify which DirectInput devices also support XInput and thus skip them when iterating on directinput deivces to prevent duplicated events for that device.

sagaceilo:
What about pad vibration ? Using XInput gives us ability to add vibration to the pad (XInputSetState), and that would be desired by most of the game devs (I just came back from PGA - huge game exibition)  with no way to use pad vibration w/o rewriting pad system to use XInput and not glfw.

It'd be lovely to see some glfwSetJoystickVibrate soon :)

Niko113355:
I agree with you, I also think force feedback is a must and unfortunatly using XInput is the only way to enable it with XInput devices. I hope too there will be a fix for this soon.

tikhomirov:
@sagaceilo @Niko113355  there is separate issue for this – #57

tikhomirov:
@Niko113355  Also it's not a problem to detect XInput devices while iterating DirectInput devices, look at #232. On my opinion we can have XInput as main gamepad API fro Win32 with WinMM as an option for those who want to support old devices so we don't need to add support for DirectInput which is deprecated according to XInput documentation. But if @elmindreda will say that we definitely need DirectInput I'll continue my work :)

elmindreda:
I already have XInput up and running (since before 3.1.2) and am finishing up DirectInput support.

tikhomirov:
okay, so I'm closing this pull request then

elmindreda:
@tikhomirov Thank you for the pull request, though! If I hadn't already written something very similar, this would have been a solid foundation!

Since you're all here, does anyone _not_ want GLFW to handle the dead zone for you?

Niko113355:
@elmindreda I think it should be handled and set to a reasonable default value, with of course the possibility to tweak it.

elmindreda:
@Niko113355 Please elaborate on "possibility to tweak".

Niko113355:
have a function such as glfwSetDeadZoneThreshold receving a float value from 0.0 to 1.0.
0 being no input
1 being 100% input
when one of the axis passes the threshold only then will the input be recived and sent to a callback function.
And then giving a default value of about 0.05.
Another enhancment might be split the X and Y fot sticks, giving each one a different threshold to provide more control.

sagaceilo:
I think that the very basic dead zone should be handled by glfw, since its implementation defined as Macro. Other platforms should have something similar. But futher tweaks on axis values must be on the client side. Its super easy to add value clamping/renormalizing and its unnecessary logic and API bloat.
glfw by its design have to provide consistent experience and values from expected ranges that are fundation to your app that will change it as you wish.

Issue #43, opened 2013-06-13
Window mode switching


elmindreda:

theagentd:
Is there an ETA on this? It's an important feature for us.

elmindreda:
@theagentd I'm sorry, I can't give an ETA, but I will help if I can. What is the time frame of your need? I'm about to release 3.1.2 and this is one of the first things I will tackle after that.

theagentd:
It's not extremely urgent, but within one or two months would be great so we can bug-test and integrate the functionality into our engine and menu system. We're working on some big rewrites/library switched right now.

CreamyCookie:
@elmindreda Does this restore the window on the monitor it was previously on? I'm currently having a problem in LibGDX where after going from fullscreen to windowed, the window ends up on the primary monitor instead of the one it was before.

elmindreda:
@CreamyCookie It places the window where you specify.

CreamyCookie:
@elmindreda Oh.. true. I should have seen that. It depends on the xpos and ypos, right?

And LibGDX does this when going from fullscreen to windowed.

```java
Lwjgl3DisplayMode currentMode = (Lwjgl3DisplayMode) getDisplayMode();
int windowPosX = (currentMode.width - width) / 2;
int windowPosY = (currentMode.height - height) / 2;
GLFW.glfwSetWindowMonitor(window.getWindowHandle(), 0,
		windowPosX, windowPosY, width, height, currentMode.refreshRate);
```

Now, I think they did that because they didn't know what the windowPosX (and Y) was before it was set to fullscreen.. Is there a way to get that from GLFW? Or should they just store the position it was before setting it to fullscreen?

elmindreda:
@CreamyCookie You need to save it if you want to restore it.

CreamyCookie:
@elmindreda Thank you!

PR #618, opened 2015-10-20
Add option to toggle between windowed mode and fullscreen without closing and creating windows

Niko113355:
~~When autoIconify is enabled the window attempts to enter fullscreen mode twice. The reason is explained in comments in the area of trouble.~~

~~Also consider seperating the function _glfwPlatformShowWindow (win32_window.c:1035)
to to different functions. The name ShowWindow does not necessarily imply the grab of focus, I think the name of the function should change or the code should be split entirely.~~

### DIFF
--- a/src/internal.h
+++ b/src/internal.h
@@ -178,7 +178,7 @@ struct _GLFWwndconfig
     GLFWbool      focused;
     GLFWbool      autoIconify;
     GLFWbool      floating;
-	GLFWbool      fullscreen;
+    GLFWbool      fullscreen;
     _GLFWmonitor* monitor;
 };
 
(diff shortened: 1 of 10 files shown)


Niko113355:
Added a possible implementation on windows to enable toggeling fullscreen mode of an existing window using the glfwToggleWindowFullscreen. This is a more intuitive way than destroying the window entirely and creating a new one in the desired mode.

I must add that this implementation breaks backward compatibility of GLFW due to the addition of a new window flag GLFW_FULLSCREEN which is now used to identify if a window is a fullscreen window or a windowed mode window. The previous way of checking by comparing the result of glfwGetWindowMonitor to null no longer works.

elmindreda:
See #43.

src/internal.h
```diff
@@ -178,6 +178,7 @@ struct _GLFWwndconfig
     GLFWbool      focused;
     GLFWbool      autoIconify;
     GLFWbool      floating;
+	GLFWbool      fullscreen;
```
Spartan322: just like to point out it appears there is a hard tab here. (Nit picking, opps)

PR #620, opened 2015-10-22
Implemented glfwCreateStandardCursor in the Wayland platform.

ricardomv:
Tested on weston and gnome-wayland

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -207,6 +207,28 @@ createAnonymousFile(off_t size)
     return fd;
 }
 
+// Translates a GLFW standard cursor to a theme cursor name
+//
+static char *translateCursorShape(int shape)
+{
+    switch (shape)
+    {
+        case GLFW_ARROW_CURSOR:
+            return "left_ptr";
+        case GLFW_IBEAM_CURSOR:
+            return "xterm";
+        case GLFW_CROSSHAIR_CURSOR:
+            return "crosshair";
+        case GLFW_HAND_CURSOR:
+            return "grabbing";
+        case GLFW_HRESIZE_CURSOR:
+            return "sb_h_double_arrow";
+        case GLFW_VRESIZE_CURSOR:
+            return "sb_v_double_arrow";
+    }
+    return NULL;
+}
+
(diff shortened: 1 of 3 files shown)


src/wl_window.c
```diff
@@ -469,13 +491,25 @@ int _glfwPlatformCreateCursor(_GLFWcursor* cursor,
...
 {
-    // TODO
-    fprintf(stderr, "_glfwPlatformCreateStandardCursor not implemented yet\n");
-    return GLFW_FALSE;
+    struct wl_cursor* standard_cursor;
+
+    standard_cursor = wl_cursor_theme_get_cursor(_glfw.wl.cursorTheme,
+                                                 translateCursorShape(shape));
+    if (!standard_cursor) {
+        fprintf(stderr, "Wayland: Standard cursor not found\n");
```
linkmauve: Use `_glfwInputError(GLFW_PLATFORM_ERROR, "…")` instead of fprintf, to give the message to the application instead.

src/wl_window.c
```diff
@@ -498,8 +532,12 @@ void _glfwPlatformSetCursor(_GLFWwindow* window, _GLFWcursor* cursor)
     if (window->cursorMode == GLFW_CURSOR_NORMAL)
     {
         if (cursor == NULL)
-        {
-            image = _glfw.wl.defaultCursor->images[0];
+            image = wl_cursor_theme_get_cursor(_glfw.wl.cursorTheme,
+                                               "left_ptr")->images[0];
```
linkmauve: This can crash in the (unlikely) case where the current theme doesn’t have a `left_ptr` cursor, but better take care of this case and abort with an error.

linkmauve:
Aside from those two nitpicks, LGTM. :)

ricardomv:
@linkmauve thanks for the tips, now fixed.

ricardomv:
also fixed some coding style issues

elmindreda:
Thank you!

Fixed a few remaining style issues with efdbfd38834ee86feb12e86e476c4e6fefb62b14.

PR #627, opened 2015-10-28
Don't delete the POSIX TLS key if it hasn't been created yet

SpecLad:
0 (the initial value of `context`) is a valid TLS key, so doing this can delete someone else's key.


elmindreda:
Thank you!

PR #633, opened 2015-11-02
Wayland: Create the surface before the context

linkmauve:
This fixes a regression introduced in 496f559c where a context would be created for a NULL egl_surface, which subsequently fails.


ricardomv:
LGTM :+1:

PR #639, opened 2015-11-10
Fix for MinGW compilation

podgorskiy:
The issue:

    libs/glfw/src/libglfw3.a(win32_window.c.obj):win32_window.c:(.text+0xa6e): undefined reference to `GET_XBUTTON_WPARAM'
    collect2.exe: error: ld returned 1 exit status


elmindreda:
Thank you!

Issue #266, opened 2014-04-04
Explicit maximization support


retep998:
A maximized window is treated differently from a normal window on many platforms (most notable being Windows).
Having an API to determine whether a window is maximized, and to change whether or not it is maximized would be very useful.

andrewrk:
This is one of those issues that needs to be solved for GLFW to be a good choice for tool writers.

elmindreda:
Moving this to 3.3 because I want to put out a release including Vulkan support as soon as possible.

The blocking issues for this feature are:
- [ ] Should it use zoom or Lion full screen on OS X?
- [ ] Is it possible to have a newly created window enter this state before being shown?
- [ ] Should un-maximization be done with `glfwRestoreWindow` or a new function?
- [ ] Should there be an event callback?

shurcooL:
> Should it use zoom or Lion full screen on OS X?

I think they are 2 separate modes.

Zoom is most like the Windows maximize. Lion full screen is more like a way of switching to a borderless fullscreen window.

elmindreda:
@shurcooL True, but Lion full screen is what currently happens when you click the analog of the maximization button for GLFW windows and many other OS X applications.

shurcooL:
> but Lion full screen is what currently happens when you click the analog of the maximization button for GLFW windows and many other OS X applications.

Yeah. It's still possible to have the old zoom behavior by Option-Clicking (holding Option/Alt key, then Left Clicking with mouse). But normal click is now indeed mapped to fullscreen.

There's no ideal mapping to Windows.

elmindreda:
All right, zoom it is then.

elmindreda:
One item remaining. Unless someone objects I'll go with the results of [this poll](https://twitter.com/elmindreda/status/677125720410357760).

PR #641, opened 2015-11-11
Make the window auxiliary

elvman:
As OSX fullscreen (by pressing the green button) is not supported by GLFW (windowDidEnterFullScreen: and windowDidExitFullScreen: callbacks), I suggest making the GLFW window auxiliary (that forbids user to enter the fullscreen manually) and allow entering fullscreen only programmatically (as it is on other platforms).

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -844,7 +844,7 @@ static GLFWbool createWindow(_GLFWwindow* window,
     }
 
     if (wndconfig->resizable)
-        [window->ns.object setCollectionBehavior:NSWindowCollectionBehaviorFullScreenPrimary];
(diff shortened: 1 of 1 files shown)


elmindreda:
Lion full screen is more akin to maximization, which is enabled on other platforms and will get support with #266.

shurcooL:
> I suggest making the GLFW window auxiliary (that forbids user to enter the fullscreen manually) and allow entering fullscreen only programmatically (as it is on other platforms).

elmindreda:
> I don't agree with this being an improvement.

To be clear, I don't either. Threads like this can still be fruitful, though, so I won't close it just yet.

elmindreda:
@elvman Would maximization functionality along the lines of the current support for iconification count?

PR #645, opened 2015-11-16
Make sure the cursor’s buffer is valid before destroying it

linkmauve:
The shape should never be an unknown value, but it is imo still better to check it really is. Another way would be to `abort()` in `translateCursorShape()`, which I find less elegant than this one.

The other change is there to handle the case where the cursor creation failed, so we don’t buffer_destroy NULL.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -522,7 +522,8 @@ void _glfwPlatformDestroyCursor(_GLFWcursor* cursor)
     if (cursor->wl.image)
         return;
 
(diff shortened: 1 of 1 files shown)


elmindreda:
The test in `glfwPlatformDestroyCursor` is good, but `glfwCreateStandardCursor` already verifies that the shape value is valid.

linkmauve:
I removed the first test, and changed the commit message.

elmindreda:
Thank you!

PR #651, opened 2015-11-29
Fix for key events filtered by Win32 IME being passed to end-user

ocornut:
This is a proposed fix for the issue discussed in https://github.com/glfw/glfw/issues/41#issuecomment-160421273


elmindreda:
Thank you. Will review.

elmindreda:
Looks good.

PR #660, opened 2015-12-14
Implement HiDPI support on Wayland

linkmauve:
Windows now keep track of the monitors they are on, so we can calculate
the best scaling factor for them, by using the maximum of each of the
monitors.

The compositor scales down the buffer automatically when it is on a
lower density monitor, instead of the previous way where it was scaling
up the buffer on higher density monitors, which makes the application
look much better on those ones.

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -381,7 +381,8 @@ static void registryHandleGlobal(void* data,
     if (strcmp(interface, "wl_compositor") == 0)
     {
         _glfw.wl.compositor =
-            wl_registry_bind(registry, name, &wl_compositor_interface, 1);
+            wl_registry_bind(registry, name, &wl_compositor_interface,
+                             min(3, version));
     }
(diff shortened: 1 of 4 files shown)


linkmauve:
Pushed a newer version, addressing all of @jadahl’s comments.

jadahl:
I just pushed the commit with two changes:
 - you used min(x1, y1) which doesn't exist. I defined in wl_init.c for now.
 - I added monitor->wl.scale = 1; to monitor creation as so is specified by the spec (no .scale event == scale is 1). Not doing this would not cause a bug since you always set the buffer scale to at least 1, but so changed just to do what the spec says.

I also added a related extra commit doing version checks on wl_compositor so we don't try to set the buffer scale if it's not supported. This is very unlikely to ever to happen since compositors will probably implement both these features at the same time, so it was more or less only for correctness, not as much for practical reasons.

PR #667, opened 2015-12-18
Fast window position updates

ckaran:
Rewrote `glfwPlatformGetWindowPos()` in `src/cocoa_window.m` to use code from Matt Reagan on the Cocoa-dev mailing list to determine the position of a window in real time.  This bypasses the window notification problems under OS X that made window position updates not be updated until after a window was no longer being dragged.  At this point window updates are almost perfectly realtime - there is a little bit of stuttering going on under certain circumstances, but i don't yet know if that's due to my test code, or if there is something else going on.


ckaran:
@andsve I've updated the code again, this time to the `fast_window_position_updates` branch.  This is really just an OS X thing, so if you're not interested in OS X, you can ignore that branch, but I thought you might like to know about it.

PR #678, opened 2015-12-30
Fix SetProcessDpiAwareness case mismatch

Spasi:
This patch corrects the [SetProcessDpiAwareness](https://msdn.microsoft.com/en-us/library/windows/desktop/dn302122(v=vs.85).aspx) function name.


elmindreda:
Thank you!

PR #683, opened 2016-01-05
Update egl_context.c

yousry:
- Knowing the thread id greatly simplifies debugging.
- The return value of eglMakeCurrent should be checked.

### DIFF
--- a/src/egl_context.c
+++ b/src/egl_context.c
@@ -30,7 +30,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <assert.h>
-
+#include <sys/types.h>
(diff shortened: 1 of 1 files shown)


elmindreda:
There is no part of this that I want to merge.

andrewrk:
@elmindreda I agree that this PR is not worth merging, but shouldn't we be checking the return value of `eglMakeCurrent` and panicking or calling the error handler if it is not successful?

elmindreda:
@andrewrk Definitely. I'm currently reading the EGL 1.5 spec end-to-end in preparation for an overhaul of that backend.

PR #697, opened 2016-02-01
Add GLFW includes to target_include_directories

mbitsnbites:
This makes it easier to build GLFW as a sub-project of another CMake based project. Instead of having to use include_directories(path/to/glfw/include) the project using GLFW will get the correct include path by just doing target_link_libraries(myapp glfw ${GLFW_LIBRARIES}).

Note: I'm still learning CMake, so I hope that this is the correct solution.

@elmindreda - Does this look OK? This would also change the documentation: In http://www.glfw.org/docs/latest/build.html#build_link_cmake_source, the step about "To be able to include the GLFW header from your code..." can be removed.

### DIFF
--- a/docs/build.dox
+++ b/docs/build.dox
@@ -158,13 +158,6 @@ will add the `glfw` target and the necessary cache variables to your project.
 add_subdirectory(path/to/glfw)
 @endcode
 
-To be able to include the GLFW header from your code, you need to tell the
-compiler where to find it.
-
-@code{.cmake}
-include_directories(path/to/glfw/include)
-@endcode
-
 Once GLFW has been added to the project, the `GLFW_LIBRARIES` cache variable
 contains all link-time dependencies of GLFW as it is currently configured.  To
 link against GLFW, link against them and the `glfw` target.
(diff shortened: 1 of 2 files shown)


mbitsnbites:
I found and updated the documentation too. The commits can be squashed before merge.

tamaskenez:
I'm sorry, this is wrong. Also, [shaxbee's fork](https://github.com/shaxbee/glfw) (see #460) which has been waiting for inclusion for a long time solves this and other CMake issues.
You can check [this line](https://github.com/shaxbee/glfw/blob/master/src/CMakeLists.txt#L55) about how  `target_include_directories` should be used.

elmindreda:
Yes, this requires some generator expressions to work properly, and the corresponding lines in `examples/CMakeLists.txt` and `tests/CMakeLists.txt` should be removed, but unlike #460 it's blessedly clean.

I'll take this and the bits I like of #460 and thank you @mbitsnbites for providing a much needed ray of light.

elmindreda:
This has been merged, together with a good chunk of #460. Now you don't need `GLFW_LIBRARIES` either.

adsk-mage:
Thank's alot @elmindreda ! I actually haven't used GLFW for ages, but now we're picking it up for an in house test application. It's one of those dependencies that you're really happy to include in a project (clean build on all platforms with all compilers, zero extra dependencies, and taking only 0.2s to build!).

elmindreda:
@adsk-mage Well, I did inherit a good foundation. Cool to see it being used inside Autodesk!

Issue #694, opened 2016-01-26
Presence of game pad causes segmentation fault (and proposed fix)


jtdaugherty:
I am experiencing a segmentation fault inside of GLFW version `3.1.2`. All that is required to trigger this for me is to:
 * Connect one of my game pads (http://www.amazon.com/Buffalo-Classic-USB-Gamepad-PC/dp/B002B9XB0E)
 * Call `glfwInit()`

I believe the problem is that `src/cocoa_joystick.m` line 280 is not checking for a possible `NULL` return value from `IOHIDDeviceGetProperty`. Then `NULL` is passed to `CFStringGetCString` which crashes. I've found reports of other applications crashing with this symptom when `NULL` is passed to that function. When I unplug the game pad, `glfwInit()` runs successfully.

I'd submit a pull request but I can't test this first since I can't get glfw3 to link on my machine.

For completeness, here is the call stack:
```
Process 3963 stopped
* thread #1: tid = 0x890a, 0x00007fff8b29587c CoreFoundation`CFStringGetCString + 44, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00007fff8b29587c CoreFoundation`CFStringGetCString + 44
CoreFoundation`CFStringGetCString:
->  0x7fff8b29587c <+44>: movq   (%r12), %rax
    0x7fff8b295880 <+48>: testq  %rax, %rax
    0x7fff8b295883 <+51>: je     0x7fff8b2958c4            ; <+116>
    0x7fff8b295885 <+53>: leaq   -0x13d4bba4(%rip), %rcx   ; __CFConstantStringClassReferencePtr
(lldb) bt
* thread #1: tid = 0x890a, 0x00007fff8b29587c CoreFoundation`CFStringGetCString + 44, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00007fff8b29587c CoreFoundation`CFStringGetCString + 44
    frame #1: 0x0000000100011f86 libglfw3.3.dylib`matchCallback + 148
    frame #2: 0x00007fff8a848c21 IOKit`__IOHIDManagerDeviceApplier + 153
    frame #3: 0x00007fff8b2c0372 CoreFoundation`__CFSetApplyFunction_block_invoke + 18
    frame #4: 0x00007fff8b2a4f80 CoreFoundation`CFBasicHashApply + 128
[...]

PR #701, opened 2016-02-08
Fix crash when joystick name is unavailable

aaronmjacobs:
Fixes issue #694

I've been unable to test this locally (since I don't have any joysticks / controllers that don't report a name), but it's a pretty simple fix. This matches the behavior found in linux_joystick.c


elmindreda:
Thank you!

PR #715, opened 2016-02-22
support for X11/GLX window transparency

datenwolf:
Added support for transparent X11/GLX windows.

### DIFF
--- a/examples/gears.c
+++ b/examples/gears.c
@@ -172,6 +172,7 @@ static GLfloat angle = 0.f;
 /* OpenGL draw function & timing */
 static void draw(void)
 {
+  glClearColor(0., 0., 0., 0.);
   glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
   glPushMatrix();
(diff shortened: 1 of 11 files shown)


elmindreda:
Thanks! I'll start a target branch for future pull requests based on this.

datenwolf:
Just FYI: As obvious by the commit message I also took care about the EGL codepath. Apparently Mesa (at least for the Intel integrated graphics in my notebook) will not report visuals with picture format that has a nonzero alpha mask. _However it works flawlessly with NVidia proprietary drivers!_ IMHO this is something that absolutely must be addressed by Mesa, so someone should file a feature request (GLFW now provides easy to use test cases).

christopherp:
I also have been working on this for the past couple of days and support all three [X11, OSX and Windows](https://github.com/christopherp/glfw/tree/transparent-window). Unlike @datenwolf I didn't add support for the EGL codepath.

christopherp:
I can remove X11 support from my fork to not conflict with @datenwolf . Let me know @elmindreda .

christopherp:
Also, I believe the correct way to implement the `GLFW_TRANSPARENT` hint is to pass the value to `_glfw.hints.framebuffer.transparent` since it is used in places where the fbconfig is needed then we don't need to change function signatures in different places.

Also, since we need to know `window.transparent` in OSX and for when the client query `glfwGetWindowAttrib` , it should be set in the definition of `glfwCreateWindow` after `_glfwIsValidContextConfig` is called [window->transparent = fbconfig.transparent;](https://github.com/christopherp/glfw/commit/3360b933105f68e237d486511532dc9c36900401#diff-b18e3b188ecb005e52e769a4e906a845R184) or else `window.transparent` is always true even it isn't available.

elmindreda:
This has been merged to the [transparent-window](https://github.com/glfw/glfw/tree/transparent-window) branch unmodified except for [commit message fixes](https://github.com/erlang/otp/wiki/writing-good-commit-messages). Thank you!

PR #716, opened 2016-02-24
Add accelerated transparent window support for Windows Vista and up

christopherp:
Because `DwmEnableBlurBehindWindow` is supported since Vista, previous versions will simply get an opaque window.

It currently doesn't support setting a decorated window at the same time. If the decorated window hint is set to true, we just return an opaque window. It probably isn't a big issue since transparent windows are usually used for overlays or unconventional window shapes. This issue could actually be fixed by adding support for Windows XP and choose the non accelerated path instead, when decoration is required. A solution for accelerated transparent window with decoration probably does exist but I haven't been able to find it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ Debug
 Release
 MinSizeRel
 RelWithDebInfo
+*.VC.opendb
 
 # CMake files
 Makefile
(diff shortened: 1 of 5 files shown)


christopherp:
Whoops I just realized I forgot to filter the pixelformat for composition. I am closing this PR.

PR #723, opened 2016-03-06
Add support for wgl window transparency (Vista and later)

christopherp:
This method uses desktop composition for efficient window transparency.

Since windows 8, composition is always on. If composition is disabled on vista or 7, transparency is simply ignored. Also, since it requires a pixel format supporting compositing, if we can't find it (e.g. windows xp) we ignore transparency similarly to how @datenwolf is doing it for glx's frame buffer configs.

This shouldn't be tested in a virtual machine.
On a side note, desktop composition can be enabled in the "Performance Options" dialog (win+r systempropertiesperformance). It is on by default but won't work unless the Aero score index has been refreshed after installing gpu drivers (Control Panel\All Control Panel Items\Performance Information and Tools).

### DIFF
--- a/src/wgl_context.c
+++ b/src/wgl_context.c
@@ -493,6 +493,28 @@ GLFWbool _glfwCreateContextWGL(_GLFWwindow* window,
         }
     }
 
+    if (window->transparent) {
+        if (!isCompositionEnabled || !_glfw_DwmEnableBlurBehindWindow) {
+            window->transparent = GLFW_FALSE;
+        }
+        else
+        {
+            HRESULT hr = S_OK;
+
+            DWM_BLURBEHIND bb = { 0 };
+            bb.dwFlags = DWM_BB_ENABLE | DWM_BB_BLURREGION;
+            bb.hRgnBlur = CreateRectRgn(0, 0, -1, -1);  // makes the window transparent
+            bb.fEnable = TRUE;
+            hr = _glfw_DwmEnableBlurBehindWindow(window->win32.handle, &bb);
+
+            if (!SUCCEEDED(hr)) {
+                window->transparent = GLFW_FALSE;
+                _glfwInputError(GLFW_PLATFORM_ERROR,
+                    "WGL: Failed to enable blur behind window required for transparency");
+            }
+        }
+    }
+
     return GLFW_TRUE;
 }
 
--- a/src/wgl_context.c
+++ b/src/wgl_context.c
@@ -133,6 +133,20 @@ static GLFWbool choosePixelFormat(_GLFWwindow* window,
     {
         const int n = i + 1;
         _GLFWfbconfig* u = usableConfigs + usableCount;
+        PIXELFORMATDESCRIPTOR pfd;
+
+        if (window->transparent) {
+            if (!DescribePixelFormat(window->context.wgl.dc,
+                n,
+                sizeof(PIXELFORMATDESCRIPTOR),
+                &pfd))
+            {
+                continue;
+            }
+
+            if (!(pfd.dwFlags & PFD_SUPPORT_COMPOSITION))
+                continue;
+        }
 
         if (_glfw.wgl.ARB_pixel_format)
         {
@@ -188,11 +202,9 @@ static GLFWbool choosePixelFormat(_GLFWwindow* window,
         }
         else
         {
-            PIXELFORMATDESCRIPTOR pfd;
-
             // Get pixel format attributes through legacy PFDs
 
-            if (!DescribePixelFormat(window->context.wgl.dc,
+            if (!window->transparent && !DescribePixelFormat(window->context.wgl.dc,
                                      n,
                                      sizeof(PIXELFORMATDESCRIPTOR),
                                      &pfd))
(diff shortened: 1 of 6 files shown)


christopherp:
I tested it on Windows Vista, 7, 8, 8.1 and 10.

It ins't so straight forward to achieve transparency with opengl on Windows but I hope that you will find that this is heading to the right direction. This method has quirks which can all be resolved with hacks but by comparison to other methods, this method is the fastest. The glReadPixels or BitBlt methods to achieve transparency use a lot of cpu when fullscreen vs almost idle with Dwm (bellow 1%).

It is worth also noting that there is a color key method on layered windows. This would require a different type of glfw hint because all pixels painted by the window in this color will be transparent and forces the user to not use that color in his rendering.

elmindreda:
Thank you!

src/wgl_context.c
```diff
@@ -325,6 +346,101 @@ void _glfwTerminateWGL(void)
...
+    // Decorated windows on Windows 8+ don't repaint the transparent background
+    // leaving a trail behind animations.
+    // Hack: making the window layered with a transparency color key seems to fix this.
+    // Normally, when specifying a transparency color key to be used when composing
+    // the layered window, all pixels painted by the window in this color will be transparent.
+    // That doesn't seem to be the case anymore on Windows 8+, at least when used with
+    // DwmEnableBlurBehindWindow + negative region.
+    if (window->decorated && isWindows8OrGreater())
+    {
+        long style = GetWindowLong(handle, GWL_EXSTYLE);
```
uglycoder: ' To write code that is compatible with both 32-bit and 64-bit versions of Windows, use GetWindowLongPtr'
https://msdn.microsoft.com/en-us/library/windows/desktop/ms633584(v=vs.85).aspx
...
Curiously SetWindowLongPtr is used further down.

elmindreda: `GWL_EXSTYLE` is not a pointer or a handle, so it's fine. I'd prefer it to call `GetWindowLongW`, however.

uglycoder: Line 419: SetWindowLongPtr(handle, GWL_EXSTYLE, style))
This is also fine?
When compiling for 32-bit Windows, GetWindowLongPtr is defined as a call to the GetWindowLong function.
Why not abide by the MS's recommendation? Fully?
Regarding the W. Yes, new code should be explicitly Unicode for Windows API.

elmindreda: > Line 419: SetWindowLongPtr(handle, GWL_EXSTYLE, style))
> This is also fine?

Well, it works. I'd change it to `GetWindowLongW`.

Microsoft's recommendation applies to pointers and handles.

I'll be cleaning up all of this before merging it to master.

uglycoder: You are correct of course about  the MS's recommendation. My mistake.
But SetWindowLongPtr works as intended.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms644898(v=vs.85).aspx
'The function also sets a value at the specified offset in the extra window memory'
Same for GetWindowLongPtr.
I'm just in the habit of using these consistently for this Window manipulation vis-à-vis 32/64bit issues.
You will have the last word.

elmindreda: Yup, it's a matter of taste. Your way works just as well and has the advantage of being a nod to people that you have thought about 64-bit.

christopherp:
@uglycoder good catch! I did purposely use SetWindowLongPtr but I overlooked the Get version.
A quick search shows that we previously used GetWindowLongPtrW in the windowProc callback to retrieve the _GLFWwindow pointer. For the sake of consistency, maybe we could use `GetWindowLongPtrW` and `SetWindowLongPtrW` everywhere.

src/wgl_context.c
christopherp: @uglycoder good catch! I did purposely use SetWindowLongPtr but I overlooked the Get version.
A quick search shows that we previously used GetWindowLongPtrW in the windowProc callback to retrieve the _GLFWwindow pointer. For the sake of consistency, maybe we could use GetWindowLongPtrW and SetWindowLongPtrW everywhere.

elmindreda: @uglycoder @christopherp This rule is simpler. Fixed with 43f116fa40d5147d66c6392e1eb5690c7497dd93.

thomthom:
I checked out 43f116f and used `cmake -G "Visual Studio 12 2013 Win64" ..` to create the solution and then tried to build with VS2013 - got these errors:

image

Anyone else sees this? Or have I borked my build?

thomthom:
I removed the `inline` keyword of `isWindowsVersionOrGreater` and then it built fine. I removed it based on a hint from ReSharper++.

image

Once built it's working fine on my Win10 machine: (nVidia GeForce GTX 750)

image

I'll try on Win7 later.

src/wgl_context.c
dedred: Hello,

I've suddenly came sccross that information and thought it may be useful for you, read more here please <http://nkotrengynto.freestuffestore.com/aehqulo>

Typos courtesy of my iPhone, acreator@blueyonder.co.uk

PR #729, opened 2016-03-23
Remove unnecessary joystick polling

IntellectualKitty:
Removed unnecessary joystick event polling on OS X.  Specifically, polling was unnecessary when getting the joystick name or when determining if a joystick is present.  And the pollJoystickEvents function was split into pollJoystickAxisEvents and pollJoystickButtonEvents to prevent unnecessary polling for the opposite element type.

### DIFF
--- a/src/cocoa_joystick.m
+++ b/src/cocoa_joystick.m
@@ -192,9 +192,35 @@ static void removeJoystick(_GLFWjoydeviceNS* joystick)
     memset(joystick, 0, sizeof(_GLFWjoydeviceNS));
 }
 
-// Polls for joystick events and updates GLFW state
+// Polls for joystick axis events and updates GLFW state
 //
-static GLFWbool pollJoystickEvents(_GLFWjoydeviceNS* joystick)
+static GLFWbool pollJoystickAxisEvents(_GLFWjoydeviceNS* joystick)
+{
+    CFIndex i;
+
+    if (!joystick->present)
+        return GLFW_FALSE;
+
+    for (i = 0;  i < CFArrayGetCount(joystick->axisElements);  i++)
+    {
+        _GLFWjoyelementNS* axis = (_GLFWjoyelementNS*)
+            CFArrayGetValueAtIndex(joystick->axisElements, i);
+
+        long value = getElementValue(joystick, axis);
+        long readScale = axis->maxReport - axis->minReport;
+
+        if (readScale == 0)
+            joystick->axes[i] = value;
+        else
+            joystick->axes[i] = (2.f * (value - axis->minReport) / readScale) - 1.f;
+    }
+
+    return GLFW_TRUE;
+}
+
+// Polls for joystick button events and updates GLFW state
+//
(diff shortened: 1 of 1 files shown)


IntellectualKitty:
I hope that this was helpful.  Also, this was my first attempt at doing anything more than simple commenting on GitHub, so I hope that I did the pull request correctly.

I thought it would more helpful to write the code and make a pull request rather than just raising it as an issue, but if you have a preference please let me know.

src/cocoa_joystick.m
```diff
@@ -496,9 +508,9 @@ int _glfwPlatformJoystickPresent(int joy)
 const char* _glfwPlatformGetJoystickName(int joy)
 {
     _GLFWjoydeviceNS* joystick = _glfw.ns_js.devices + joy;
-    if (!pollJoystickEvents(joystick))
+    if (joystick->present)
+        return joystick->name;
+    else
```
elmindreda: Please follow the pattern above of checking failure cases and then successful return at the end.

src/cocoa_joystick.m
```diff
@@ -469,13 +481,13 @@ void _glfwTerminateJoysticksNS(void)
 int _glfwPlatformJoystickPresent(int joy)
 {
     _GLFWjoydeviceNS* joystick = _glfw.ns_js.devices + joy;
-    return pollJoystickEvents(joystick);
+    return joystick->present ? GLFW_TRUE : GLFW_FALSE;
```
elmindreda: `joystick->present` is already `GLFW_TRUE` or `GLFW_FALSE`.

elmindreda:
It definitely was! I've commented on two minor issues, but everything related to GitHub looks good.

Don't really have a preference. Either way is fine.

src/cocoa_joystick.m
IntellectualKitty: I assumed that was the case, but since the code in cocoa_joystick.m typically evaluated joystick->present as a compiler-time boolean (e.g., if (joystick->present)) I thought it was safer to be explicit since a user might evaluate the return value specifically with GLFW_TRUE.

Coming from a C++ background, I exclusively use enumerated values rather than #defines or constants since the compiler will automatically restrict variables to their enumerated values (unless you explicitly typecast, after which I always include an assert statement to confirm that the variable is still within the limits of the enumeration); also, there are additional benefits such as the compiler warning you if any enumerated values are missing within a switch statement.  However, when I program in C, I try to be more careful and explicit.

src/cocoa_joystick.m
IntellectualKitty: When a test case is most likely true as it is here, I like to handle the successful case first because it's generally more efficient: (1) it avoids the cost of the branch and (2) it avoids the cost of loading the instruction cache when large instruction blocks follow the condition.  That's a fairly ingrained habit, and I failed to notice that the code followed the pattern of handling failure cases first.

IntellectualKitty:
I'm glad this was helpful.

elmindreda:
Thank you!

PR #735, opened 2016-04-09
Misc Wayland fixes

linkmauve:
This implements mostly everything still missing and implementable in the Wayland backend from 3.2.

The last remaining implementable item is clipboard support, but I don’t know much about this.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -444,8 +444,8 @@ void _glfwPlatformSetWindowTitle(_GLFWwindow* window, const char* title)
 void _glfwPlatformSetWindowIcon(_GLFWwindow* window,
                                 int count, const GLFWimage* images)
 {
-    // TODO
-    fprintf(stderr, "_glfwPlatformSetWindowIcon not implemented yet\n");
+    _glfwInputError(GLFW_PLATFORM_ERROR,
+                    "Wayland: Setting window icon not supported");
 }
 
 void _glfwPlatformGetWindowPos(_GLFWwindow* window, int* xpos, int* ypos)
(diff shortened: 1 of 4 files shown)


src/wl_window.c
```diff
@@ -429,14 +454,14 @@ void _glfwPlatformSetWindowSizeLimits(_GLFWwindow* window,
                                       int minwidth, int minheight,
                                       int maxwidth, int maxheight)
 {
-    // TODO
-    fprintf(stderr, "_glfwPlatformSetWindowSizeLimits not implemented yet\n");
+    // TODO: find out how to trigger a resize.
```
jadahl: Do you mean interactive resize here? What we need to do is to add client side window decorations where we would trigger resize and move. How exactly to do these I'd have to ask @elmindreda for her opinion. Initially we could just add a few transparent pixels around (using 4 subsurfaces), but might want something better than that later.

linkmauve: Actually no, this function is specified as modifying the current width and height of the window when it gets called.  Now that I think about it, it should probably call `_glfwInputWindowSize()` to achieve the desired effect.

src/wl_window.c
```diff
@@ -499,13 +528,22 @@ void _glfwPlatformMaximizeWindow(_GLFWwindow* window)
 
 void _glfwPlatformShowWindow(_GLFWwindow* window)
 {
-    wl_shell_surface_set_toplevel(window->wl.shell_surface);
+    if (!window->monitor)
```
jadahl: I know this was broken before, but shouldn't we also create the shell surface even if window->monitor != NULL?

linkmauve: Documentation for `glfwHideWindow()` says “If the window is already hidden or is in full screen mode, this function does nothing.” which means it shouldn’t be possible for a window to be both fullscreen (`window->monitor != NULL`) and hidden (`window->wl.visible == GLFW_FALSE`).

elmindreda:
Looks good. I updated the commit messages to fit within [50/72 characters](http://chris.beams.io/posts/git-commit/).

PR #739, opened 2016-04-16
Need to create the native window before creating the context. Fixes button states

BrandonSchaefer:
As creating the context creates the egl surface.

### DIFF
--- a/src/mir_window.c
+++ b/src/mir_window.c
@@ -160,10 +160,15 @@ static void handlePointerButton(_GLFWwindow* window,
                               int pressed,
                               const MirPointerEvent* pointer_event)
 {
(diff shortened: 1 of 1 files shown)


BrandonSchaefer:
Also this is fairly critical since you cant run any apps on mir because of it. Im not sure what changed, but now it just creates the mir needed things such as the NATIVE WINDOW before creating to the egl context which requires the NATIVE WINDOW :)

BrandonSchaefer:
I've another bug ... that Ill push to this pull. Fixes button states not being correct (ie. right left right left would fail to get the current button state). Now it works!

elmindreda:
Will review and merge asap. Thanks!

elmindreda:
Thank you! Merged with [updated commit messages](http://chris.beams.io/posts/git-commit/).

PR #741, opened 2016-04-18
Fixes  #740 - size callbacks incorrect on monitor orientation change

### LINKED ISSUE #740: glfwSetWindowSizeCallback and glfwSetFramebufferSizeCallback incorrect on monitor orientation change.
Avery Musbach (username on discourse DigitalEmotions) has [reported an issue with the size and framebuffersize callbacks on Windows](http://discourse.glfw.org/t/misleading-api-documentation/636).

I've verified that this issue exists as described on my Windows 8.1 system.

Specifics:

If the monitor undergoes an orientation change from landscape to portrait or visa versa in Windows the **glfwSetWindowSizeCallback** and **glfwSetFramebufferSizeCallback** functions report the incorrect width and height, with the these being transposed. This causes any viewport change using the values provided to result in incorrect rendering. This can be confirmed by calling **glfwGetWindowSize()** in the callback which will return the correct result.

Further investigation shows that the WM_SIZE message posted is the root of the error, as it appears to be incorrect after an orientation change. According to msdn the WM_DISPLAYCHANGE message should provide the correct width and heigh, but it is simpler to alter the WM_SIZE to check the client area using either glfwGetFramebufferSize, glfwGetWindowSize. I have a fix with this in which I'll provide a pull request for.

**Workaround** Use glfwGetFramebufferSize for viewport size setting. Note that this returns the same values  in Windows as glfwGetWindowSize.

**OS:** Windows 7 (unknown specifics), Windows 8.1 64bit.

dougbinks:
The WM_SIZE callback  of a fullscreen window appears incorrectly inverts the width & height when the monitor orientation is changed from landscape to portrait or visa versa.

This fix uses the glfwGet*Size functions to get the correct size via the client area. Although the functions for window and framebuffer size are currently the same on windows I've use both here in case future window scaling changes cause these to differ - but this may be over engineering things and a single get size call may be preferable.

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -637,8 +637,17 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
                 _glfwInputWindowIconify(window, GLFW_FALSE);
             }
 
-            _glfwInputFramebufferSize(window, LOWORD(lParam), HIWORD(lParam));
-            _glfwInputWindowSize(window, LOWORD(lParam), HIWORD(lParam));
+            // WM_SIZE params do not appear to catch display orientation
+            // changes correctly, so use Get functions to retreive.
+            // Both framebuffer and window sizes are the same, but get both
+            // here for potential future proofing.
+            int fbwidth, fbheight;
+            _glfwPlatformGetFramebufferSize(window, &fbwidth, &fbheight);
+            _glfwInputFramebufferSize(window, fbwidth, fbheight);
+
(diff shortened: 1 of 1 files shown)


dedred:
Hey,

Here are some of my meditations on life and art, please read them and tell me what you think, here is the link <http://nkuchocroja.floridacitygames.com/aelsloiw>

Yours sincerely, acreator@blueyonder.co.uk

PR #753, opened 2016-04-27
Problem checking for pre-existing joystick in matchCallback in cocoa_jo…

IntellectualKitty:
…ystick.m

The matchCallback function has an initial loop to filter out redundant joystick additions based on matching deviceRef values.  However, the if statement incorrectly combines this test with the condition that the joystick is not present, which is obviously incorrect.


elmindreda:
Thank you!

IntellectualKitty:
You're welcome!

PR #754, opened 2016-04-27
calloc error in cocoa_joystick.m

IntellectualKitty:
The calloc call for js->buttons uses 1 instead of sizeof(unsigned char) for the element size.

### DIFF
--- a/src/cocoa_joystick.m
+++ b/src/cocoa_joystick.m
@@ -322,7 +322,7 @@ static void matchCallback(void* context,
 
     js->axes = calloc(CFArrayGetCount(js->axisElements), sizeof(float));
     js->buttons = calloc(CFArrayGetCount(js->buttonElements) +
-                         CFArrayGetCount(js->hatElements) * 4, 1);
+                         CFArrayGetCount(js->hatElements) * 4, sizeof(unsigned char));
 
     _glfwInputJoystickChange(joy, GLFW_CONNECTED);
 }


tombsar:
I believe `sizeof char` is guaranteed to be 1 by the C standard. Have you come across a counter-example?

elmindreda:
What @tombsar said.

IntellectualKitty:
I believe section 6.5.3.4 of the C99 standard specifies that the size of char, whether signed, unsigned, or unspecified, is one.  The problem is that you are encoding an assumption rather than the actual size of your data type, which is dangerous.  Early versions of Microsoft Windows ran into problems because they made assumptions that 16-bit addressing was all that was ever going to be needed.  Now 32-bit addressing is virtually archaic, and it probably won't be that long really before even 64-bit addressing becomes obsolete.

Similarly, character strings used to be encoded as simple ASCII because that was all that was necessary and programmers assumed that that was all that would ever be necessary, but the world expanded with the Internet and now we encode strings in UTF-8 because we need to support multilingual character sets.  Eventually, however, support for UTF-8 will likely be abandoned as computers continue to become more powerful and the need for multilingual support expands since it will be faster and easier to process single, large characters rather than multi-byte compression encodings.

The only thing that we can really rely on in our discipline is that computer systems will inevitably become larger, faster, and more complex, and that more will be required of them than before.  That means that standards that were ironclad and seemingly unchangeable in the past have a high probability of falling by the wayside in the future.  The point here is that if you encode the actual size of your data type rather than an assumption about the size of your data type (which is based on a nearly 20 year old standard), you are future-proofing your code.

shurcooL:
Question, does `sizeof(unsigned char)` have the same runtime performance as the constant `1`?

IntellectualKitty:
The sizeof operator is evaluated at compilation time, so it has no effect on runtime performance.

PR #755, opened 2016-04-28
Use VK_API_VERSION_1_0 to make the Vulkan example build again

linkmauve:
The previous VK_API_VERSION define has been removed from recent headers.


elmindreda:
Fixed with 3f25610d2f30014382ddc508882adcd374fb3aa3. Thank you!

Issue #157, opened 2013-10-20
Build error with 10.7 SDK


sephirok:
Linking against the 10.7 SDK (CMAKE_OSX_SYSROOT /Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.7.sdk) instead of the 10.8 SDK produces the following compilation error:

<pre>
/Users/sephirok/Documents/glfw/src/cocoa_monitor.m:97:45: error: unknown type name 'CVDisplayLinkRef'; did you mean 'CGDisplayModeRef'?
                                            CVDisplayLinkRef link)
                                            ^~~~~~~~~~~~~~~~
                                            CGDisplayModeRef
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.7.sdk/System/Library/Frameworks/ApplicationServices.framework/Frameworks/CoreGraphics.framework/Headers/CGDirectDisplay.h:17:31: note: 
      'CGDisplayModeRef' declared here
typedef struct CGDisplayMode *CGDisplayModeRef;
                              ^
/Users/sephirok/Documents/glfw/src/cocoa_monitor.m:106:29: warning: implicit declaration of function
      'CVDisplayLinkGetNominalOutputVideoRefreshPeriod' is invalid in C99 [-Wimplicit-function-declaration]
        const CVTime time = CVDisplayLinkGetNominalOutputVideoRefreshPeriod(link);
                            ^
/Users/sephirok/Documents/glfw/src/cocoa_monitor.m:106:22: error: initializing 'const CVTime' with an expression of incompatible type 'int'
        const CVTime time = CVDisplayLinkGetNominalOutputVideoRefreshPeriod(link);
                     ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/sephirok/Documents/glfw/src/cocoa_monitor.m:344:5: error: unknown type name 'CVDisplayLinkRef'; did you mean 'CGDisplayModeRef'?
    CVDisplayLinkRef link;
    ^~~~~~~~~~~~~~~~
    CGDisplayModeRef
[...]

PR #757, opened 2016-05-05
Always make the window surface opaque

linkmauve:
This makes the examples and tests render as they do on other platforms,
with a black background instead of a transparent or semi-transparent
one.

Too bad, it made some examples look way better, but consistency is a good thing. Also waiting on #157 to provide the API to re-enable transparency.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -174,6 +174,21 @@ static const struct wl_surface_listener surfaceListener = {
     handleLeave
 };
 
+// Makes the surface considered as XRGB instead of ARGB.
+static void setOpaqueRegion(_GLFWwindow* window)
+{
+    struct wl_region* region;
+
+    region = wl_compositor_create_region(_glfw.wl.compositor);
+    if (!region)
+        return;
+
+    wl_region_add(region, 0, 0, window->wl.width, window->wl.height);
+    wl_surface_set_opaque_region(window->wl.surface, region);
+    wl_surface_commit(window->wl.surface);
+    wl_region_destroy(region);
+}
+
 static GLFWbool createSurface(_GLFWwindow* window,
                               const _GLFWwndconfig* wndconfig)
 {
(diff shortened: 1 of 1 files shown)


src/wl_window.c
```diff
@@ -174,6 +174,21 @@ static const struct wl_surface_listener surfaceListener = {
...
+    struct wl_region* region;
+
+    region = wl_compositor_create_region(_glfw.wl.compositor);
+    if (!region)
+        return;
+
+    wl_region_add(region, 0, 0, window->wl.width, window->wl.height);
+    wl_surface_set_opaque_region(window->wl.surface, region);
+    wl_surface_commit(window->wl.surface);
+    wl_region_destroy(region);
```
jadahl: Are we also getting a XRGB and not an ARGB surface from EGL?

elmindreda:
Hmm, I see I have more to learn about the new merge options. Sorry about losing your commit description.

jadahl:
@elmindreda I don't think the patch is correct. What wl_surface_set_opaque_region() does is to notify what part of the surface is opaque. GLFW has no clue about the content of the frames it passes along, but as far as I can see (at least when using mesa) the buffers are still ARGB8888, not XRGB8888.

Only if we always glClear() with black, or request an XRGB framebuffer configuration would it be Ok to set the opaque region to the whole surface.

elmindreda:
@jadahl Noted. Will look into it.

linkmauve:
jadahl, what XRGB8888 means is that the buffer has eight bits of padding where an ARGB8888 would have its alpha channel.  The representation is the same, only the way the compositor composites (or scanout) the buffer will change.

jadahl:
It's probably often harmless yes, but for example a compositor that may still just use the full buffer to do a complete composition including alpha blending while only using the opaque region to determine whether a surface that would be hidden should be painted or not the compositor may end up not painting/replying to frame callbacks that surface while it still is actually visible.

PR #774, opened 2016-06-06
Correct program name printed by tearing test

tombsar:
The tearing test incorrectly calls itself "iconify" when printing usage information; probably just a copy-paste error.


elmindreda:
Thank you!

PR #775, opened 2016-06-06
Add explanation of parameters to gamma test

tombsar:
Unlike similar test programs (e.g. tearing.c), the gamma test does not explain the meanings of the optional parameters `h` and `f`. This patch adds these explanations to the usage text.


elmindreda:
Thank you!

PR #785, opened 2016-06-14
Fix example build errors on VS 2010

ndogxj:
According to C89, all block-local variables need to be declared at the beginning of the block. MSVC2013 relaxes this restriction.

### DIFF
--- a/deps/linmath.h
+++ b/deps/linmath.h
@@ -192,19 +192,21 @@ static inline void mat4x4_rotate(mat4x4 R, mat4x4 M, float x, float y, float z, 
 	vec3 u = {x, y, z};
 
 	if(vec3_len(u) > 1e-4) {
-		vec3_norm(u, u);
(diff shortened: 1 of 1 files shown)


tombsar:
I do not think this patch is correct. You seem to have moved lines where variables are modified to before/after they are used by other code. i.e. `vec3_norm(u,u)` needs to come before the definition of `S`, and the definition of `idet` should come after the lines where `s` and `c` are modified, not before.

elmindreda:
What @tombsar said. This is a real bug but the fix is broken.

tombsar:
-1 the merged commit does not address my concerns about changing code behaviour. The line `vec3_norm(u, u);` now comes after the vector `u` is used in the definition of matrix `S`.

Unless, of course, the existing behaviour was broken?

elmindreda:
@tombsar No, you're right. :owlofshame:

PR #795, opened 2016-06-24
Disable default Alt key processing

halx99:
solve dead block problem when glfw-window is a cross-thread child window.

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -566,10 +566,10 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
             else
                 _glfwInputKey(window, key, scancode, action, mods);
 
-            if(wParam == VK_MENU) {
+            if(wParam == VK_MENU || wParam == VK_F10) {
                 return 0;
             }
-            break; // #fix: disable system default ALT process, solve dead block problem when glfw-window is a cross-thread child window.
+            break; // #fix: disable system default ALT/F10 process, solve dead block problem when glfw-window is a cross-thread child window.
         }
 
         case WM_LBUTTONDOWN:
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -566,7 +566,10 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
             else
                 _glfwInputKey(window, key, scancode, action, mods);
 
-            break;
+            if(wParam == VK_MENU) {
+                return 0;
+            }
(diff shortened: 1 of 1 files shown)


elmindreda:
That's an unusual setup. Why create the child window in another thread?

halx99:
Well, this problem confuse me long time, I write a game develop tool: http://x-studio365.com, in this case, I need create the glfw window as child of MFC MainFrame in a unique thread.
For detail, please see the reference question: https://social.msdn.microsoft.com/Forums/Windowsdesktop/zh-CN/373bd320-112a-4e55-bab5-74b19fe46bf3/my-mfc-application-will-blocking-when-alt-released-windows-10-rs1-14367-who-can-help-me?forum=vcgeneral

elmindreda:
This might be a good candidate for the upcoming platform-specific window hints.

elmindreda:
There are now platform-specific window hints.

halx99:
The F10 also should be disabled.

elmindreda:
This can now be controlled by the `GLFW_WIN32_KEYBOARD_MENU` window hint, which will be included in version 3.4.

halx99:
well, it doesn't works for me, I think also should add check at WM_SYSKEYDOWN/WM_SYSKEYUP just like follow:
```c
if ((wParam == VK_MENU || wParam == VK_F10) && !window->win32.keymenu) {
    // #Fix: disable system default ALT process, solve dead block problem when glfw-window is a cross-thread child window.
    // if user don't care.
    return 0;
}
```

elmindreda:
Oh, right, you use a child window. GLFW doesn't officially support having child windows, but I will look into whether this addition has any side-effects.

elmindreda:
This ties in somewhat with native window handle attachment, when GLFW will need to handle some level of this anyway.

halx99:
Yeah, glfw child window is works well for me, and I  use parent window to process alt key for shortcuts

halx99:
I think, it time to glfw official support child window, if you agree, I will send PR to support the feature

elmindreda:
@halx99 That sounds potentially interesting. What kind of support for child windows did you have in mind?

halx99:
At my project, I add a function glfwSetParent before call glfwCreateWindow

halx99:
I have another mind, change the 2nd parameter of glfwWindowHint to uintpr_t, then add a hint enum: GLFW_HINT_PARENT to store parent handle

halx99:
> I have another mind, change the 2nd parameter of glfwWindowHint to uintpr_t, then add a hint enum: GLFW_HINT_PARENT to store parent handle

PR #2497  created

Issue #789, opened 2016-06-19
XSetInputFocus Error on Non-Reparenting Linux WM


CaffeineViking:
The following minimal program below, with the specified setup, results in the later shown error message.
```C
#include <GLFW/glfw3.h>
int main() {
    glfwInit();
    GLFWwindow* handle = glfwCreateWindow(1280, 720, "bug?",
                                          NULL, NULL);
    glfwDestroyWindow(handle);
    glfwTerminate();
    return 0;
}
```
- **Library version:** `glfw3 3.2-4` (community/glfw-x11) on `gcc 6.1.1-2` (core/gcc)
- **Library version string:** `GLFW version 3.2.0 X11 GLX EGL clock_gettime /dev/js Xf86vm shared`
- **Operating system:**  `Linux 4.6.2-1-ARCH #1 SMP PREEMPT x86_64 GNU/Linux (Arch Linux)`
- **Window manager:** `dwm 6.1-3` (community/dwm) on `xorg-server 1.18.3-2` (extra/xorg-server)
```
X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  42 (X_SetInputFocus)
  Serial number of failed request:  132
  Current serial number in output stream:  134
```

Seems similar to what was reported in SFML 2.3, recently: https://github.com/SFML/SFML/issues/991
And also SDL2, some time ago: https://bugzilla.libsdl.org/show_bug.cgi?id=2997
It seems to be regarding the unique behaviour of some non-reparenting WMs.

Both GLFW and SFML seemed to start reporting this same issue on my setup a couple of weeks ago after a system upgrade, I have reproduced this on my other machine which also runs a similar setup. Lastly, I would like to thank for the amazing library, besides this issue, it has been working flawlessly!

elmindreda:
@CaffeineViking Thank you for an incredibly detailed bug report! The bugs you linked to are not only related but point to a possible second bug in GLFW.

roboman2444:
I get this same error using DWM. Someone suggested putting a sleep(1) after glfwInit(), but that does not fix the problem.

elmindreda:
Tentative fix available in the `xsetinputfocus-sync` branch.

elmindreda:
The fix has been merged. Thank you!

PR #798, opened 2016-06-27
Fixed XSetInputFocus before VisibilityNotify

CaffeineViking:
Should resolve glfw/glfw#789. Calling glfwCreateWindow might fail on non-reparenting window managers (like dwm or xmonad), since they don't set their visibility flag right after being mapped with XMapWindow. Attempting to print map_state (retrieved with XGetWindowAttributes) results in either IsUnmapped or IsUnviewable on the tested window manager (which is dwm 6.1-3) right before calling XSetInputFocus. Doing this gives BadMatch errors when actually calling XSetInputFocus, which should be expected given that the documentation in XLib says that this happens if a window isn't viewable when called. References to these follow below to make it easier to double check and test the changes I have made for this.

XMapWindow: https://tronche.com/gui/x/xlib/window/XMapWindow.html
XSetInputFocus: https://tronche.com/gui/x/xlib/input/XSetInputFocus.html
VisibilityNotify: https://tronche.com/gui/x/xlib/events/window-state-change/visibility.html
XGetWindowAttributes: https://tronche.com/gui/x/xlib/window-information/XGetWindowAttributes.html

I haven't programmed with Xlib before, so take everything I have said with a huge grain of salt. It does seem to fix the problem for me, with my window manager. However, I don't know if this patch works or breaks for everyone else involved. If you have any spare time, could anyone check if this patch seems somewhat reasonable (doesn't break anything and conforms with the planned glfw architecture, also not messing up too much of the code)? If it  doesn't indeed seem to break for anyone else, and fixes these issues for other similar WMs, could you consider merging these changes into mainline glfw for others?

Any feedback/improvements are welcome.
Thanks for your help and time!

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ does not find Doxygen, the documentation will not be generated.
 
 ## Changelog
 
+ - [X11] Bugfix: Didn't wait until window was visible before setting focus (#789)
  - Bugfix: Single compilation unit builds failed due to naming conflicts (#783)
  - Bugfix: The range checks for `glfwSetCursorPos` used the wrong minimum (#773)
  - [Win32] Bugfix: `glfwSetClipboardString` created an unneccessary intermediate
--- a/src/x11_platform.h
+++ b/src/x11_platform.h
@@ -112,6 +112,7 @@ typedef struct _GLFWwindowX11
     XIC             ic;
 
     GLFWbool        overrideRedirect;
+    GLFWbool        visuallyMapped;
 
     // Cached position and size used to filter out duplicate events
     int             width, height;
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -604,6 +604,7 @@ static GLFWbool createWindow(_GLFWwindow* window,
 
     _glfwPlatformGetWindowPos(window, &window->x11.xpos, &window->x11.ypos);
     _glfwPlatformGetWindowSize(window, &window->x11.width, &window->x11.height);
+    window->x11.visuallyMapped = GLFW_FALSE; // Assume window isn't mapped yet.
 
     return GLFW_TRUE;
 }
@@ -1365,6 +1366,29 @@ static void processEvent(XEvent *event)
             return;
         }
 
+        case VisibilityNotify:
+        {
+            int visibilityState = event->xvisibility.state;
+            // Some window managers (mostly non-reparenting
+            // like dwm, xmonad, ratpoision, openbox, cwm),
+            // don't set the visibility flag after mapping,
+            // a pre-condition for e.g. XSetInputFocus()...
+            // leads to BadMatch error as described in X11.
+            // Most of these will however notify with event
+            // VisibilityFullyObscured on visibility unset.
+            if (visibilityState != VisibilityFullyObscured)
+                window->x11.visuallyMapped = GLFW_TRUE;
+            return;
+        }
+
+        case UnmapNotify:
+        {
+            // Should be the only case when a window
+            // becomes completely visually unmapped.
+            window->x11.visuallyMapped = GLFW_FALSE;
+            return;
+        }
+
         case FocusIn:
         {
             if (window->cursorMode == GLFW_CURSOR_DISABLED)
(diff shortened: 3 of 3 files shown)


CaffeineViking:
Described bug succinctly in the changelog, as per the contribution guidelines. Also, please note that I haven't submitted a pull request before, and might be doing something stupid, notify if you need any additional fixes to be made regarding this and I'll get to work. Also, now testing additional WM setup.

CaffeineViking:
Quick update, I have confirmed that the X11 bug #789 is also present in `ratpoison 1.4.8`, another WM that is non-reparenting. Based on this, I surmise this is true for all of them, so `xmonad`, `cwm`, `uwm`, and `PLWM` [[1]](https://en.wikipedia.org/wiki/Re-parenting_window_manager); it would be great if someone else with such a setup could confirm the validity of this. Applying patch #798 (that is, this one), seems to resolve the issue on the tested window managers. Also, I have tested with `twm 1.0.9-1`, a normal re-parenting window manager, and both versions, with and without this patch, seem to work. Also, I have no idea why Travis failed, it seems it couldn't fetch the repository.

tgfrerer:
Checked out this patch on ubuntu xenial (16.04) / dwm 6.1 and I can confirm that it fixes the crash on init issue on my setup.

sdemos:
Tried out this patch on archlinux using xmonad-0.12 and it fixed the issue for me. Any idea when this will get pulled into the main repository?

elmindreda:
@sdemos Looking into it now.

elmindreda:
It seems from reading dwm (which is super-readable :heart_eyes:) that all we need to do is wait for the regular `MapNotify` (and should have been doing that all along).

sdemos:
Awesome, thanks for looking into it. Is that approach different than the fix that is already in this pr? If it is, I would be totally willing to test out that patch to confirm for xmonad as well.

elmindreda:
@sdemos It is, although this approach resolves it as well. Thank you! Will update when I have a branch.

CaffeineViking:
Same here @elmindreda, just call me over if you need further help in getting your branch tested. Also, I'd like to note that it might be better to trigger on `VisibilityNotify` instead of `MapNotify`, according towards some documentation I have read. e.g. `XSetInputFocus` requires given window to be mapped and also visible. Most window managers seem to become visible when mapped, but unfortunately this isn't true for all of them (e.g. awesomewm), at least based on the observations given over here: [SFML](https://github.com/SFML/SFML/blob/61526628d1ea7c2ac1b610d843ca3b8c8e1b0ff5/src/SFML/Window/Unix/WindowImplX11.cpp#L2127-L2147) and also in the Xlib documentation [XSetInputFocus](https://tronche.com/gui/x/xlib/input/XSetInputFocus.html). But this might not really apply here, let's try with a `MapNotify` first as you said, since that is usually how it works, and we can always verify this issue later.

derhass:
I can confirm that this issue was present with ctwm, too, and that this fix resolves it.

elmindreda:
I can reproduce this consistently on dwm and have a fix following what is outlined in the ICCCM, but I cannot reproduce it on awesome. Has anyone here encountered this issue on awesome?

CaffeineViking:
Sorry @elmindreda, I've made a mistake. Seems Awesome WM isn't non-reparenting, so issue #789 shouldn't affect it directly, therefore, my argument that VisibilityNotify is needed in this case is not true. Also testing said awesome on my setup without this patch, and it seems to be working just fine for me too, so needless to say, it's unaffected. Also, you mentioned a solution based the ICCCM? The bugfix proposed in this PR is very hacky, so it would be great with a proper standard compliant fix. Please do message me if you need additional help in testing your fix, I have a couple of machines which use the affected window managers (dwm, ratpoison and xmonad), so it should be no sweat if you need it :+1:

elmindreda:
@CaffeineViking The ICCCM agrees with you; VisibilityNotify is (part of) the correct solution, most clearly described in [section 4.2.2](https://tronche.com/gui/x/icccm/sec-4.html#s-4.2.2). I probably wouldn't have looked further than MapNotify if you hadn't pointed that out, so thank you, and again for the offer to help with testing!

elmindreda:
Please see if the `xsetinputfocus-sync` branch fixes this issue for you.

sdemos:
confirmed for xmonad-0.12 on archlinux. Thanks!

CaffeineViking:
@elmindreda verified that the issues present on ``dwm 6.1`` and ``ratpoison 1.4.8`` are indeed fixed!

derhass:
@elmindreda verified that `xsetinputfocus-sync` fixes the issue on ctwm 3.7 and 3.8.

elmindreda:
Thank you @CaffeineViking, @sdemos, @derhass and @tgfrerer for the thorough testing! This concludes the next to last issue for 3.2.1 and then it's time for 3.3.

PR #807, opened 2016-07-14
Change name in acknowledgements list

tombsar:
I recently noticed that I was added to the glfw acknowledgements list a while back (not sure my contributions have been worth that much to the project, so I'm hugely flattered!). Would it be okay to use my full name, rather than first initial?

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ skills.
  - Nathan Sweet
  - TTK-Bandit
  - Sergey Tikhomirov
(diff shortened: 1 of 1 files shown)


elmindreda:
Absolutely, I just didn't find your full name at the time.

elmindreda:
As for why, you have a very good eye for catching my more embarrassing mistakes. That's a Good Thing.

PR #823, opened 2016-08-04
Added preproccessor directive to build Vulkan DLL

rhynodegreat:
If `vulkan.h` was included or if `GLFW_INCLUDE_VULKAN` was defined in `glfw_config.h`, then attempting to build GLFW as a DLL would cause compilation errors, since `internal.h` redefines a few vulkan types.


elmindreda:
You must not define public header macros when compiling GLFW. There is a note to that effect at the end of the [section on manual compilation](http://www.glfw.org/docs/latest/compile_guide.html#compile_manual).

elmindreda:
Thank you for reporting this!

Issue #680, opened 2015-12-31
glfwSwapBuffers does not block with swap interval > 0 when window is minimized or covered by another window


badlogic:
The docs of [`glfwSwapBuffers`](http://www.glfw.org/docs/latest/group__window.html#ga15a5a1ee5b3c2ca6b15ca209a12efd14) say that this call will vsync (to it's best ability in windowed mode i guess) if the swap interval is > 0.

This works if the window is partially visible (either full-screen or windowed mode). It actually doesn't "vsync" if the window is fully covered by another window or if the window is minimized. Is this by design? Only tested on Mac OS X.

I'm not sure if this could/should be "fixed" in GLFW or if it's the responsibility of the application to take care of it. However, it should at least be documented in the glfwSwapBuffers docs that no vsync'ing happens in case the window is fully covered or minimized.

elmindreda:
That is by design, but it's Apple's design. OS X ignores the NSOpenGL swap interval when the window is obscured. No other OS does this.

I will look into working around this with display link at some point. If that doesn't work, I don't think there's anything GLFW can do.

elmindreda:
I will add a note to the documentation for the next release.

meshula:
This is an energy saving feature called App Nap. You need to disable it in applications that expect to do background processing such as render serving, or that should continue running when obscured. It can be accomplished from code, and would probably be a good option for glfw to expose. Look for App Nap on this page:

https://developer.apple.com/library/mac/releasenotes/MacOSX/WhatsNewInOSX/Articles/MacOSX10_9.html

elmindreda:
That looks like yet another good candidate for the platform specific hints I'm sketching out.

shurcooL:
If App Nap is working as I understand, is this a problem? Hopefully, what's happening is that `glfwSwapBuffers` isn't blocking, but the app is still being suspended enough that the power savings are even greater than with just vsync.

Or did you want to disable that power saving functionality instead?

meshula:
If you implement a render server, for example, an application that creates images to be displayed on some other device, or just for screenshots, that application might run in the background. App Nap might suspend it. A primary reason to disable app nap is to enable server applications.

badlogic:
My use case is to put the app mostly to sleep when obscured/minified. Doable from the application code itself. Thanks for the clarification @elmindreda!

ithron:
I think an application should suspend its rendering when the window is occluded. There is no point continue rendering when there is nothing to see.
[...]

PR #834, opened 2016-08-13
Treat occluded windows as hidden

ithron:
If a window is not hidden but occluded, e.g. by another window or the
window is on another screen/desktop that is currently not shown,
_glfwPlatformWindowVisible() returns 1 although the window is physically
not visible. This commit fixes this issue by treating occluded windows
as hidden, i.e. non visible.
Usage example:
When the render loop is limited by v-sync, glfwSwapBuffers() does not
block on OS X when AppNap is enabled and the window is occluded.
Therefore the application starts to burn CPU cycles althoug nothing is
displayed. With the fix in this commit the application could check if
the window is visible using glfwGetWindowAttrib(window, GLFW_VISIBLE)
and if the window is hidden (or occluded) it could pause the render
loop.
This is a workaround for issue #680

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1339,7 +1339,8 @@ int _glfwPlatformWindowIconified(_GLFWwindow* window)
 
 int _glfwPlatformWindowVisible(_GLFWwindow* window)
 {
-    return [window->ns.object isVisible];
+  return [window->ns.object isVisible] &&
+         ([window->ns.object occlusionState] & NSWindowOcclusionStateVisible);
 }
 
 int _glfwPlatformWindowMaximized(_GLFWwindow* window)


shurcooL:
As an observation, I think it's dangerous to conflate window visibility and occlusion into one.

An invisible window is one that can never be seen because it's kept invisible. A user may decide to call some API to make it visible, in order for the window to appear.

An occluded window isn't visible, but that's a temporary state that can't and shouldn't be controlled by the same APIs. To make it not occluded, you could close/hide other windows, or bring it to top. But you shouldn't try to "make it visible", since it's already visible, just occluded.

ithron:
I totally agree. Maybe a better solution would be to add another sate to the system, like GLFW_OCCLUDED. But I don't have any idea if this could be implemented for system other than macOS. An ideal solution would add a callback, so that the user can be notified about the window occlusion status.

elmindreda:
As a side-effect of fixing #1337, OpenGL vsync now uses a display link and is not affected by window occlusion.

PR #836, opened 2016-08-15
examples/particles: Fix periodic stutter: ensure that tv_nsec<1e9

Eliasvan:
Each time when "ts.tv_nsec" became greater or equal than 1e9, the cnd_timedwait() function returned immediately, causing the thread to become a busy waiting loop (slowing the other thread, on which the current was waiting, down), until "ts.tv_nsec" became smaller than 1e9.
This caused noticeable stutter one 10th of a second, since the timeout was 100ms.

### DIFF
--- a/examples/particles.c
+++ b/examples/particles.c
@@ -457,7 +457,12 @@ static void draw_particles(GLFWwindow* window, double t, float dt)
     {
         struct timespec ts;
         clock_gettime(CLOCK_REALTIME, &ts);
-        ts.tv_nsec += 100000000;
+        ts.tv_nsec += 100 * 1000 * 1000;
+        if (ts.tv_nsec >= 1000 * 1000 * 1000)
+        {
+            ts.tv_sec++;
+            ts.tv_nsec -= 1000 * 1000 * 1000;
+        }
         cnd_timedwait(&thread_sync.p_done, &thread_sync.particles_lock, &ts);
     }
 
@@ -908,7 +913,12 @@ static int physics_thread_main(void* arg)
         {
             struct timespec ts;
             clock_gettime(CLOCK_REALTIME, &ts);
-            ts.tv_nsec += 100000000;
+            ts.tv_nsec += 100 * 1000 * 1000;
+            if (ts.tv_nsec >= 1000 * 1000 * 1000)
+            {
+                ts.tv_sec++;
+                ts.tv_nsec -= 1000 * 1000 * 1000;
+            }
             cnd_timedwait(&thread_sync.d_done, &thread_sync.particles_lock, &ts);
         }
 


elmindreda:
Thank you for finding the cause of this very annoying issue! The fix should probably be upstreamed to TinyCThread.

examples/particles.c
```diff
@@ -457,7 +457,12 @@ static void draw_particles(GLFWwindow* window, double t, float dt)
     {
         struct timespec ts;
         clock_gettime(CLOCK_REALTIME, &ts);
-        ts.tv_nsec += 100000000;
+        ts.tv_nsec += 100 * 1000 * 1000;
+        if (ts.tv_nsec >= 1000 * 1000 * 1000)
```
shurcooL: Should this be `if` or `while`?

Eliasvan: It should be `if`, so the patch is valid.

`if` because `clock_gettime()` is guaranteed to write a valid value for `ts` (such that `ts.tv_nsec >= 0 && tv_nsec < 1e9`), at least when `clock_gettime()` succeeds (returns 0), and because the added value (= timeout) to `ts.tv_nsec` is only 100e6.
Thus in the worst case, `ts.tv_nsec` can become 1_099_999_999 (= (1e9 - 1) + 100e6), which is smaller than 2e9, thus at most one subtraction needs to be done, hence no `while` loop needed.

Note that when `clock_gettime()` doesn't succeed, which could be possible when `CLOCK_REALTIME` is not valid for the OS (and thus when it would return `EINVAL`), `ts` is *not* valid anymore.
Actually, this should be checked, but considering almost no error checking is done for the rest of that example (e.g., see the call to `cnd_timedwait()`; it's only an example after all), I did not change that part.

Eliasvan: @elmindreda 
Quoting the reference http://en.cppreference.com/w/c/thread/thrd_errors :

    enum {
        thrd_success = /* unspecified */,
        thrd_nomem = /* unspecified */,
        thrd_timedout = /* unspecified */,
        thrd_busy = /* unspecified */,
        thrd_error = /* unspecified */
    };

Therefore, the following [defines](https://github.com/tinycthread/tinycthread/blob/b8a4c7d4ddd1e8ec456d87e24bbe3b28b24ae343/source/tinycthread.h#L170) in TinyCThread look very suspicious to me (especially since success is normally always defined as 0 in POSIX!): 

    /* Function return values */
    #define thrd_error    0 /**< The requested operation failed */
    #define thrd_success  1 /**< The requested operation succeeded */
    #define thrd_timedout 2 /**< The time specified in the call was reached without acquiring the requested resource */
    #define thrd_busy     3 /**< The requested operation failed because a tesource requested by a test and return function is already in use */
    #define thrd_nomem 4 /**< The requested operation failed because it was unable to allocate memory */`

Anyway, `cnd_timedwait()` returned `thrd_error` when `ts` was invalid (I'm testing on Linux), so apart from the weird defines, TinyCThread is working as it should.
Therefore I don't see what you mean with:

> The fix should probably be upstreamed to TinyCThread.

shurcooL: > `if` because `clock_gettime()` is guaranteed to write a valid value for `ts` (such that `ts.tv_nsec >= 0 && tv_nsec < 1e9`), at least when `clock_gettime()` succeeds (returns 0), and because the added value (= timeout) to `ts.tv_nsec` is only 100e6.
Thus in the worst case, `ts.tv_nsec` can become 1_099_999_999 (= (1e9 - 1) + 100e6), which is smaller than 2e9, thus at most one subtraction needs to be done, hence no `while` loop needed.

@Eliasvan, I agree with your assessment. Thanks for confirming. 👍

elmindreda:
Fixed it without a conditional. I prefer avoiding them where possible.

Forgot to add credit for locating the bug in the closing commit (sorry) but added that with 2e6a110181b1aa08e0a8a43cdb6c88c4a642ab3a.

Eliasvan:
Great. I suppose the compiler will be smart enough to optimize the "/" and "%" operators.

PR #839, opened 2016-08-16
Add relative mouse support and the cursor modes

BrandonSchaefer:
Implements a few other missing functions. Cleaning up naming convention
as well.

Fix FindMir.cmake was not finding the correct mirclient.so when
you wanted something other then the system library.

### DIFF
--- a/CMake/modules/FindMir.cmake
+++ b/CMake/modules/FindMir.cmake
@@ -2,17 +2,36 @@
 #
 # This will define:
 #
-#   MIR_LIBRARIES   - Link these to use Wayland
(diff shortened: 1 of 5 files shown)


BrandonSchaefer:
Note this depends on mir 0.24 (which has not been released quite yet!)

elmindreda:
So it would be a bad idea to merge this for a release that's going out this week?

BrandonSchaefer:
Hmm maybe. Not sure how many people are using the mir side of things atm :). So no rush. I can always fix conflicts that come in later if you want to sit on this for a bit!

Issue #281, opened 2014-04-23
OSMesa context support


andrewcorrigan:
Can GLFW please support creating an OSMesa context for pure software rendering, as an alternative to GLX or EGL on Linux?

My OpenGL 2.0+ application needs to support rendering on remote systems with X forwarding over SSH, without VNC or anything.  With X forwarding, I can get at most an OpenGL 1.x context regardless of the hardware supported on the client, and even worse, large three-dimensional datasets need to be transferred over the network prior to being rendered on the client.  With Mesa, I can get an OpenGL 2.1 rendering context, with all rendering performed on the server (albeit in software), and then only the final 2D image needs to be transferred back to the client.

Currently my application uses a weird combination of SDL2, software rendering only, and OSMesa, but it would ideal to be able to just compile GLFW to use X11 + OSMesa, without GLX or EGL.

Alternatively, if it would be possible to get GLFW to just setup a window and handle input, but not context management, that would be helpful.  I could then figure out how to blit pixels to the screen via the existing Native Access functionality.

Finally, if anyone has a better way to perform OpenGL 2.0+ rendering on a remote server, without transferring large 3D data sets to the client, and using only X forwarding (no VNC), I would really appreciate any advice.  Perhaps there is a simpler way to accomplish these things?

chuanqichen:
@elmindreda What's the status of OSMesa support? I am looking for doing offscreen rendering even without video driver. Great work guys, please let me know how can I do that.

jshin49:
@elmindreda Any status updates regarding this issues? Very appreciated guys!

elmindreda:
See the `osmesa` branch for a headless implementation of OSMesa support.

elmindreda:
Support for OSMesa context creation on all supported backends has been added with e9560ef021c101def23a5c3b383d057b2a0e58d3.

Sorry for the long delay. Perhaps this is still useful to some of you.

PR #850, opened 2016-08-31
OSMesa backend

lyceel:
This adds basic support for using OSMesa to create an in-memory rendering surface and OpenGL context.  Many of the GLFW entry points (cursors, window positioning, etc) are no-ops or stubs since they don't make sense in this context.  Also, input events don't work at all, since there's not really a good way for an offscreen window to gain input focus.

Since none of the existing examples made it obvious that this was actually working, I created duplicate examples of the simple and particles examples that will run briefly, then dump the framebuffer to an image file and exit.  These examples are only compiled when OSMesa is enabled.

### DIFF
--- a/examples/particles_osmesa.c
+++ b/examples/particles_osmesa.c
@@ -1,0 +1,1117 @@
+//========================================================================
+// A simple particle engine with threaded physics
+// Copyright (c) Marcus Geelnard
+// Copyright (c) Camilla Berglund <elmindreda@glfw.org>
+//
+// This software is provided 'as-is', without any express or implied
+// warranty. In no event will the authors be held liable for any damages
+// arising from the use of this software.
+//
+// Permission is granted to anyone to use this software for any purpose,
+// including commercial applications, and to alter it and redistribute it
+// freely, subject to the following restrictions:
+//
+// 1. The origin of this software must not be misrepresented; you must not
+//    claim that you wrote the original software. If you use this software
+//    in a product, an acknowledgment in the product documentation would
+//    be appreciated but is not required.
+//
+// 2. Altered source versions must be plainly marked as such, and must not
+//    be misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source
+//    distribution.
+//
+//========================================================================
+
+#if defined(_MSC_VER)
+ // Make MS math.h define M_PI
+ #define _USE_MATH_DEFINES
+#endif
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <math.h>
+#include <time.h>
+
+#include <tinycthread.h>
+#include <getopt.h>
+#include <linmath.h>
+
+#include <glad/glad.h>
+#include <GLFW/glfw3.h>
+
+// Define tokens for GL_EXT_separate_specular_color if not already defined
+#ifndef GL_EXT_separate_specular_color
+#define GL_LIGHT_MODEL_COLOR_CONTROL_EXT  0x81F8
+#define GL_SINGLE_COLOR_EXT               0x81F9
+#define GL_SEPARATE_SPECULAR_COLOR_EXT    0x81FA
+#endif // GL_EXT_separate_specular_color
+
+
+//========================================================================
+// Type definitions
+//========================================================================
+
+typedef struct
+{
+    float x, y, z;
+} Vec3;
+
+// This structure is used for interleaved vertex arrays (see the
+// draw_particles function)
+//
+// NOTE: This structure SHOULD be packed on most systems. It uses 32-bit fields
+// on 32-bit boundaries, and is a multiple of 64 bits in total (6x32=3x64). If
+// it does not work, try using pragmas or whatever to force the structure to be
+// packed.
+typedef struct
+{
+    GLfloat s, t;         // Texture coordinates
+    GLuint  rgba;         // Color (four ubytes packed into an uint)
+    GLfloat x, y, z;      // Vertex coordinates
+} Vertex;
+
+
+//========================================================================
+// Program control global variables
+//========================================================================
+
+// Window dimensions
+float aspect_ratio;
+
+// "wireframe" flag (true if we use wireframe view)
+int wireframe;
+
+// Thread synchronization
+struct {
+    double    t;         // Time (s)
+    float     dt;        // Time since last frame (s)
+    int       p_frame;   // Particle physics frame number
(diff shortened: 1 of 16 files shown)


elmindreda:
This is very nicely done, but why wouldn't one use OSMesa directly instead of this? It seems to add nothing beyond OSMesa except perhaps the timer.

elmindreda:
This is related to #281.

lyceel:
In our case, we'd like to use it as a replacement backend to run unit tests
on some glfw-based code.  The tests are run on a headless server, so normal
OpenGL implementations aren't available.

Having an OSMesa backend let's us simply swap out the glfw library for one
compiled to use OSMesa and test the code in place.

elmindreda:
@lyceel I see, thank you! This is certainly low impact and nearly perfect for hooking up to CI to stave off bitrot. Will very likely merge.

lyceel:
Sounds good.  Thanks!

elmindreda:
Will you be maintaining this code or is it a one-time drop?

lyceel:
Realistically, it'll probably be somewhere in between.  I'll maintain them as long as I'm using them (which should be for a decent while at least).  I also don't have the ability to test or fix issues on platforms other than Linux).

elmindreda:
@lyceel It's fine either way; it's just useful for me to know. Thanks!

src/osmesa_init.c
```diff
@@ -0,0 +1,35 @@
...
+//////////////////////////////////////////////////////////////////////////
+
+
+//////////////////////////////////////////////////////////////////////////
+//////                       GLFW platform API                      //////
+//////////////////////////////////////////////////////////////////////////
+
+int _glfwPlatformInit(void)
+{
+  if (!_glfwInitThreadLocalStoragePOSIX()) {
```
elmindreda: GLFW uses [Allman style](https://en.wikipedia.org/wiki/Allman_style) for code.

Indentation is four spaces.

examples/particles_osmesa.c
```diff
@@ -0,0 +1,1117 @@
+//========================================================================
```
elmindreda: I won't merge this file.

examples/simple_osmesa.c
```diff
@@ -0,0 +1,209 @@
+//========================================================================
```
elmindreda: I'll merge this file as `offscreen.c` if it is built for all platforms, uses `stb_image_write.h` and sets `GLFW_VISIBLE` to false.

elmindreda:
With minimal changes this works as expected on Cygwin.

elmindreda:
A heads-up to avoid duplicate work: I've made most of these changes and a bunch more, some trivial, others subtle. Will publish a counter-proposal branch for your approval as soon as I'm done, both because it involves copyrights and because you've expressed interest in maintainership.

lyceel:
Thanks for that.  I was just about to dive in and make changes :-)

elmindreda:
@lyceel Please have a look at the `osmesa` branch and see if it's acceptable.

lyceel:
Code changes look fine, but I'm getting "Error: OSMesa: Failed to load
required entry points" when running the new offscreen example.

Relevant strace lines:

kakariko:~/github/glfw/examples$ strace ./offscreen 2>&1  | grep -i osmesa
open("/usr/local/google/home/dalyj/github/glfw/src/libOSMesa.so.6",
O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/usr/lib/x86_64-linux-gnu/libOSMesa.so.6", O_RDONLY|O_CLOEXEC) = 3
write(2, "Error: OSMesa: Failed to load re"..., 52Error: OSMesa:
Failed to load required entry points


So the library is getting loaded, checking the symbols:

kakariko:~/$ objdump -T /usr/lib/x86_64-linux-gnu/libOSMesa.so.6 | grep OSMesa
/usr/lib/x86_64-linux-gnu/libOSMesa.so.6:     file format elf64-x86-64
00000000000515f0 g    DF .text  0000000000000020  Base        OSMesaColorClamp
0000000000050be0 g    DF .text  00000000000003db  Base
OSMesaCreateContextExt
00000000000512c0 g    DF .text  0000000000000082  Base
OSMesaDestroyContext
0000000000051690 g    DF .text  00000000000000ec  Base        OSMesaGetIntegerv
0000000000050b80 g    DF .text  000000000000005c  Base
OSMesaGetColorBuffer
0000000000051780 g    DF .text  0000000000000072  Base
OSMesaGetProcAddress
00000000000515e0 g    DF .text  0000000000000005  Base
OSMesaGetCurrentContext
0000000000051610 g    DF .text  0000000000000074  Base        OSMesaPixelStore
0000000000052ce0 g    DF .text  0000000000000215  Base        OSMesaMakeCurrent
0000000000051800 g    DF .text  000000000000001d  Base        OSMesaPostprocess
0000000000050fc0 g    DF .text  0000000000000014  Base
OSMesaCreateContext
0000000000050b10 g    DF .text  000000000000006c  Base
OSMesaGetDepthBuffer


So my copy of the library doesn't have the OSMesaCreateContextAttribs entry
point that you added, which kinda puts
a damper on the nice context creation changes as well :-(


This was apparently added in Mesa 11.2 (April 2016), and my machine has:

kakariko:~/github/glfw/src$ dpkg-query --show libosmesa6-dev
libosmesa6-dev  10.1.3-0ubuntu0.6

Unfortunately, I don't control the development environment on my machine.
If you want to go ahead with the newer OSMesa version, that's OK with me,
but I'll have to revert those changes on my side and go back to the plain
vanilla OSMesaCreateContext.  On the other hand, it might be prudent to
avoid such a new (OK, not *that* new, but still...) feature as other users
might hit this issue as well.

I'm OK with whatever you decide.

elmindreda:
@lyceel Does it have `OSMesaCreateContextExt`?

elmindreda:
I'll add `OSMesaCreateContextExt` as a fallback.

lyceel:
Yes OSMesaCreatecontextExt is there in 10.1.3.

elmindreda:
@lyceel Fallback added.

lyceel:
That did it.  Looks good to me 👍

elmindreda:
Cool!

I may re-purpose some of these files to a generic null backend in the future, of course without compromising your use case.

elmindreda:
Thank you!

lyceel:
Thank you!  Looking forward to 3.3 :-)

PR #858, opened 2016-09-10
Update vulkan.c

PDirSG:
I know it is overkill but in first OpenGL releases, there is have only two or three Extensions. I think it is same for Vulkan. More futuristic...

### DIFF
--- a/src/vulkan.c
+++ b/src/vulkan.c
@@ -104,20 +104,66 @@ GLFWbool _glfwInitVulkan(void)
         return GLFW_FALSE;
     }
 
-    for (i = 0;  i < count;  i++)
+    for (i = 0;  i < count;  i++) 
     {
-        if (strcmp(ep[i].extensionName, "VK_KHR_surface") == 0)
-            _glfw.vk.KHR_surface = GLFW_TRUE;
-        if (strcmp(ep[i].extensionName, "VK_KHR_win32_surface") == 0)
-            _glfw.vk.KHR_win32_surface = GLFW_TRUE;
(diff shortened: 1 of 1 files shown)


elmindreda:
There is no reason to do this.

BestThingsInLiveFree:
Now, if we do not add "continue;" statement, it will continue even if find an required extension. then performance--

Now, if we do not add "if(!<extensionName>)"  statement, it will comparise a founded extension. then performance--

if we do not  add 'if(_glfw.vk.KHR_surface && _glfw.vk.KHR_win32_surface)' statement, it will continue to search extensions even required extensions found. Then performance--

And the preprocessor statements: what is the mean of searching an Linux extension in a Windows machine? Then performance--

I know it is overkill, but in OpenGL 1.0, there is only few extensions. How many now? More than 100... I think it is same for Vulkan.

shurcooL:
> I know it is overkill, but in OpenGL 1.0, there is only few extensions. How many now? More than 100... I think it is same for Vulkan.

I think that's an argument for why this may become important to do in the future. It may or may not be worth the cost now.

PDirSG:
Then maybe we should do a performance test?

elmindreda:
@PDirSG I won't merge this. You are right, however, that there is unnecessary and easily avoided work being done. This has been resolved with 3e59b7345c8ebef6ded0a217a41f61c785bff3e7 and 54930ddd33b83c18bbdde634fc6e443d7ca1b20f, more because of the clarity it adds than any performance concern. The EGL code will look similar once support for EGL 1.5 platform extensions is added.

This is still O(n³), but _n_ is small and the data is hot.

PR #859, opened 2016-09-11
Only input a character on GLFW_PRESS action

### LINKED ISSUE #804: Wayland character callback registers on key release
The character callback event triggers on both key press and key release, this does not happen when using X11.

linkmauve:
Fixes #804.


linkmauve:
Merged in 2eb1657.

PR #864, opened 2016-09-20
Add version directives to GLSL shaders

tombsar:
Although not technically required, I believe it is best practice for shaders to specify which version of the GLSL spec they conform to. To quote the spec:

> Shaders should declare the version of the language they are written to.

NB: I have presumed the version is 1.10, but it should be corrected if a later version was intended.

### DIFF
--- a/tests/cursor.c
+++ b/tests/cursor.c
@@ -47,6 +47,7 @@
 #define CURSOR_FRAME_COUNT 60
 
 static const char* vertex_shader_text =
+"#version 110\n"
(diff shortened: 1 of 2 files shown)


elmindreda:
Thank you! Would you mind doing this for all tests and examples with GLSL (except `heightmap.c`)?

tombsar:
Sorry I didn't make time to do this...

elmindreda:
@tombsar No worries, and thank you for getting the ball rolling on this.

PR #867, opened 2016-09-22
Stop github thinking this is Mathematica

tbodt:
`*` has a higher precedence than `++`, so this doesn't change the functionality. It does remove any occurrences of `(*`, which starts a Mathematica comment.


tbodt:
Fuck, I'm wrong.

PR #871, opened 2016-09-28
Use EGL_KHR_get_all_proc_addresses if possible.

kbrenneman:
On systems that support the EGL_KHR_get_all_proc_addresses extension, there's no need to load a client API library, since you can just load any OpenGL function through eglGetProcAddress.

This is especially useful for full OpenGL contexts (rather than GLES), because there isn't a client library to use with EGL. The libGLES\*.so libraries are supposed to be for GLES, and libGL.so.1 is only for GLX.

The change itself is pretty simple. It just checks for whether EGL_KHR_get_all_proc_addresses is supported in `_glfwInitEGL`, and then in `_glfwCreateContextEGL`, it'll skip trying to load the client library.


elmindreda:
Thank you!

Issue #347, opened 2014-09-02
Long delay when calling XRRGetScreenResources


EloiMarin:
Since calls to XRRGetScreenResources can be very long (due to hardware polling), maybe would be better to use XRRGetScreenResourcesCurrent when this performance penalty can be noticeable, ie:

_glfwPlatformGetVideoModes
_glfwPlatformGetVideoMode
_glfwPlatformGetMonitorPos (also?)
(always in X11 platform)

I measured in my system an improvement of 3 orders of magnitude between XRRGetScreenResources and XRRGetScreenResourcesCurrent, and of 2 orders of magnitude in the execution time of the function glfwGetVideoMode between the two alternatives.

Maybe the rest of "setters" functions in x11_monitor.c can be left with XRRGetScreenResources, and also the one in initExtensions (x11_init.c), since I think, although I'm not sure, that these functions are not prone to be called very often, like in every render frame.

(Also, I've to say that I don't have much experience in Github...I think I did well but I was not sure if what I've to do is open an issue like I did or send a patch via pull request, so...I wait for your comments!)

PR #877, opened 2016-10-07
Optimize calls to XRRGetScreenResources

atom-symbol:
This patch substantially improves the startup time of GLFW apps when using Mesa OpenGL.

Please merge. Thanks.


elmindreda:
Thank you, will review.

elmindreda:
This is related to #347.

atom-symbol:
Hello

Will this patch be part of release 3.2.2 or 3.3.0? I am just asking because I don't know what your plans about this patch are.

Thanks

elmindreda:
@atom-symbol There very likely won't be a 3.2.2 but it will very likely be included in 3.3.

src/x11_init.c
```diff
@@ -458,6 +458,16 @@ static void detectEWMH(void)
     XFree(supportedAtoms);
 }
 
+XRRScreenResources* _glfwXRRGetScreenResources()
+{
+#if (RANDR_MAJOR > 1 || (RANDR_MAJOR == 1 && RANDR_MINOR >= 3))
```
elmindreda: Version 1.3 is already required for RandR to be used (see `initExtensions`) so this function should be replaced with direct calls to `XRRGetScreenResourcesCurrent`.

src/x11_init.c
```diff
@@ -492,8 +502,7 @@ static GLFWbool initExtensions(void)
 
     if (_glfw.x11.randr.available)
     {
-        XRRScreenResources* sr = XRRGetScreenResources(_glfw.x11.display,
-                                                       _glfw.x11.root);
+        XRRScreenResources* sr = _glfwXRRGetScreenResources();
```
elmindreda: This invalid call emits no warnings because of the the bad declaration below. It will pull its arguments from random garbage on the stack and likely segfault as the first parameter is a pointer.

src/x11_platform.h
```diff
@@ -299,4 +299,6 @@ void _glfwInputErrorX11(int error, const char* message);
 
 void _glfwPushSelectionToManagerX11(void);
 
+XRRScreenResources* _glfwXRRGetScreenResources();
```
elmindreda: In C this is __not__ a shorthand for `void` as it is in C++. It's [an older form of function declaration](http://en.cppreference.com/w/c/language/function_declaration) that means the parameter number and types are unknown.

src/x11_monitor.c
```diff
@@ -114,7 +114,7 @@ GLFWbool _glfwSetVideoModeX11(_GLFWmonitor* monitor, const GLFWvidmode* desired)
         if (_glfwCompareVideoModes(&current, best) == 0)
             return GLFW_TRUE;
 
-        sr = XRRGetScreenResources(_glfw.x11.display, _glfw.x11.root);
+        sr = _glfwXRRGetScreenResources();
```
elmindreda: This invalid call emits no warnings because of the the bad declaration below. It will pull its arguments from random garbage on the stack and likely segfault as the first parameter is a pointer.

PR #879, opened 2016-10-11
Implement compose key for character input

linkmauve:
This commit has been copied almost verbatim from Bryce Harrington’s patch against Weston’s toytoolkit:
https://patchwork.freedesktop.org/patch/114661/

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -172,7 +172,10 @@ static void keyboardHandleKeymap(void* data,
 {
     struct xkb_keymap* keymap;
     struct xkb_state* state;
+    struct xkb_compose_table* composeTable;
+    struct xkb_compose_state* composeState;
     char* mapStr;
+    const char* locale;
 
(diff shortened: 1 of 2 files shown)


jadahl:
lgtm.

elmindreda:
> This commit has been copied almost verbatim

`process_key_press` and `getComposeSymbol` are indeed very similar. What license was the code under originally?

linkmauve:
Weston is licensed under MIT, which would require us to copy its header, it’s probably simpler to ask @bryceharrington to allow this code to be released under GLFW’s zlib.

It could also be possible that this code is considered trivial and thus usable without licensing restrictions, as it is the only way to use libxkbcommon, but I’m not absolutely sure about that.

elmindreda:
If you can get it re-licensed, that would be perfect.

bryceharrington:
I am the original author of the patch, and agree to its use under the zlib license.

elmindreda:
@bryceharrington Hello and thank you!

@linkmauve :+1:

PR #880, opened 2016-10-11
Set a proper name for outputs

linkmauve:
glfwGetMonitorName()’s documentation says “this function returns a
human-readable name”, which “typically reflects the make and model of
the monitor”.  We get these two strings in the geometry event, so we
replace the default name at this point.

### DIFF
--- a/src/wl_monitor.c
+++ b/src/wl_monitor.c
@@ -50,11 +50,21 @@ static void geometry(void* data,
                      int32_t transform)
 {
     struct _GLFWmonitor *monitor = data;
+    char* name;
+    size_t nameLength;
 
     monitor->wl.x = x;
     monitor->wl.y = y;
     monitor->widthMM = physicalWidth;
     monitor->heightMM = physicalHeight;
+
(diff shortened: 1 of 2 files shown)


src/wl_monitor.c
```diff
@@ -130,7 +136,8 @@ void _glfwAddOutputWayland(uint32_t name, uint32_t version)
         return;
     }
 
-    monitor = _glfwAllocMonitor(name_str, 0, 0);
+    // The actual name of this output will be set in the geometry handler.
+    monitor = _glfwAllocMonitor("wl_output", 0, 0);
```
jadahl: It's a bit awkward to have to pass a temporary value. Can't we just fix _glfwAllocMonitor() to only strdup if name is not NULL?

elmindreda: Fine by me.

linkmauve: I replaced this one with NULL, but kept the `realloc()` dance since outputs can get a new geometry with a potentially new make/model couple at any time.

PR #881, opened 2016-10-11
Add remarks to glfw3.h whenever the behaviour differs

linkmauve:
This should describe all of the Wayland-specific points, mostly in
three categories:
- Missing GLFW features, for example window frame, screensaver
  inhibition, clipboard or file drop.
- Missing protocols, like setting gamma on a monitor or specifying a
  window icon.
- Fundamental incompatibilities with Wayland’s concepts, like trying to
  handle global positioning of a window in a 2D space, trying to bring
  a window to front, or trying to change the monitor’s mode.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1621,6 +1621,9 @@ GLFWAPI const GLFWvidmode* glfwGetVideoMode(GLFWmonitor* monitor);
  *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED, @ref
  *  GLFW_INVALID_VALUE and @ref GLFW_PLATFORM_ERROR.
  *
+ *  @remark @wayland Gamma handling is currently unavailable, this function will
+ *  always emit @ref GLFW_PLATFORM_ERROR.
+ *
  *  @thread_safety This function must only be called from the main thread.
  *
  *  @sa @ref monitor_gamma
@@ -1641,6 +1644,9 @@ GLFWAPI void glfwSetGamma(GLFWmonitor* monitor, float gamma);
  *
  *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED and @ref
  *  GLFW_PLATFORM_ERROR.
+ *
+ *  @remark @wayland Gamma handling is currently unavailable, this function will
+ *  always return `NULL` and emit @ref GLFW_PLATFORM_ERROR.
  *
  *  @pointer_lifetime The returned structure and its arrays are allocated and
  *  freed by GLFW.  You should not free them yourself.  They are valid until the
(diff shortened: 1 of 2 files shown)


include/GLFW/glfw3.h
```diff
@@ -1621,6 +1621,9 @@ GLFWAPI const GLFWvidmode* glfwGetVideoMode(GLFWmonitor* monitor);
  *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED, @ref
  *  GLFW_INVALID_VALUE and @ref GLFW_PLATFORM_ERROR.
  *
+ *  @remark @wayland Gamma handling is currently unavailable, this function will
```
elmindreda: You need to define the `wayland` alias. See `ALIASES` in `docs/Doxyfile.in`.

include/GLFW/glfw3.h
```diff
@@ -2453,7 +2508,7 @@ GLFWAPI GLFWmonitor* glfwGetWindowMonitor(GLFWwindow* window);
  *
  *  When a window transitions from full screen to windowed mode, this function
  *  restores any previous window settings such as whether it is decorated,
- *  floating, resizable, has size or aspect ratio limits, etc..
+ *  floating, resizable, has size or aspect ratio limits, etc.
```
elmindreda: Oops!

include/GLFW/glfw3.h
```diff
@@ -2276,6 +2324,10 @@ GLFWAPI void glfwGetWindowFrameSize(GLFWwindow* window, int* left, int* top, int
  *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED and @ref
  *  GLFW_PLATFORM_ERROR.
  *
+ *  @remark @wayland There is no concept of iconification in wl_shell, this
+ *  function will always emit @ref GLFW_PLATFORM_ERROR.  TODO: move to
```
elmindreda: I'd prefer the todo be an issue tracker entry and that user-facing documentation describe only what is.

elmindreda:
If you want and have the time, in this PR or another, there may also be Wayland-specific material to add to `docs/compat.dox`.

include/GLFW/glfw3.h
linkmauve: Done, I should have built the documentation first. ^^'

include/GLFW/glfw3.h
linkmauve: True, this was more of a reminder to myself about implementing that, I guess I’ll open an issue for every remark I had to add to the documentation that was caused by a missing GLFW feature.

linkmauve:
This commit was only about glfw3.h, I will handle the rest of the documentation at some point, and might as well open an issue so I don’t forget. :)

include/GLFW/glfw3.h
```diff
@@ -2107,6 +2142,9 @@ GLFWAPI void glfwGetWindowSize(GLFWwindow* window, int* width, int* height);
  *  @remark If you set size limits and an aspect ratio that conflict, the
  *  results are undefined.
  *
+ *  @remark @wayland The size limits will not be applied until the window is
+ *  actually resized, either by the user or by the compositor.
```
elmindreda: This is a bug. If necessary, the limits [should be applied manually](https://github.com/glfw/glfw/blob/3.2.1/src/win32_window.c#L1151).

linkmauve: This is indeed a bug, and will be fixed before 3.3, like most of the other remarks I hope.

elmindreda:
Shortened the subject line to <= 50 characters.

Issue #816, opened 2016-07-29
AppSandbox warning on Mac


sulewicz:
I have built the GLFW 3.2 for Mac according to instructions.
When running a really simple project on Mac with AppSandbox enabled, I get the following warning in the log console:

2016-07-29 19:29:04.737 AppleSample[93301:25708587] **\* attempt to register for all distributed notifications thwarted by sandboxing.

Date/Time:     Fri Jul 29 19:29:04 2016
OS Version:    15E65
Application:   AppleSample

Backtrace:
0   CoreFoundation                      0x00007fff94ab357d __CFGenerateReport + 285
1   CoreFoundation                      0x00007fff94a2d1d2 __CFXNotificationRegisterObserver + 994
2   CoreFoundation                      0x00007fff9492ba4d _CFXNotificationRegisterObserver + 29
3   Foundation                          0x00007fff8acb91b8 -[NSDistributedNotificationCenter addObserver:selector:name:object:suspensionBehavior:] + 224
4   Foundation                          0x00007fff8acc32ff -[NSDistributedNotificationCenter addObserver:selector:name:object:] + 29
5   AppleSample                         0x0000000108dfdf17 _glfwPlatformInit + 199
6   AppleSample                         0x0000000108df714e glfwInit + 62
7   AppleSample                         0x0000000108df56c5 main + 37
8   libdyld.dylib                       0x00007fff916575ad start + 1

Has anyone run into this problem before?
I tried to contact Apple asking if some extra entitlements are required to make it work properly, but they weren't to helpful to be honest.

I have uploaded the sample project here: https://drive.google.com/file/d/0B4kVxFhoet_VRFd3ZlJXaERMT0k/view?usp=sharing

I am testing this on OSX 10.11.4 and XCode 7.3.

elmindreda:
I will look into whether it's possible to support sandboxing without compromising the feature set too much.

woc2006:
The same issue with latest version.

PR #882, opened 2016-10-14
Fix issue with sandboxing

tikhomirov:
Fixes the issue described in #816 (tested with sample sample application attached to the original issue).


elmindreda:
Merged as 11a241347699151131b7001470a3bb2245c08dc2 with additional cleanup in 887e58bd21587dba9fd685e1b80d44bbb0cec6df. Thank you!

Issue #876, opened 2016-10-07
Error on first responder set to nil on 10.12 Sierra


sp-erik-podetti:
[General] ERROR: Setting <private> as the first responder for window <private>, but it is in a different window ((null))! This would eventually crash when the view is freed. The first responder will be set to nil.
<private>

There is a way to avoid or fix this error? It comes with every example when creating the window. It wasn't occurring on max osx 10.11 (El Capitan)

elmindreda:
@sp-erik-podetti What is the error and how would I go about reproducing it?

HoMeBoYErik:
Just a basic main with default init functions on osx Sierra and you will get the message in the console.
For OpenGL headers I user glLoadGen with pointer_cpp spec

```
`int main(int argc, char* argv[])
{
    // Initialize GLFW
    if(!glfwInit())
    {
        exit(EXIT_FAILURE);
    }

#ifdef __APPLE__
    // Select OpenGL 4.1
    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 1);
#else
    // Select OpenGL 4.3
    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
#endif
    glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, gl::TRUE_);
    glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
    glfwWindowHint(GLFW_RESIZABLE, gl::FALSE_);
    glfwWindowHint(GLFW_OPENGL_DEBUG_CONTEXT, gl::TRUE_);

    // Open the window, get current resolution and set fullscreen
    std::string title = "My Window -- ";
    GLFWmonitor* monitor = glfwGetPrimaryMonitor();
    const GLFWvidmode* mode = glfwGetVideoMode(monitor);
    glfwWindowHint(GLFW_RED_BITS, mode->redBits);
    glfwWindowHint(GLFW_GREEN_BITS, mode->greenBits);
    glfwWindowHint(GLFW_BLUE_BITS, mode->blueBits);
    glfwWindowHint(GLFW_REFRESH_RATE, mode->refreshRate);
    
    window = glfwCreateWindow(800, 600, title.c_str(), nullptr, nullptr);

    if(!window)
[...]

PR #883, opened 2016-10-15
Fix 'first responder' error on 10.12

tikhomirov:
Error message is displayed during the startup because Cocoa view is set as the first responder for window but this view isn't in this window (actually it is not in any window at all).

Fixes the issue described in #876. **Tested only on macOS 10.12.**

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1032,11 +1032,11 @@ static GLFWbool createNativeWindow(_GLFWwindow* window,
     [window->ns.view setWantsBestResolutionOpenGLSurface:YES];
 #endif /*_GLFW_USE_RETINA*/
 
+    [window->ns.object setContentView:window->ns.view];
     [window->ns.object makeFirstResponder:window->ns.view];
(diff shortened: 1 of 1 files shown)


AlterionX:
I i i i 40oppl

On Oct 15, 2016 7:43 AM, "Sergey Tikhomirov" <notifications@github.com>
wrote:
>
> Error message is displayed during the startup because Cocoa view is set
as the first responder for window but this view isn't in this window
(actually it is not in any window at all).

elmindreda:
Instead of nested views, why not move the call to `NSWindow setContentView` to the top of the block?

elmindreda:
In either case, thank you for spotting the cause of the error!

tikhomirov:
> why not move the call to NSWindow setContentView

right, somehow I missed this option 😄 I've updated PR

elmindreda:
This has been merged as 967e837c73f0e0d7f9f27c7f6bb1aaa1d55dae54, with additional cleanup in 99c65bb67cf7cd705ab5ec67d0c96d9bc0bbe621. Thank you!

PR #888, opened 2016-10-26
Bad range handling of hat values

yerejm:
Joystick hat switch button in the up-left direction is always pressed in neutral position (at least on 10.11). It looks like on macOS the directions array should start with the neutral bitfield first rather than last like on Windows. This was tested using tests/joysticks.app. However, it has not been tested on other versions of macOS.

### DIFF
--- a/src/cocoa_joystick.m
+++ b/src/cocoa_joystick.m
@@ -246,7 +246,7 @@ static GLFWbool pollJoystickButtonEvents(_GLFWjoystickNS* js)
             CFArrayGetValueAtIndex(js->hatElements, i);
 
         // Bit fields of button presses for each direction, including nil
-        const int directions[9] = { 1, 3, 2, 6, 4, 12, 8, 9, 0 };
+        const int directions[9] = { 0, 1, 3, 2, 6, 4, 12, 8, 9 };
 
         long j, value = getElementValue(js, hat);
         if (value < 0 || value > 8)


elmindreda:
Thank you, will review ASAP.

elmindreda:
Cannot reproduce with a no-name 360 controller on 10.12.3 and other project I've checked interprets 0-8 in the same way as GLFW currently does. What hardware did you use?

yerejm:
Used Saitek Cyborg 3D USB (real) joystick. Unfortunately don't have a handheld controller.

elmindreda:
Are you using a OEM or third-party driver for it?

yerejm:
I am using whatever OSX provides as the driver. No specific driver from Saitek is being used.

elmindreda:
Sorry, I don't know what to do then, unless it's to hard-code a workaround for this specific device.

yerejm:
Hmm... it seems like more information (or hardware) is required. I don't have a handheld controller to compare against and it sounds like you don't have a joystick. I guess this issue is stuck.

elmindreda:
@yerejm Please check if current master still has this problem.

yerejm:
Still has same problem although the code looks different. It works for me only if center position is the first in the states array. However, perhaps this code in getElementValue might be the real problem:
```
71             // Record min and max for auto calibration
72             if (value < element->minimum)
73                 element->minimum = value;
```
`value` for center using the joystick is 0, but the minimum is 1. However, this changes the meaning of `value` since it is assumed to correspond with the states array that provides the hat direction. (Maybe `max` handling is questionable also?)

If that auto-calibration of `min` was removed and if the code in _glfwPlatformPollJoystick was like this:
```
390             long state = getElementValue(js, hat) - hat->minimum;
391             if (state < 0 || state > 8)
392                 state = 8;
```
The state condition would move the center value to the correct state index.

But again, I have no handheld controller to see what *its* minimum value is reporting. Since it works for you, perhaps it returns `0`, so everything works out?

elmindreda:
My controller reported the range [0,7] for the hat and sent 15 for the center position. Thank you for spotting this! It should be fixed now with 55d0560746632deb2df0807f4ba190383a353ad2.

Issue #889, opened 2016-10-27
Joystick hat is not a set of buttons


yerejm:
The handling of the joystick hat seems flawed.

Currently the 9 possible states are treated as distinct buttons, but this is problematic because the diagonal directions are detected as two buttons. For example, UP and UPLEFT cannot be distinguished simply via the API. The callback model proposed by #856 would pass through UPLEFT as two distinct button pushes for UP and LEFT. This may not be desirable.

If the caller is responsible for working out the hat state, then there isn't a simple way to distinguish between a joystick with a hat and a joystick without a hat when the button count matches. For example, a joystick with a primary and secondary trigger and a hat would show up as 6 buttons as would a joystick with a primary and secondary trigger and 4 extra buttons and no hat. I suppose the last four buttons could be tested with user input to determine this, but it seems drastic to ask the user for help when GLFW can figure it out already.

Also, the code for hat processing seems to indicate that multiple hats are possible.
Perhaps retrieved using `const int *glfwGetJoystickHats(int joystick_id, int *count)`?

So is it possible for the hat to not be treated as extra buttons but as its own thing? For example, `int glfwGetJoystickHat(int joystick_id, int hat_id)` returning the direction where direction is 0 for the neutral (centered) position and 1, 2, 3, etc for each of the possible directions (clockwise). A reason to not use the `GLFW_PRESSED`/`GLFW_RELEASED` style would be that it is impossible for the hat to be in more than one direction. In other words, the hat could be seen as being a single button, not a set of individual buttons.

A possible callback model API would take a function signature `void(*fn)(int joystick_id, int hat_id, int direction)`.

IntellectualKitty:
I've provided code to resolve this issue (pull request #906).

elmindreda:
This will be included in 3.3 using the public interface provided by @IntellectualKitty and @yerejm but with a slightly different implementation due to #900.

elmindreda:
Joystick hat support has been added with 368dec7ac712fc4a4a2f77aaec68b62fb8f98749 and 798d7c6d68bb17480fcda6074b1d5a946a05ed52, slightly messier than it should be due to backward compatibility.

IntellectualKitty:
@elmindreda thanks for adding this.  Again I apologize for the headaches it caused because I closed the original pull request.

elmindreda:
@IntellectualKitty No worries and thank you for an elegant patch!

Issue #900, opened 2016-11-20
Add direct support for gamecontrollerdb


CreamyCookie:
Currently using the [**gamecontrollerdb**](https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt) might not even be possible as ["as the joystick functions don't expose POVs separately from buttons (or axes, in the case of the Linux 360 controller driver)."](https://github.com/glfw/glfw/issues/278#issuecomment-256910460)

Additionally to that we need a method to get the GUID of a joystick, which gamecontrollerdb uses to differentiate between joysticks with the same name. Note that the GUID's there may be changing soon: https://github.com/gabomdq/SDL_GameControllerDB/issues/110

Preferably this should support both polling and events.

elmindreda:
The joystick functions don't expose POVs or GUIDs but GLFW uses the same underlying APIs as SDL2, so the internals will push the native data to both the joystick and controller functions and each will translate and present it after its own fashion.

elmindreda:
This work has begun in the `gamecontrollerdb` branch.

elmindreda:
- [ ] Joystick state handled by shared code
- [ ] Support for joystick hats
- [ ] Support for joystick GUIDs
- [ ] THE GREAT AND HORRIFYING SDL2 ELEMENT ORDER MATCHING
- [ ] Support for hats on Linux
- [ ] SDL_GameControllerDB parsing
- [ ] Element remapping logic
- [ ] Public gamepad state struct
- [ ] Public gamepad functions

elmindreda:
Parsing, matching, remapping and state retrieval are working. There's still a little work left to do, but most of it is done.

Below is the raw state of an Xbox 360 controller on Linux (top) and filtered through a mapping (bottom) from the current `gamecontrollerdb.txt` file.

![mapping](https://user-images.githubusercontent.com/133714/27480294-a58485f2-5817-11e7-8eb0-c62a427c46cf.png)

The current (but not necessarily final) API is as follows:
```c
#define GLFW_GAMEPAD_DPAD_UP        0
#define GLFW_GAMEPAD_DPAD_RIGHT     1
#define GLFW_GAMEPAD_DPAD_DOWN      2
#define GLFW_GAMEPAD_DPAD_LEFT      3
#define GLFW_GAMEPAD_A              4
#define GLFW_GAMEPAD_B              5
#define GLFW_GAMEPAD_X              6
#define GLFW_GAMEPAD_Y              7
[...]

PR #906, opened 2016-11-26
Joystick Hat Support

IntellectualKitty:
This code implements joystick hat support per issue #889.

The API change is to add the following function (and #defines for each of the hat positions such as GLFW_HAT_RIGHT_UP):

    GLFWAPI const unsigned char* glfwGetJoystickHats(int jid, int* count);

The events test (events.c) has been updated to reflect the number of hat switches when a joystick is connected.

The joystick test (joystick.c) has been updated to add a new row for hat switches that displays the position in human-readable text (e.g., "right_up").

Code for macOS has been provided and tested.

Code for Windows has been provided, but has not been compiled or tested since I do not have access to a Windows system.

Code for Linux has also been provided, but relevant portions are compiled out with the #defines JSIOCGHATS and JS_EVENT_HAT -- following the JSIOCGAXES / JSIOCGBUTTONS and JS_EVENT_AXIS / JS_EVENT_BUTTON syntax, respectively -- since <linux/joystick.h> does not currently provide support for hat switches; if such support is added to <linux/joystick.h>, the code should work however.  This code also has not been compiled or tested since I do not have access to a Linux system.

A remark about the lack of support for hat switches on Linux appears in the documentation for glfwGetJoystickHats in glfw3.h.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -286,6 +286,23 @@ extern "C" {
 #define GLFW_REPEAT                 2
 /*! @} */
 
+/*! @defgroup hat_directions Joystick hat directions
+ *
+ *  See [joystick hat input](@ref joystick_hat) for how these are used.
+ *
+ *  @ingroup input
+ *  @{ */
+#define GLFW_HAT_CENTERED           0
+#define GLFW_HAT_UP                 1
+#define GLFW_HAT_RIGHT              2
+#define GLFW_HAT_DOWN               4
+#define GLFW_HAT_LEFT               8
+#define GLFW_HAT_RIGHT_UP           (GLFW_HAT_RIGHT | GLFW_HAT_UP)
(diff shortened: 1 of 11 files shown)


elmindreda:
This should definitely be supported and I like the interface. The implementation will be somewhat different both due to the work being done in the `gamecontrollerdb` branch for #900 and because pulling hats out of the button array will break backward compatibility.

Instead I propose being able to query hat mappings, where each hat is four button indices. Every platform puts the hat buttons at the end of the array already, which should help people transition. Once it's time for GLFW 4, of course, this can be cleaned up.

Waiting for the Linux joystick API to change is probably futile, but have a look at evdev.

elmindreda:
It's annoying to work with a closed issue. I've opened a replacement.

elmindreda:
Linux evdev joystick input is pending in #1005.

elmindreda:
Linux evdev joystick input has been merged. Hats are now properly supported on Linux (X11, Wayland and Mir).

PR #912, opened 2016-12-06
Crash in Debian 9 on glfwCreateWindow

paul4334:
Fix crash when glXGetClientString returns a null string

### DIFF
--- a/src/glx_context.c
+++ b/src/glx_context.c
@@ -59,7 +59,7 @@ static GLFWbool chooseGLXFBConfig(const _GLFWfbconfig* desired, GLXFBConfig* res
     // HACK: This is a (hopefully temporary) workaround for Chromium
     //       (VirtualBox GL) not setting the window bit on any GLXFBConfigs
     vendor = glXGetClientString(_glfw.x11.display, GLX_VENDOR);
-    if (strcmp(vendor, "Chromium") == 0)
+    if (vendor && strcmp(vendor, "Chromium") == 0)
         trustWindowBit = GLFW_FALSE;
 
     nativeConfigs =


elmindreda:
Does everything else work as expected with this change?

elmindreda:
Judging by #913 this is not the case.

paul4334:
No, I am trying to track down whether it is a driver configuration issue causing GLX to not work correctly, but this change does allow the application to exit gracefully under these circumstances

elmindreda:
@paul4334 Is this a general problem with that version of Debian or is it specific to your machine?

paul4334:
I tried to reproduce the problem on a normal HDD install of Debian 9 and did not have the issue, so I think it is limited to a live-boot ISO created with [https://www.debian.org/devel/debian-live/](debian live) with the newest packages. I think my submission #913 is superfluous, but regardless, the central problem, IMHO, is that glXGetClientString is returning a nullptr and it is being passed into strcmp.

elmindreda:
I don't want to add (and maintain) a workaround for a driver bug in a pre-release live CD, especially when it doesn't even work around the problem.

drivehappy:
There might be some confusion here: #913 had been opened to make additional notes of the environment behavior when just running glxinfo (and not GLFW).

This change does solve the problem on our side where GLFW is causing our application to crash in this environment. I think it's safe to always check the return value of glXGetClientString in case it is NULL. The docs here seem to indicate there are conditions on when it would return NULL as a valid value: https://www.opengl.org/sdk/docs/man2/xhtml/glXGetClientString.xml

elmindreda:
@drivehappy Every workaround like this adds cumulatively to the cost of maintaining the library. I need to weigh the cost against the benefit. The benefit here seems extremely small and specific, and the cost of testing whether the workaround is intact after future changes seems very high (even ignoring that I don't have a Pascal card). Since Debian 9 isn't out yet, I think it would be better to report this issue to the package maintainer and have everyone benefit from the fix in future revisions of the Debian 9 live CD.

zao:
FWIW, anything that tries to speculatively use `glXGetClientString` during most kinds of X forwarding will also trigger this problem. We had a case where plain boring emacs would crash as their X mode uses parts of GTK+ that uses a silly library that doesn't check the return of this function.

(there's two kinds of X forwarding, one where each client runs in its own little world and GLX isn't exposed, and one where pretty much every extension of the server is forwarded, regardless if they work or not)

glfwinfo too explodes in such a configuration, but I can't really envision an application that depends on GLFW that would work in such scenarios, I don't see too much point in patching it up on the GLFW side.
It's likely that the only thing you can do once you realize that GLX is non-functional is to quit gracefully.

elmindreda:
@zao Yes, agreed.

@drivehappy @paul4334 If you are deploying your application on this specific CD or your users are running it, it shouldn't be a lot of work to maintain this one-line patch as you seem to have the setup for testing it already (although please correct me if I'm wrong). If the issue remains (or looks likely to remain) in the final release version, I'll gladly add a workaround for it. A stable Debian release is a sufficiently large and slow-to-update target to warrant a workaround. In the meantime, please report this issue to Debian so that won't happen.

elmindreda:
@zao I seem to remember someone at the X Developer's Conference a few years back mentioning that GLX over X forwarding had been broken for a while and no one had noticed so perhaps they didn't need to worry about breaking it with... DRI3, I think? Haven't tried myself for years. Did you have any luck getting anything else GLX-related running on that setup?

drivehappy:
@elmindreda We can apply the patch to our build, however it is a minor maintenance burden on our end if we need to update.

I'm bit surprised this patch is causing this much discussion. I don't consider this change a workaround, my view on the addition of the NULL check is that is defensive programming, and much less severe than the original workaround already in place that introduced the call. Additionally, the docs for the glXGetClientString have clearly stated that a NULL return value is valid - to purposefully leave a potential segfault around the existing Chromium workaround seems much more dangerous.

elmindreda:
>  the original workaround

@drivehappy Thank you for reminding me! I'd forgotten to check up on that one for a while. Seems it was [fixed in VirtualBox 4.3](https://www.virtualbox.org/ticket/8329) three years ago and has even made it into Debian stable, so it might be time to remove the original workaround.

zao:
@elmindreda None of the tests or examples from GLFW ran on that configuration (Ubuntu 16.04), and I believe we ended up advising the user to run their emacs in text mode. X forwarding at our site is just a if-it-works-it-works convenience, for actual graphical applications we run ThinLinc.

As a side note, everything runs fine on the previous cluster that has Ubuntu 12.04.

elmindreda:
@drivehappy It turns out I was already doing checking [other similar strings](https://github.com/glfw/glfw/blob/3.2.1/src/glx_context.c#L199), so should have been doing it here as well all along. I'll merge the fix shortly. Thank you for your patience. 🦉

I may also later transform these checks into an earlier and explicit test with a more useful error message since it can apparently happen in other situations as well.

PR #915, opened 2016-12-12
X11: send timestamp of last key press to NET_ACTIVE_WINDOW

hesiod:
The EWMH specifications require that the timestamp of the last
user interaction should be sent to `NET_ACTIVE_WINDOW`.
Since GLFW send just 0 currently, the GNOME Shell would issue a warning
upon receiving a `NET_ACTIVE_WINDOW` message:

> Window manager warning: Buggy client sent a _NET_ACTIVE_WINDOW message
> with a timestamp of 0

This commit fixes this by sending `lastKeyTime` as the timestamp
which should be a reasonable approximation to the last user interaction.

I have not tested this patch extensively, but it silences the warning and it should not alter existent behaviour.

### DIFF
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -1945,7 +1945,8 @@ void _glfwPlatformHideWindow(_GLFWwindow* window)
 void _glfwPlatformFocusWindow(_GLFWwindow* window)
 {
     if (_glfw.x11.NET_ACTIVE_WINDOW)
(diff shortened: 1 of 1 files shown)


elmindreda:
I started on a full implementation of this requirement a while back in the [x11-event-time](https://github.com/glfw/glfw/tree/x11-event-time) branch. Note that it is not complete.

PR #895, opened 2016-11-05
Command-tabbing fix

IntellectualKitty:
The new code uses presentation options to properly configure the application for fullscreen operation when appropriate.

Also, the window id is supplied as a parameter to the orderFront, makeKeyAndOrderFront, and orderOut calls instead of nil, which I *believe* is more correct.

Issue #848, opened 2016-08-26
Command-tabbing issues with fullscreen windows


IntellectualKitty:
I used the code directly from the HTML guide for creating "windowed full screen" windows to modify the Boing demo application.  Specifically, I replaced this line:
```c
   window = glfwCreateWindow( 400, 400, "Boing (classic Amiga demo)", NULL, NULL );
```
with the following code:
```c
   GLFWmonitor* monitor = glfwGetPrimaryMonitor();
   const GLFWvidmode* mode = glfwGetVideoMode(monitor);
   glfwWindowHint(GLFW_RED_BITS, mode->redBits);
   glfwWindowHint(GLFW_GREEN_BITS, mode->greenBits);
   glfwWindowHint(GLFW_BLUE_BITS, mode->blueBits);
   glfwWindowHint(GLFW_REFRESH_RATE, mode->refreshRate);
   window = glfwCreateWindow( mode->width, mode->height, "Boing (classic Amiga demo)", monitor, NULL );
```
The fullscreen window is created correctly and initially works fine, but problems ensue when command-tabbing from and then back to the Boing application.  First, when command-tabbing from Boing to another application, the fullscreen window is iconified rather than simply hidden.  Second, a number of issues occur when command-tabbing back to the Boing application: (1) the fullscreen window remains iconified and does not reappear; (2) when you click on the Boing application in the Dock, the fullscreen window is de-iconified, but then is immediately hidden (and remains hidden) leaving you with only the menu bar; (3) if you then click anywhere but the menu bar, the fullscreen window re-appears, but then is immediately iconified.

This, at least, is the general behavior.  However, if you repeatedly command-tab back to the Boing application (or try to activate it by clicking on its icon in the Dock), the fullscreen window will sometimes no longer re-appear at all.

Desirable behavior would be that the fullscreen window is simply hidden and then restored (rather than being iconified and de-iconified) when command-tabbing from and to the Boing application.  Also, it would be desirable to be able to disable command-tabbing when in fullscreen mode, particularly per a user's request.

This test was performed with the latest GLFW code (downloaded directly from GitHub on August 26, 2016) on a late 2011 MacBook Pro running OS X 10.11.6 and built with Xcode 7.3.1.

trustin:
Having the same issue. Probably related with #503?

elmindreda:
Full screen window restoration should work as expected now. Spaces-unaware full screen with minimization is of course horrible even when it works and is what I was really looking into improving upon when I found this fix.

PR #926, opened 2016-12-28
Command-tabbing fix for issue #848

IntellectualKitty:
I'm re-submitting the command-tabbing fix.  The code for the presentation options is identical to what I submitted with PR #895, but I've returned to the original code that you had using nil for the sender id in the `orderFront` and `makeKeyAndOrderFront` calls.

I've found it difficult to get good information about the sender id, but my impression from what I have come across is that the sender id is most commonly used when coordinating with the Interface Builder (IB), which is not the case here.  Since most of the Cocoa code that I've seen in the past came from Apple, which seems to use IB exclusively, I mistakenly thought that it was necessary to provide the window as the sender id.  However, it seems that nil is clearly appropriate in this case.

I've also been having difficulty finding detailed information on presentation options, but here is Apple's documentation: https://developer.apple.com/reference/appkit/nsapplicationpresentationoptions?language=objc

The following presentation options are used in fullscreen mode (i.e., when there is a monitor attached to the window):
```
            NSApplicationPresentationHideDock
            NSApplicationPresentationHideMenuBar
            NSApplicationPresentationDisableProcessSwitching
            NSApplicationPresentationDisableHideApplication
```
There is a fullscreen option (`NSApplicationPresentationFullScreen`), which you would think should be ideal and even necessary for fullscreen support, but I did not include it because it caused buggy behavior (which was actually quite similar to the erroneous behavior that I reported in issue #848).  From the Googling that I did about the fullscreen option, it seems like other people are encountering buggy behavior with it as well.

When there is no monitor attached to the window, the option `NSApplicationPresentationDefault` is set explicitly to return to normal behavior.

This change occurs in the following three functions and is bracketed in an if/else statement based on the presence of `window->monitor` and kept separate from the existing code:
```
            _glfwPlatformShowWindow
            _glfwPlatformFocusWindow
            _glfwPlatformSetWindowMonitor
```

The only exception is in `_glfwPlatformShowWindow`, where `makeKeyAndOrderFront` is used instead of `orderFront` when the window is being shown in fullscreen mode.

Finally, while this change addresses the issues I mentioned in #848, I'm not sure if these changes match GLFW's behavior on other platforms nor if it is always the ideal behavior to disable command-tabbing in fullscreen mode.  I don't recall anything descriptive about this in the documentation, so if you can provide me with feedback, I'll be glad to update the code and resubmit, if necessary.  Otherwise, I've tried to follow the contributing guidelines and added a line about this bugfix in README.me, which I hope is helpful.  Also, per our recent discussions, I hope the additional detailed information about how this code works is helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ information on what to include when reporting a bug.
 - [Cocoa] Bugfix: Disabled use of deprecated `CGDisplayModeCopyPixelEncoding`
                   function on macOS 10.12+
 - [Cocoa] Bugfix: Running in AppSandbox would emit warnings (#816,#882)
+- [Cocoa] Bugfix: Command-tabbing issues with fullscreen windows (#848)
 - [EGL] Added support for `EGL_KHR_get_all_proc_addresses` (#871)
 
 
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1248,7 +1248,25 @@ void _glfwPlatformMaximizeWindow(_GLFWwindow* window)
 
 void _glfwPlatformShowWindow(_GLFWwindow* window)
 {
-    [window->ns.object orderFront:nil];
+    if (window->monitor)
+    {
+        [window->ns.object makeKeyAndOrderFront:nil];
+
+        NSApplicationPresentationOptions options =
+            NSApplicationPresentationHideDock +
+            NSApplicationPresentationHideMenuBar +
+            NSApplicationPresentationDisableProcessSwitching +
+            NSApplicationPresentationDisableHideApplication;
+        [NSApp setPresentationOptions:options];
+    }
+    else
+    {
+        [window->ns.object orderFront:nil];
+
+        NSApplicationPresentationOptions options =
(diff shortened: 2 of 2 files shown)


elmindreda:
> I'm not sure if these changes match GLFW's behavior on other platforms nor if it is always the ideal behavior to disable command-tabbing in fullscreen mode.

Addressing just this bit: it should always be possible to switch away from full screen windows unless that isn't supported by the OS.

elmindreda:
This patch builds, runs and solves the problem but does so by disabling expected functionality, i.e. Cmd+Tab, switching between spaces, etc. I will not merge it in its current state.

elmindreda:
I stumbled upon a solution in 3d8aa53573d488e45fa4a9f8ac40a6893386db63. This doesn't improve the current and weird full screen behavior, but it does appear to make that behavior work as designed.

I would love to have a discussion about how full screen windows _should_ behave on macOS, especially now that we have the additional tool of platform specific hints for behavior that just doesn't generalize to other systems.

PR #930, opened 2017-01-07
Fix build on OpenBSD

akamch:
OpenBSD [doesn't have librt](http://marc.info/?l=openbsd-tech&m=139559970907629), but provides all librt's functions in libc. Here's a shot for a fix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,11 +199,13 @@ endif()
 # Find and add Unix math and time libraries
 #--------------------------------------------------------------------
 if (UNIX AND NOT APPLE)
-    find_library(RT_LIBRARY rt)
-    mark_as_advanced(RT_LIBRARY)
-    if (RT_LIBRARY)
-        list(APPEND glfw_LIBRARIES "${RT_LIBRARY}")
-        list(APPEND glfw_PKG_LIBS "-lrt")
+    if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD")
+	find_library(RT_LIBRARY rt)
+	mark_as_advanced(RT_LIBRARY)
+	if (RT_LIBRARY)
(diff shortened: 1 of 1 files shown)


elmindreda:
I don't see the problem. If the stub librt isn't there, it won't be added as a dependency.

elmindreda:
How does the build fail? It would be useful to see your `CMakeFiles/CMakeOutput.log`.

akamch:
The build fails with this error:
```
-- Looking for IceConnectionNumber in ICE
-- Looking for IceConnectionNumber in ICE - found
-- Found X11: /usr/X11R6/lib/libX11.so.16.1
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
RT_LIBRARY (ADVANCED)
    linked by target "particles" in directory /home/rev/opengl/glfw/examples
    linked by target "threads" in directory /home/rev/opengl/glfw/tests
    linked by target "empty" in directory /home/rev/opengl/glfw/tests

-- Configuring incomplete, errors occurred!
See also "/home/rev/opengl/glfw/build/CMakeFiles/CMakeOutput.log".
See also "/home/rev/opengl/glfw/build/CMakeFiles/CMakeError.log".
```
`CMakeOutput.log`: http://pastebin.com/raw/bJ7v3Pr3

akamch:
OpenBSD currently packages glfw 3.1.2 and similarly [patches out librt](http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/graphics/glfw/patches/patch-CMakeLists_txt?rev=1.1&content-type=text/x-cvsweb-markup), though indeed, it may be better to fix checks for librt in `{examples,tests}/CMakeLists.txt`. I can file a new issue if you wish.

elmindreda:
I'll have a fix merged this week.

PR #935, opened 2017-01-25
Support for automatic graphics switching

kkaefer:
Another attempt at #377:

There is a way to control graphics switching behavior on macOS without modifying the plist file (many GLFW apps are outside an app bundle): `NSOpenGL` is backed by `CGL` on macOS. Instead of adding `NSOpenGLPFAAllowOfflineRenderers`, and adding the plist setting, you can also add `kCGLPFASupportsAutomaticGraphicsSwitching` to the `NSOpenGLPixelFormatAttribute` array. Of course, once Apple decides to change the underlying implementation, this might fail in spectacular ways.

However, there seems to be an officially supported API to accomplish this: [`[NSOpenGLPixelFormat initWithCGLPixelFormatObj:]`](https://developer.apple.com/reference/appkit/nsopenglpixelformat/1436129-initwithcglpixelformatobj?language=objc) takes a `CGLPixelFormatObj`.

This pull request changes the initialization of the pixel format object from `NS*` to `CGL*` (it supports the same enums, just with a different prefix), then adds a flag `GLFW_ALLOW_GRAPHICS_SWITCHING` that allows opting in to graphics switching.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,12 +341,13 @@ if (_GLFW_COCOA)
 
     list(APPEND glfw_LIBRARIES
         "-framework Cocoa"
+        "-framework OpenGL"
         "-framework IOKit"
         "-framework CoreFoundation"
         "-framework CoreVideo")
 
     set(glfw_PKG_DEPS "")
-    set(glfw_PKG_LIBS "-framework Cocoa -framework IOKit -framework CoreFoundation -framework CoreVideo")
+    set(glfw_PKG_LIBS "-framework Cocoa -framework OpenGL -framework IOKit -framework CoreFoundation -framework CoreVideo")
 endif()
 
 #--------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


elmindreda:
Thank you. Will review.

elmindreda:
> Instead of adding NSOpenGLPFAAllowOfflineRenderers, and adding the plist setting, you can also add kCGLPFASupportsAutomaticGraphicsSwitching to the NSOpenGLPixelFormatAttribute array.

Seems that the UnrealEngine 4 standalone renderer already does it this way.

elmindreda:
Thank you so much for finding the missing piece for making this a reliable window hint!

I (now) have on good authority that it should be all right to pass `kCGLPFASupportsAutomaticGraphicsSwitching` directly so went with the simpler solution, but that is no criticism of your code. This patch is solid.

PR #937, opened 2017-01-29
Fix incomplete initialization of a matrix in linmath.h

cher-nov:
Introduced by a28236d939b70ade9b25cdaade46bd2c59bf7385.
Related commit comment: https://github.com/glfw/glfw/commit/a28236d939b70ade9b25cdaade46bd2c59bf7385#commitcomment-20651393

### DIFF
--- a/deps/linmath.h
+++ b/deps/linmath.h
@@ -197,12 +197,10 @@ static inline void mat4x4_rotate(mat4x4 R, mat4x4 M, float x, float y, float z, 
 		vec3_norm(u, u);
 		mat4x4_from_vec3_mul_outer(T, u, u);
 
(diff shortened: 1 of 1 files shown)


cher-nov:
@elmindreda, oh, your solution is really better.
For unknown reason I have thought that old VC++'s don't support designated initializers.

elmindreda:
Even C89 allows [flat initializer lists](http://en.cppreference.com/w/c/language/array_initialization) for multidimensional arrays, which is very handy for cases like this.

Thank you for reporting the bug!

cher-nov:
@elmindreda, yes, I know, of course. It was just a fleeting mistake. :)

Issue #42, opened 2013-06-13
Touch input


elmindreda:

nippoo:
Is this still slated for 3.2? :)

elmindreda:
I hope so.

elmindreda:
[Multitouch](http://www.x.org/wiki/Development/Documentation/Multitouch/)

ec1oud:
Looks like not much has been happening with this lately, although on the touch branch it looked to be off to a good start (on Windows only?)  Having some second thoughts about the approach, or just too much work/not enough time, or because there is a solution provided by some other library?

elmindreda:
@ec1oud For my part, not enough time and no hardware to test on.

NeatSketch:
@elmindreda I have a touchscreen laptop with Windows 10 & Ubuntu installed. Is there any way I can help with testing?

Renanse:
Interested in this as well.  Would love to be able to implement glfw-backed multi-touch input in Ardor3D.  Right now I'm only able to implement this in SWT land. :/

torkeldanielsson:
I played around with this and did a rebase of the touch branch on top of the latest master:
https://github.com/torkeldanielsson/glfw/tree/touch

It isnt't so clean (not every commit builds, there is a fixup commit that fixes misses during the rebase). End state should be as intended though. ..and let me know if there is an interest in a cleaner rebase and I can fix it :)

Thought: from "WM_TOUCH is totally bananas" it seems like maybe WM_TOUCH is not the right choice? I would guess win8/win10 is a fine target for this. Or is it that WM_TOUCH is more similar to what is available on Linux?

cypherdare:
I don't currently have a touch screen to test. But does the existing GLFW treat screen touches as a mouse click and drag, or is it ignored?

rseymour:
I have a razer blade stealth 2018 w/ a QHD screen and Ubuntu Studio and I'm also up for testing.

raysan5:
Any progress on multi-touch support for desktop platforms? Is it even possible coding in C?

alectron:
Random thought - with VCV (a music / modular eurorack style software ) requiring this to be addressed on glfw officially to make use of multi-touch for VCV... and the VCV community being quite active and there being regular requests for touch support: 
[...]

PR #952, opened 2017-02-23
Updated touch structure

eriksunden:
I wanted touch support inside GLFW, for convenience over software setups like TUIO, so continued on your touch branch, and based on previous experiences with multi-touch applications, I have bundled all the touch points into one container before sending them out through the callbacks.

I have merged to "latest" and then later on to "master", so it's up-to-date for a convenient merge (or to new feature branch) and possibly then part of 3.3 (as I saw in an issue was the plan).
I have built this branch with Visual Studio 2015 x64, XCode 8.2.1, gcc-4.8(on two linux systems).

The actual touch support is, as with your previous commits, only supporting WM_TOUCH, i.e. Windows. If desired, I could gather effort on implementing support on x11 and cocoa side.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -272,18 +272,18 @@ extern "C" {
 #define GLFW_FALSE                  0
 /*! @} */
 
-/*! @name Key and button actions
+/*! @name Key, button and touch actions
  *  @{ */
-/*! @brief The key or mouse button was released.
+/*! @brief The key or button was released or the touch ended.
  *
- *  The key or mouse button was released.
+ *  The key or button was released or the touch ended.
  *
  *  @ingroup input
  */
 #define GLFW_RELEASE                0
-/*! @brief The key or mouse button was pressed.
+/*! @brief The key or button was pressed or the touch started.
  *
- *  The key or mouse button was pressed.
+ *  The key or button was pressed or the touch started.
  *
  *  @ingroup input
  */
@@ -295,6 +295,10 @@ extern "C" {
  *  @ingroup input
  */
 #define GLFW_REPEAT                 2
+/*! @brief The touch was moved.
+ *  @ingroup input
+ */
+#define GLFW_MOVE                   3
 /*! @} */
 
 /*! @defgroup keys Keyboard keys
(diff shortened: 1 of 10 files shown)


include/GLFW/glfw3.h
```diff
@@ -870,6 +874,7 @@ extern "C" {
 #define GLFW_CURSOR                 0x00033001
 #define GLFW_STICKY_KEYS            0x00033002
 #define GLFW_STICKY_MOUSE_BUTTONS   0x00033003
+#define GLFW_TOUCH                  0x00030004
```
tombsar: Should this value not be 0x00033004 to match the surrounding defines?

src/win32_window.c
```diff
@@ -810,6 +851,75 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
...
+            if (_glfw_GetTouchInputInfo((HTOUCHINPUT) lParam,
+                                        count, inputs, sizeof(TOUCHINPUT)))
+            {
+                UINT i;
+                int width, height, xpos, ypos;
+
+                _glfwPlatformGetWindowSize(window, &width, &height);
+                _glfwPlatformGetWindowPos(window, &xpos, &ypos);
+
+				//Create storage
```
tombsar: Indenting is weird here and elsewhere in this file.

tests/events.c
```diff
@@ -482,6 +494,20 @@ static void joystick_callback(int jid, int event)
     }
 }
 
+static void touch_callback(GLFWwindow* window, GLFWtouch* touchPoints, int count)
+{
+	printf("Priting info about all touch points");
```
tombsar: Typo: "priting" vs "printing"

elmindreda:
@eriksunden Thank you, will review.

elmindreda:
This is related to #42.

tests/events.c
eriksunden: Sorry for these flaws. Is it preferred that I fix these things or do any of you desire to do it yourself?

include/GLFW/glfw3.h
elmindreda: It should, though that mistake was not introduced by @eriksunden.

elmindreda:
It's not a bad-looking interface, but it is more complex than the current one. Is there often more than one related touch event per frame?

eriksunden:
Sorry for the ultra late reply.
Their is sort of several touch events per frame, if you consider one touch event to be equal to one finger, which one really shouldn't. One event per frame, with a consideration of all touch points that current resides on the touch surface is needed to perform a correct interaction response (even the fingers/touch points which have not changed).
Before you have a list of all touch points on the surface, you don't really now what kind of interaction the user has made, which can in one frame actually can be multiple interactions, if for instance the user had 2 fingers on the surface and performed a rotation and pinch.
There is even touch screens which support pressure, also an attribute part of most low-level touch API such as Windows Touch.

PR #955, opened 2017-02-26
Add support for the idle-inhibit protocol

linkmauve:
This makes every window disable the screensaver when they are displayed, on supported compositors.

This PR has only been tested against Weston using this (as of yet unmerged) series: https://patchwork.freedesktop.org/series/4875/

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -189,6 +189,24 @@ static void setOpaqueRegion(_GLFWwindow* window)
     wl_region_destroy(region);
 }
 
+static void setIdleInhibitor(_GLFWwindow* window, GLFWbool enable)
+{
+    if (enable && !window->wl.idleInhibitor && _glfw.wl.idleInhibitManager)
+    {
+        window->wl.idleInhibitor =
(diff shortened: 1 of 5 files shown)


elmindreda:
I like how this is an actual object on Wayland.

Note that only non-iconified full screen windows should inhibit the screensaver, currently.

jadahl:
The inhibitation protocol is designed so that only visible surfaces will affect idle inhibitation, so there is no need to deal with being iconified vs non-iconified.

src/wl_window.c
```diff
@@ -208,6 +208,13 @@ static GLFWbool createSurface(_GLFWwindow* window,
     if (!window->wl.native)
         return GLFW_FALSE;
 
+    window->wl.idleInhibitor =
```
jadahl: This must only be done if _glfw.wl.idleInhibitManager is set, i.e. the compositor supports the protocol.

linkmauve:
Hi, I fixed all of the comments, this should be good to go now.

PR #973, opened 2017-03-18
add input lag test

zholos:
This is a new test program – `tests/inputlag.c` – to measure input lag, i.e. the time between new input being polled and it's effect getting displayed.

inputlag

It renders a marker at the cursor position reported by GLFW so you can see how much it lags behind the hardware mouse cursor, as well as forecasts for 1-3 frames ahead based on mouse movement speed to get an exact measurement of lag in frames. E.g. in the screenshot input lag is 1 frame.

The main purpose is to find the best way to write a main loop with vsync enabled and CPU as well as GPU synchronized to it, so that input lag is bounded at 1 frame. I think for games this is a good balance of reasonably low input lag and no tearing. But by default the CPU is not synchronized (`glfwSwapBuffers` is non-blocking), so the CPU can queue up too many frames that become stale by the time they're displayed. Consider this main loop:

```c
glfwSwapInterval(1);
while (!glfwWindowShouldClose(window))
{
    render();
    glfwSwapBuffers(window);
    glfwPollEvents();
}
```

The timing of these calls on a 60 Hz monitor might be:

```c
render();                // t=0.000
glfwSwapBuffers(window); // t=0.001, enqueues frame to be displayed at t=0.017, doesn't block
glfwPollEvents();        // t=0.002
render();                // t=0.003, blocks when command queue fills up until t=0.017
glfwSwapBuffers(window); // t=0.018, enqueues frame to be displayed at t=0.033
glfwPollEvents();        // t=0.019
```

Notice that the input we polled at `t=0.002` only gets displayed at `t=0.033` – almost 2 frames of input lag for no benefit!

The OpenGL wiki page [Swap Interval](https://www.khronos.org/opengl/wiki/Swap_Interval) recommends using `glFinish` after swapping buffers to achieve what I'm trying to do, but I've found that it doesn't always work. This test allows trying a couple of other methods to force synchronization, in particular fetching results from an occlusion query.

Besides this, it's a test for input handling. For instance, I've found that on X11 data comes through `glfwSetCursorPosCallback` a bit late compared to the synchronous `glfwGetCursorPos`.

And it can be used to test any future features that help reduce input lag. A good candidate is a function to expose [GLX_NV_delay_before_swap](https://www.khronos.org/registry/OpenGL/extensions/NV/GLX_NV_delay_before_swap.txt).

I used code from `tests/tearing.c` and `tests/gamma.c` as a basis.

Some observations from the test so far:

* disabling vsync makes input lag 0 on X11 and in fullscreen mode on Windows
* enabling vsync with `glFinish` on an Nvidia GPU makes input lag 1 frame
* enabling vsync without `glFinish` makes input lag 2 frames (as in the timing example above)
* `glFinish` doesn't work for this purpose on an Intel GPU, but fetching results from an occlusion query does
* enabling compositing in the window manager adds 1 frame of input lag

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -30,6 +30,7 @@ add_executable(cursor cursor.c ${GLAD})
 add_executable(empty WIN32 MACOSX_BUNDLE empty.c ${TINYCTHREAD} ${GLAD})
 add_executable(gamma WIN32 MACOSX_BUNDLE gamma.c ${GLAD})
 add_executable(icon WIN32 MACOSX_BUNDLE icon.c ${GLAD})
+add_executable(inputlag WIN32 MACOSX_BUNDLE inputlag.c ${GETOPT} ${GLAD})
 add_executable(joysticks WIN32 MACOSX_BUNDLE joysticks.c ${GLAD})
 add_executable(sharing WIN32 MACOSX_BUNDLE sharing.c ${GETOPT} ${GLAD})
 add_executable(tearing WIN32 MACOSX_BUNDLE tearing.c ${GETOPT} ${GLAD})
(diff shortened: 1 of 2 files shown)


shurcooL:
I have no comments on whether this belongs here (@elmindreda will know better), but I just wanted to say this is very interesting, thank you for sharing. 👍

elmindreda:
This is really nifty! I'll be merging it into the Nuklear rewrite of `tests/cursor.c`.

zholos:
Great!

shurcooL:
I wanted to share this relevant article https://www.blurbusters.com/4k-120hz-with-bonus-240hz-and-480hz-modes/. For anyone interested in this topic, it should be a fascinating read.

PR #979, opened 2017-03-29
Get ready for Mir 1.0 by cleaning up deprecations

BrandonSchaefer:
Quite a removal. Then main diff +/- is changing the name of MirSurface to MirWindow.

As well as the display config change from using a defined structure to an opaque pointer. As well as other minor changes.

Cleaned up CMake so we can force only mir 0.26.2 or greater will be allowed to be compiled against.

There was a change to a public header MirSurface -> MirWindow but those are both opaque 

Fixed a pthread bug where it was only  if (emptyEventQueue(_glfw.mir.eventQueue)) before checking the cond, but the cond might return but not be done. So changed to a while to ensure we only ever move out when we are *actually* done.

Fixed the cursor animation, was completely wrong and now its better!


BrandonSchaefer:
Also note you'll need 16.04 plus this ppa to test https://launchpad.net/~ci-train-ppa-service/+archive/ubuntu/stable-phone-overlay

Or 17.04! Ive tested on my machine that all the text programs + examples were working (that are supported)

elmindreda:
What about 16.10 (which is what I'm running now)?

BrandonSchaefer:
Sadly no, we didnt end up releasing mir 0.26* for 16.10 (only to the ppa for the LTS and 17.04). I can look at getting a ppa setup for you for 16.10 though, unless yakkety has something broken in it!

elmindreda:
If it's not a hassle; otherwise I'll wait until 17.04 is out.

BrandonSchaefer:
It appears to have built:
https://launchpad.net/~brandontschaefer/+archive/ubuntu/mir-testing

Just in case ppa-purge is nice to clear out bad ppas. Not sure if everything in yakkety is *up* to date enough. Though it should be!

elmindreda:
It builds perfectly. Thank you!

PR #983, opened 2017-04-05
Make glfwGetKeyName always return UTF8 key names

### LINKED ISSUE #981: glfwGetKeyName depends on current locale
Steps to reproduce:

- switch the keyboard to some variant of French AZERTY
- run a program that calls `glfwGetKeyName(GLFW_KEY_2, 0)`

This should return "é". On a system with locale fr_FR.UTF-8, this should be "\xC3\xA9" but it returns "\xE9", the same character but in ISO8859-15 encoding (this is a limitation of `XkbTranslateKeySym`).

I would expect `glfwGetKeyName` to return unicode keynames across all platforms, regardless of the user's locale.

This is related to #837.

An easy solution is to use the a public domain keysym dataset compiled by Markus G. Kuhn ([C code](http://www.cl.cam.ac.uk/~mgk25/ucs/keysym2ucs.c)).

db47h:
Fixes #981.

I could not find a non-combining symbol for some dead keys. For these, I put in the corresponding combining symbol, but left the entries commented out. We could either:

- leave them out (they do not seem to be mapped to level 1 on any keyboard layout, so this should not be an issue).
- return the combining symbol as-is. It would be up to the programmer to make sure that they display properly. They are all in the Combining Diacritical Marks Unicode block (0x0300-0x036F).
- precede them with a space in the returned value from glfwGetKeyName().

### DIFF
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -876,6 +876,36 @@ static void releaseMonitor(_GLFWwindow* window)
     }
 }
 
+// Encode a Unicode code point to a UTF-8 stream
+// Based on cutef8 by Jeff Bezanson (Public Domain)
+//
+static size_t encodeUTF8(char *dest, uint32_t ch)
+{
+    if (ch < 0x80) {
+        dest[0] = (char)ch;
+        return 1;
+    }
+    if (ch < 0x800) {
+        dest[0] = (ch>>6) | 0xC0;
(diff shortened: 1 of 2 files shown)


elmindreda:
That looks really promising! I have a holiday coming up, but I'll review and test as soon as I can.

PR #985, opened 2017-04-06
Fix EGL support

kkaefer:
This adds support for using GLFW with EGL backends on macOS. While EGL isn't available by default on macOS, there are third-party implementations like [SwiftShader](https://github.com/google/swiftshader).

EGL is typically used in conjunction with OpenGL ES, so we're now preferring inclusion of these headers when `GLFW_INCLUDE_ES*` is defined by moving them before the `#if __APPLE__` check.

Additionally, the addition of [MoltenVK support](https://github.com/glfw/glfw/commit/e94d16667b1696e90da6c1b5c551815b0f873534) added an override to `makeBackingLayer`. In non-Vulkan mode, this means that we are returning `nil` from that function. The NSGL implementation apparently creates this layer manually later on, but other OpenGL backends don't necessarily do that.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -423,6 +423,8 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (id)makeBackingLayer
 {
+    if (!window->ns.layer)
+        window->ns.layer = [super makeBackingLayer];
     return window->ns.layer;
 }
 
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -139,7 +139,32 @@ extern "C" {
 
 /* Include the chosen client API headers.
  */
-#if defined(__APPLE__)
+#if defined(GLFW_INCLUDE_ES1)
+ #include <GLES/gl.h>
+ #if defined(GLFW_INCLUDE_GLEXT)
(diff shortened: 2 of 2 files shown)


elmindreda:
Thank you, will review.

src/cocoa_window.m
```diff
@@ -423,6 +423,9 @@ - (BOOL)wantsUpdateLayer
 
 - (id)makeBackingLayer
 {
+    if (!window->ns.layer) {
```
elmindreda: GLFW uses the [Allman brace style](https://en.wikipedia.org/wiki/Indent_style#Allman_style) and for single lines leaves out braces altogether.

kkaefer: Removed braces.

elmindreda:
Oops, forgot to update my local pull request branch and didn't merge your updated version. Sorry about that.

This has been merged as d1d08ef99f9b4ea6726df92ef313e320caf32f24 and 244d6d782375f4277123e43c5dc2ede7d41aa4f0 with additional cleanup in 018ae69b33f3bef10b9387d6606b45371735b730 and fa0b5e1b85ce9178f5279f27ef99a9a114077f6d.

PR #987, opened 2017-04-06
Window drag

felipefs:
Add a function ```glfwDragWindow``` that starts a drag operation for the specified window.

The implementation is done for X11 and Windows, and requires tests on Wayland.

### DIFF
--- a/examples/simple.c
+++ b/examples/simple.c
@@ -68,6 +68,12 @@ static void error_callback(int error, const char* description)
     fprintf(stderr, "Error: %s\n", description);
 }
 
+void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
+{
+    if (button == GLFW_MOUSE_BUTTON_LEFT && action == GLFW_PRESS)
+        glfwDragWindow(window);
+}
+
 static void key_callback(GLFWwindow* window, int key, int scancode, int action, int mods)
 {
     if (key == GLFW_KEY_ESCAPE && action == GLFW_PRESS)
@@ -96,6 +102,7 @@ int main(void)
     }
 
     glfwSetKeyCallback(window, key_callback);
+    glfwSetMouseButtonCallback(window, mouse_button_callback);
 
     glfwMakeContextCurrent(window);
     gladLoadGLLoader((GLADloadproc) glfwGetProcAddress);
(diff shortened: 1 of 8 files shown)


felipefs:
The Travis-CI report:
```
...
[ 15%] Linking C shared library libglfw.dylib
Undefined symbols for architecture x86_64:
  "__glfwPlatformDragWindow", referenced from:
      _glfwDragWindow in window.c.o

ld: symbol(s) not found for architecture x86_64
```

Should I make an empty definition of ```_glfwPlatformDragWindow``` on ```cocoa_window.c``` while the implementation for those platforms aren't still done? Example:

```
void _glfwPlatformDragWindow(_GLFWwindow* window)
{
}
```

Or is there another way?

elmindreda:
Yes, always add stubs to all backends when creating a new platform function.

felipefs:
The Wayland implementation is fixed by https://github.com/glfw/glfw/pull/992.

elmindreda:
This is a much nicer interface than what I had in mind. 👍 

I won't be able to review this properly until Thursday at the earliest. In the meantime, [movable​By​Window​Background](https://developer.apple.com/reference/appkit/nswindow/1419072-movablebywindowbackground?language=objc) may be of use on macOS.

felipefs:
> I won't be able to review this properly until Thursday at the earliest.

No worries.

> In the meantime, movable​By​Window​Background may be of use on macOS.

I was thinking about that one (there's even other property called ```mouseDownCanMoveWindow```), but the behaviour is different compared with Linux/Windows. The Linux/Windows lets you do this:
```c
/* Drag window only on the customized top bar */
if (cursor_y < 20) {
    glfwDragWindow(window);
}
/* or even start the drag operation using a key */
if (key == GLFW_KEY_D) {
    glfwDragWindow(window);
}
```
But, ```movable​By​Window​Background``` (and ```mouseDownCanMoveWindow```) are flags (not functions that start an operation) that if the user clicks **anywhere** (and only works with mouse click), the window will still be draggable:
```c
if (cursor_y < 20) {
    glfwDragWindow(window);
}
/* Window still will be draggable outside of that scope, on any cursor position */
```

So, the macOS implementation will definitelly be hacky (if we do it). My suggestion, if you don't mind, is to leave this patch only with the X11/Wayland/Windows implementation, and I could work on other PR with the macOS implementation.

elmindreda:
This has been staged in the `drag-window` branch for testing and completion.

codeyash:
Any updates for such a nice feature

Issue #732, opened 2016-03-29
Window attention request


stelar7:
Is there a way for glfw to request that the window should get focus?
and a way for it to drop focus?
IE. flashing the task bar icon? [(like this)](http://answers.unity3d.com/storage/temp/31886-1.png)

elmindreda:
If you mean like [FlashWindow](https://msdn.microsoft.com/en-us/library/windows/desktop/ms679346.aspx] then no, that's not yet supported.

felipefs:
I'm not sure how this feature works on windows, but I was able to create a function ```glfwRequestWindowAttention()``` for X11. If there's interest on this, I can upload the branch and make a pull request.

felipefs:
I've made the implementation for X11 and Windows (using ```FlashWindow```). I tested the Windows implementation on Linux, using Wine. Some extra test on Windows would be nice to make sure of the behavior. Here's the branch: https://github.com/felipefs/glfw/tree/attention-request

elmindreda:
This has been added with #988.

PR #988, opened 2017-04-06
Support for attention request

felipefs:
This pull contains the implementation of attention request for X11.

The added function was ```glfwRequestWindowAttention```.

The example ```simple.c``` was updated to demonstrate the functionality.

This patch is related with the issue https://github.com/glfw/glfw/issues/732.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1272,6 +1272,10 @@ void _glfwPlatformShowWindow(_GLFWwindow* window)
     [window->ns.object orderFront:nil];
 }
 
+void _glfwPlatformRequestWindowAttention(_GLFWwindow* window)
+{
+}
+
 void _glfwPlatformHideWindow(_GLFWwindow* window)
 {
     [window->ns.object orderOut:nil];
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1272,14 +1272,14 @@ void _glfwPlatformShowWindow(_GLFWwindow* window)
     [window->ns.object orderFront:nil];
 }
 
-void _glfwPlatformRequestWindowAttention(_GLFWwindow* window)
-{
-    [window->ns.object requestUserAttention:NSInformationalRequest];
-}
-
 void _glfwPlatformHideWindow(_GLFWwindow* window)
 {
     [window->ns.object orderOut:nil];
+}
+
+void _glfwPlatformRequestWindowAttention(_GLFWwindow* window)
+{
+    [NSApp requestUserAttention:NSInformationalRequest];
 }
 
 void _glfwPlatformFocusWindow(_GLFWwindow* window)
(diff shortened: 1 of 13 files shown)


src/cocoa_window.m
```diff
@@ -1272,6 +1272,11 @@ void _glfwPlatformShowWindow(_GLFWwindow* window)
     [window->ns.object orderFront:nil];
 }
 
+void _glfwPlatformRequestWindowAttention(_GLFWwindow* window)
+{
+    [window->ns.object requestUserAttention:NSInformationalRequest];
```
elmindreda: `requestUserAttention:` is part of NSApplication, not NSWindow.

elmindreda:
Because some platforms (read macOS) have application-level attention requests, perhaps the more neutral `glfwRequestAttention` would be a better name?

felipefs:
I was wondering how we could approach that without breaking compatibility between platforms, since on Linux/Window, it will look like ```glfwRequestAttention(GLFWwindow *)``` and on macOS ```glfwRequestAttention(void)```.

elmindreda:
Have it take a window and document that it's a hint and not guaranteed to point out the specific window.

felipefs:
I've updated the function to do the following:
```objective-c
void _glfwPlatformRequestWindowAttention(_GLFWwindow* window)
{
    [NSApp requestUserAttention:NSInformationalRequest];
}
```
Also, I updated the documentation, adding a remark that on OSX the attention refers to the application, not the window.

Let me know if something is incorrect and I will fix 👍

elmindreda:
This looks almost ready to merge. Please don't add any test code or additional features to the `simple` example, though. It's mean to be just that and is [used in the documentation](http://www.glfw.org/docs/latest/quick_guide.html#quick_example).

felipefs:
Do you want me to remove the added code from ```simple.c```?

elmindreda:
Yes, please do. Instead you could for example request attention whenever the `joysticks` test detects a connection or disconnection.

src/x11_window.c
```diff
@@ -2078,6 +2078,21 @@ void _glfwPlatformHideWindow(_GLFWwindow* window)
...
+
+    memset(&xev, 0, sizeof(xev));
+    xev.type = ClientMessage;
+    xev.xclient.window = window->x11.handle;
+    xev.xclient.message_type = _glfw.x11.NET_WM_STATE;
+    xev.xclient.format = 32;
+    xev.xclient.data.l[0] = _NET_WM_STATE_ADD;
+    xev.xclient.data.l[1] = _glfw.x11.NET_WM_STATE_DEMANDS_ATTENTION;
+
+    XSendEvent(_glfw.x11.display, DefaultRootWindow(_glfw.x11.display), False, SubstructureRedirectMask | SubstructureNotifyMask, &xev);
```
elmindreda: This function could be made significantly shorter by calling the helper `sendEventToWM`.

elmindreda:
This has been merged as 412eb6a6113f71f42ad56d8dd982003ba7094870, with additional cleanup in baa9cd8968e1571cd10a7265b589d997febf611e and 2f5e2303380d084ab9b097765d34f98504324d65.

Issue #920, opened 2016-12-18
Query for desktop work area


adambrown:
The screen working area is the desktop area of the display, excluding taskbars, docked windows, and docked tool bars. 

In windows you get it with (using System.Windows.Forms):

`Screen::WorkingArea`

https://msdn.microsoft.com/en-us/library/system.windows.forms.screen.workingarea

In java you get it by subtracting screen insets from screen dimensions for each screen since not all screens have these items on them:

` Toolkit.getDefaultToolkit().getScreenInsets(graphicsConfiguration)`

I don't know how Java gets them for OSX but it does somehow, possibly someone who knows something about OSX can chime in.

Either method, getting the working area directly, or getting insets and subtracting from screen space works fine, but right now there is no way in GLFW, that I could find, to get this space.

One reason this is necessary is that right now GLFW does not maximize correctly when maximizing a window with no decorations. Instead of taking up the working area when `glfwMaximizeWindow()` is called it takes up the entire screen (the function works fine for a decorated window). I don't know if that was the intended behavior or not, so I'm not calling it a bug, but the developer needs to be able to decide which behavior they want for undecorated windows as well.

There are other reasons why having the working area is necessary especially for things like floating windows for notifications or other widgets. Generally you want those to not cover the task bar, but you don't want them to be decorated.

adambrown:
This is related to: https://github.com/glfw/glfw/issues/899 as 899 is the behavior that causes the first reason this functionality is needed. However, this functionality is necessary regardless of whether that bug is fixed or not so not closing this.

elmindreda:
Fair enough. I won't be adding this to 3.3 but it may be included in 3.4. A work area query could be useful.

NkD:
I am using GLFW in Java (LWJGL). 
In Java we can get "working area" by AWT: `Toolkit.getDefaultToolkit().getScreenInsets(conf)`. 
But this fault in case if you have more then one monitor. 
This bug is reported before many years and still exists :(
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6899304

In Java doesn't have way how create work-around.
We need to provide working area directly via GLFW.

Please this is urgent for me and maybe for other Java developers.

felipefs:
I believe that on X11, the work are in question can be retrieved with the property ```_NET_WORKAREA```. I can work on this if no one is working on it yet.

elmindreda:
@felipefs Please do! I had in mind an interface something like
```c
[...]

PR #989, opened 2017-04-07
Add support for querying work area

felselva:
This patch has the implementation for querying the work area on X11 and Windows.

- **The X11 implementation**. I believe everything is done with this implementation. I tested with single monitor, but since the monitor is passed directly to the function, I don't think there will be any problem with it.

- **The Windows implementation**. This implementation surely needs test on multiple monitors. I tested with a single monitor, and the work area retrieved was correct. However, we need to make sure that the ```HMONITOR``` that we grab the information from matches the ```_GLFWmonitor``` passed to the function.

- **The Cocoa implementation**. The implementation for Cocoa is a draft and needs to be test and reviewed.

This patch is related to the issue https://github.com/glfw/glfw/issues/920.

### DIFF
--- a/examples/simple.c
+++ b/examples/simple.c
@@ -79,6 +79,7 @@ int main(void)
     GLFWwindow* window;
     GLuint vertex_buffer, vertex_shader, fragment_shader, program;
     GLint mvp_location, vpos_location, vcol_location;
+    int workarea_x, workarea_y, workarea_width, workarea_height;
 
     glfwSetErrorCallback(error_callback);
 
@@ -131,6 +132,8 @@ int main(void)
     glVertexAttribPointer(vcol_location, 3, GL_FLOAT, GL_FALSE,
                           sizeof(vertices[0]), (void*) (sizeof(float) * 2));
 
+    glfwGetMonitorWorkarea(glfwGetPrimaryMonitor(), &workarea_x, &workarea_y, &workarea_width, &workarea_height);
+    printf("Monitor work area: %d, %d, %d, %d\n", workarea_x, workarea_y, workarea_width, workarea_height);
     while (!glfwWindowShouldClose(window))
     {
         float ratio;
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1719,6 +1719,31 @@ GLFWAPI GLFWmonitor* glfwGetPrimaryMonitor(void);
  */
 GLFWAPI void glfwGetMonitorPos(GLFWmonitor* monitor, int* xpos, int* ypos);
 
+/*! @brief Returns the work area of the monitor.
+ *
+ *  This function returns the position, in screen coordinates, of the upper-left
+ *  corner of the specified monitor.
+ *
+ *  Any or all of the position arguments may be `NULL`.  If an error occurs, all
+ *  non-`NULL` position arguments will be set to zero.
+ *
+ *  @param[in] monitor The monitor to query.
+ *  @param[out] xpos Where to store the monitor x-coordinate, or `NULL`.
+ *  @param[out] ypos Where to store the monitor y-coordinate, or `NULL`.
+ *
+ *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED and @ref
+ *  GLFW_PLATFORM_ERROR.
+ *
+ *  @thread_safety This function must only be called from the main thread.
+ *
+ *  @sa @ref monitor_properties
+ *
+ *  @since Added in version 3.0.
+ *
+ *  @ingroup monitor
+ */
+GLFWAPI void glfwGetMonitorWorkarea(GLFWmonitor* monitor, int* xpos, int* ypos, int *width, int *height);
+
 /*! @brief Returns the physical size of the monitor.
  *
  *  This function returns the size, in millimetres, of the display area of the
--- a/src/monitor.c
+++ b/src/monitor.c
@@ -296,6 +296,25 @@ GLFWAPI void glfwGetMonitorPos(GLFWmonitor* handle, int* xpos, int* ypos)
     _glfwPlatformGetMonitorPos(monitor, xpos, ypos);
 }
 
+GLFWAPI void glfwGetMonitorWorkarea(GLFWmonitor* handle, int* xpos, int* ypos, int* width, int* height)
+{
+    _GLFWmonitor* monitor = (_GLFWmonitor*) handle;
+    assert(monitor != NULL);
+
+    if (xpos)
+        *xpos = 0;
+    if (ypos)
+        *ypos = 0;
+    if (width)
+        *width = 0;
+    if (width)
+        *width = 0;
+
+    _GLFW_REQUIRE_INIT();
+
+    _glfwPlatformGetMonitorWorkarea(monitor, xpos, ypos, width, height);
+}
+
 GLFWAPI void glfwGetMonitorPhysicalSize(GLFWmonitor* handle, int* widthMM, int* heightMM)
 {
     _GLFWmonitor* monitor = (_GLFWmonitor*) handle;
--- a/src/win32_monitor.c
+++ b/src/win32_monitor.c
@@ -291,6 +291,36 @@ void _glfwPlatformGetMonitorPos(_GLFWmonitor* monitor, int* xpos, int* ypos)
         *ypos = settings.dmPosition.y;
 }
 
+static BOOL CALLBACK MonitorEnumProc(HMONITOR hMonitor, HDC hdcMonitor, LPRECT lprcMonitor, LPARAM dwData)
+{
+    RECT *workarea = (RECT *)dwData;
+    MONITORINFO monitorInfo;
+    GetMonitorInfo(hMonitor, &monitorInfo);
+    workarea->left = monitorInfo.rcWork.left;
+    workarea->top = monitorInfo.rcWork.top;
+    workarea->right = monitorInfo.rcWork.right;
+    workarea->bottom = monitorInfo.rcWork.bottom;
+    return TRUE;
+}
+
+void _glfwPlatformGetMonitorWorkarea(_GLFWmonitor* monitor, int* xpos, int* ypos, int *width, int *height)
+{
+    HDC dc;
+    RECT workarea;
+
+    dc = CreateDCW(L"DISPLAY", monitor->win32.adapterName, NULL, NULL);
+
+    if (!EnumDisplayMonitors(dc, NULL, MonitorEnumProc, (LPARAM)&workarea))
+        return;
+
+    DeleteDC(dc);
+
+    *xpos = workarea.left;
+    *ypos = workarea.top;
+    *width = workarea.right;
+    *height = workarea.bottom;
+}
+
 GLFWvidmode* _glfwPlatformGetVideoModes(_GLFWmonitor* monitor, int* count)
 {
     int modeIndex = 0, size = 0;
--- a/src/win32_monitor.c
+++ b/src/win32_monitor.c
@@ -295,6 +295,7 @@ static BOOL CALLBACK MonitorEnumProc(HMONITOR hMonitor, HDC hdcMonitor, LPRECT l
 {
     RECT *workarea = (RECT *)dwData;
     MONITORINFO monitorInfo;
+    monitorInfo.cbSize = sizeof(MONITORINFO);
     GetMonitorInfo(hMonitor, &monitorInfo);
     workarea->left = monitorInfo.rcWork.left;
     workarea->top = monitorInfo.rcWork.top;
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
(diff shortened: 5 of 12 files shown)


elmindreda:
I won't be able to review this properly until Thursday at the earliest but it looks like a solid start. `_GLFWmonitorWin32` will be getting its corresponding `HMONITOR` for high DPI support, which should simplify this patch.

src/win32_monitor.c
```diff
@@ -291,6 +291,37 @@ void _glfwPlatformGetMonitorPos(_GLFWmonitor* monitor, int* xpos, int* ypos)
...
+}
+
+void _glfwPlatformGetMonitorWorkarea(_GLFWmonitor* monitor, int* xpos, int* ypos, int *width, int *height)
+{
+    HDC dc;
+    RECT workarea;
+
+    dc = CreateDCW(L"DISPLAY", monitor->win32.adapterName, NULL, NULL);
+
+    if (!EnumDisplayMonitors(dc, NULL, MonitorEnumProc, (LPARAM)&workarea))
```
elmindreda: Enumerating based on adapter DC doesn't work. Tested on Windows 7.

elmindreda: Tried every which way I could think of. Adapter device DCs yield no monitors and display device DCs yield all available monitors. Tested on dual-monitor system running Windows 7.

felipefs: For some reason, using the DC worked here. We can put this patch on halt until we have the ```high DPI``` patch  with ```HMONITOR``` on each ```_GLFWmonitorWin32```, then. And we won't have two implementations of callbacks for ```EnumDisplayMonitors```, making things cleaner.

elmindreda: Wine is amazing but it's not a very reliable oracle for Windows.

jadahl:
What is the purpose of this API? Most likely, it wont be possible to implement completely on Wayland as there is usually no valid reason for querying this type of information (since the client have no possible use for it). What could be useful though is being aware of work area *size*, as this can be used to determine a good default size for a new window.

felipefs:
@jadahl The purpose is exactly to retrieve the size of the work area. I started implementing due this issue https://github.com/glfw/glfw/issues/920.

I mentioned there that Wayland in particular doesn't have the concept of work area. In my opinion, we should at least return the resolution of the screen, even if it is redundant with ```GLFWvidmode->width```/```GLFWvidmode->height```, because it would result in an approximate result across platforms. Also add a remark on the documentation about that.

jadahl:
The concept of work area doesn't exist yet, true, but its introduction has been requested before, by for example GTK+ that uses the work area to figure out a default window size. The position, however, will probably not be exposed, as it would serve no purpose.

What has to be considered is that it might not be useful to know any work area before knowing what monitor the window will be placed at. I expect this to be more of a compositor driven thing, as it has proven to be practically impossible to let applications them self place them in any predictable manner in a multi monitor configuration. Fullscreen is of course a different matter here, and in that case work area is irrelevant, as is with maximizing, as that is also done in a completely different way.

So, I'm still wondering, what is the purpose of this API? What is its intended use case? Is it possible make any use of the API if the client doesn't have total control of exact positioning details?

elmindreda:
Ping @adambrown.

felipefs:
> What has to be considered is that it might not be useful to know any work area before knowing what monitor the window will be placed at. I expect this to be more of a compositor driven thing, as it has proven to be practically impossible to let applications them self place them in any predictable manner in a multi monitor configuration. Fullscreen is of course a different matter here, and in that case work area is irrelevant, as is with maximizing, as that is also done in a completely different way.

If the window is created with ```GLFW_VISIBLE``` as ```FALSE```, it will be placed in a monitor before showing up, then it is possible to get the monitor using ```glfwSetWindowMonitor```. It is useful for the same purpose in GTK+ and not extrapolate the size of the window beyond the work area.

adambrown:
@jadahl Sorry I've been on vacation for 3 weeks and haven't been online. The purpose of this feature is detailed here in the ticket I created: #920 

In the first usecase described there, trying to maximize an undecorated window, I suppose it wouldn't be strictly necessary to know which display it would end up on the first time. Although right now in my app I am determining the display for the window based on the current location of the mouse at init on first run unless there is a saved positioning configuration matching the current display setup. Or in other words, I check the count, size and relative positioning of the displays to see if I have a saved window position and size for that layout. If I don't have a stored position and size for the matching display layout then I put the window at a default small size in the center of the display the mouse is on. Most of this work I have to do with Java AWT right now because GLFW doesn't allow me to have all the data I need, such as the mouse location, without creating a window. The reason for doing it this way is that if you have a laptop that you take from home to work, or you install the app as portable on a USB stick, and there are different monitor setups in different locations, you can have positioning information stored for all of them.

For the other usecase, trying to position notifications or other widgets, you would need to know what display you are putting the things on.

I hope that clarifies the exact purpose and intended usecase of this API.

include/GLFW/glfw3.h
```diff
@@ -1719,6 +1719,31 @@ GLFWAPI GLFWmonitor* glfwGetPrimaryMonitor(void);
...
+ *
+ *  This function returns the position, in screen coordinates, of the upper-left
+ *  corner of the specified monitor.
+ *
+ *  Any or all of the position arguments may be `NULL`.  If an error occurs, all
+ *  non-`NULL` position arguments will be set to zero.
+ *
+ *  @param[in] monitor The monitor to query.
+ *  @param[out] xpos Where to store the monitor x-coordinate, or `NULL`.
+ *  @param[out] ypos Where to store the monitor y-coordinate, or `NULL`.
```
linkmauve: You’re missing `width` and `height` here.

include/GLFW/glfw3.h
```diff
@@ -1719,6 +1719,31 @@ GLFWAPI GLFWmonitor* glfwGetPrimaryMonitor(void);
  */
 GLFWAPI void glfwGetMonitorPos(GLFWmonitor* monitor, int* xpos, int* ypos);
 
+/*! @brief Returns the work area of the monitor.
+ *
+ *  This function returns the position, in screen coordinates, of the upper-left
+ *  corner of the specified monitor.
+ *
+ *  Any or all of the position arguments may be `NULL`.  If an error occurs, all
```
linkmauve: The length arguments can also be NULL, according to the implementation.

src/monitor.c
```diff
@@ -296,6 +296,25 @@ GLFWAPI void glfwGetMonitorPos(GLFWmonitor* handle, int* xpos, int* ypos)
...
+    assert(monitor != NULL);
+
+    if (xpos)
+        *xpos = 0;
+    if (ypos)
+        *ypos = 0;
+    if (width)
+        *width = 0;
+    if (width)
+        *width = 0;
```
linkmauve: Itym height, same on the line above.

include/GLFW/glfw3.h
```diff
@@ -1719,6 +1719,31 @@ GLFWAPI GLFWmonitor* glfwGetPrimaryMonitor(void);
...
+ *  @param[out] ypos Where to store the monitor y-coordinate, or `NULL`.
+ *
+ *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED and @ref
+ *  GLFW_PLATFORM_ERROR.
+ *
+ *  @thread_safety This function must only be called from the main thread.
+ *
+ *  @sa @ref monitor_properties
+ *
+ *  @since Added in version 3.0.
```
linkmauve: Heh, version will be at best 3.3. ^^

include/GLFW/glfw3.h
```diff
@@ -1719,6 +1719,31 @@ GLFWAPI GLFWmonitor* glfwGetPrimaryMonitor(void);
  */
 GLFWAPI void glfwGetMonitorPos(GLFWmonitor* monitor, int* xpos, int* ypos);
 
+/*! @brief Returns the work area of the monitor.
+ *
+ *  This function returns the position, in screen coordinates, of the upper-left
+ *  corner of the specified monitor.
```
linkmauve: Also size.

examples/simple.c
```diff
@@ -131,6 +132,8 @@ int main(void)
     glVertexAttribPointer(vcol_location, 3, GL_FLOAT, GL_FALSE,
                           sizeof(vertices[0]), (void*) (sizeof(float) * 2));
 
+    glfwGetMonitorWorkarea(glfwGetPrimaryMonitor(), &workarea_x, &workarea_y, &workarea_width, &workarea_height);
+    printf("Monitor work area: %d, %d, %d, %d\n", workarea_x, workarea_y, workarea_width, workarea_height);
```
linkmauve: This would probably be better in another example than the simple one, which is meant to be simple.

jadahl:
Not only is no Wayland implementation possible at the moment, the usefulness of it would there be one is limited, as if you try to manually "maximize" (just set a size equal to the work area), the window would still not be maximized, nor moved to the top-left corner.

To me, to make maximize work for undecorated windows, the only sane way would be to make glfwMaximizeWindow() do the right thing. Anything that tries to work around that function not being able to work as expected by bypassing the windowing system abstraction GLFW is trying to provide undoes exactly what GLFW is intended to do.

ocornut:
Hello,

Since the usefulness of this is being questioned above, may I humbly add that I also would _very much_ like to have this API.  My framework (same project as shown in #1236) creates pop-up windows that I position precisely relative to each others, with windows protrude out of the parent window. I currently use monitor information from GLFW to avoid creating popups that straddle multiple monitors but also want to avoid "task bars", this is one of the reason OS export the "work area" information. I can do this with Win32 and SDL2 (`SDL_GetDisplayUsableBounds` if that's any useful) but GLFW doesn't export this information at the moment.

For my use case this PR would perfectly solve the problem, and it is fine to have a fallback that return the entire monitor region.

Regards,
Omar

dougbinks:
I concur with Omar that this is very useful outside of the now fixed ```glfwMaximizeWindow `` issue I raised, mainly for multi-window apps and those which wish to auto-layout multiple processes with non overlapping windows (my own use case).

For any platform without this concept the function can return coordinates from ```glfwGetMonitorPos`` and ```glfwGetVideoMode```.

If the @ferreiradaselva does not have the time I can look into fixing any issues with the PR if the basic functionality requirements are approved.

jadahl:
The fact that it cannot be made portable still remains however. What GLFW probably needs is a versatile popup placement API like the one GTK+ added.

ocornut:
> The fact that it cannot be made portable still remains however. What GLFW probably needs is a versatile popup placement API like the one GTK+ added.

For various reasons I would probably be unable to rely on such specific "popup placement" feature. I want the most low-level feature that has a comparable implementation on all major frameworks. The concept of a work area rectangle per monitor is pretty well supported across frameworks and OS and has a sane fall back any other implementation can use (return the whole screen).

dougbinks:
> The fact that it cannot be made portable still remains however. 

The fallback is portable, and makes sense where the platform does not have such a concept.

I don't think popups work for the multi-process window distribution problem, whereas the ```glfwGetMonitorWorkarea``` does.

dougbinks:
I have started work on resolving issues raised with this PR on my branch https://github.com/dougbinks/glfw/tree/pr-989.

Either this can be merged into this PR branch by @ferreiradaselva when I'm done or I can submit a new PR.

src/win32_monitor.c
dougbinks: I'm working through these issues and have a resolution for this by using the monitor virtual position to derive a handle then use that to get the work area: https://github.com/dougbinks/glfw/commit/921d9833b262b6476ddeaa86af9f2f067d0bc434

include/GLFW/glfw3.h
dougbinks: Fixed this here: https://github.com/dougbinks/glfw/commit/147b0f5e839f3c6367609320592330d6a54ec768

include/GLFW/glfw3.h
dougbinks: Fixed this here: https://github.com/dougbinks/glfw/commit/875cbdf46d12e4d23d41e80069cd921f587e5829

src/monitor.c
dougbinks: Fixed height here: https://github.com/dougbinks/glfw/commit/196a7cf7d2bafa42e1f1f53cb70a98bbf7095f29

include/GLFW/glfw3.h
dougbinks: Updated to 3.3: https://github.com/dougbinks/glfw/commit/992e5da96a8f4fb99ef92c6b82ad2be2e5e52bdf

include/GLFW/glfw3.h
dougbinks: Added size and work area definition: https://github.com/dougbinks/glfw/commit/1a1ad508abb2d8b626806dc3edc318214e0ff009

examples/simple.c
dougbinks: Moved the example/test to monitors.c: https://github.com/dougbinks/glfw/commit/a484f0d8ddba3c4a1f7aaf12858812a5c0f64607

elmindreda:
This has been merged via #1322 and additional fixes.

PR #1006, opened 2017-05-06
cmake_minimum_required() command should at the beginning file

KindDragon:


elmindreda:
Oh, they fixed it! I tried having the commands in this order on earlier versions of CMake but then `CMAKE_LEGACY_CYGWIN_WIN32` had no effect.

Thank you!

PR #1012, opened 2017-05-12
Get event time

felipefs:
This patch add support for getting the event time of key-press, key-release, button-press, button-release and cursor motion. This is useful especially for cases of frame rate variation, like frame drops or disabled v-sync.

The test ```event.c``` was updated to print the event time, showing the event time between parenthesis, as ```printf("...(event time: %0.3f)...")```.

I tested the Wayland, X11 and Windows implementations. The Mac/Cocoa and Mir implementations are untested.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -3224,6 +3224,28 @@ GLFWAPI GLFWwindowmaximizefun glfwSetWindowMaximizeCallback(GLFWwindow* window, 
  */
 GLFWAPI GLFWframebuffersizefun glfwSetFramebufferSizeCallback(GLFWwindow* window, GLFWframebuffersizefun cbfun);
 
+/*! @brief Returns the time of the last input event.
+ *
+ *  This function returns the time, in seconds, of the last event occurence. The
+ *  only events queried are the input events button-press, button-release,
+ *  key-press, key-release and cursor motion, and the proper place to call this
+ *  function is in one of the input callbacks.
+ *
+ *  @param[in] window The window to check for the input event.
+ *  @return The value, in seconds, of the last input event occurence.
+ *
+ *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED.
+ *
+ *  @thread_safety This function must only be called from the main thread.
+ *
+ *  @sa @ref events
+ *
+ *  @since Added in version 3.3.
+ *
+ *  @ingroup window
+ */
+GLFWAPI double glfwGetEventTime(GLFWwindow* window);
+
 /*! @brief Processes all pending events.
  *
  *  This function processes only those events that are already in the event
--- a/src/wl_platform.h
+++ b/src/wl_platform.h
@@ -100,6 +100,8 @@ typedef struct _GLFWwindowWayland
         struct zwp_relative_pointer_v1*    relativePointer;
         struct zwp_locked_pointer_v1*      lockedPointer;
     } pointerLock;
+
+    unsigned int lastEventTime;
 } _GLFWwindowWayland;
 
 // Wayland-specific global data
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -433,6 +433,8 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)mouseDown:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputMouseClick(window,
                          GLFW_MOUSE_BUTTON_LEFT,
                          GLFW_PRESS,
@@ -441,11 +443,15 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)mouseDragged:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     [self mouseMoved:event];
 }
 
 - (void)mouseUp:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputMouseClick(window,
                          GLFW_MOUSE_BUTTON_LEFT,
                          GLFW_RELEASE,
@@ -454,6 +460,8 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)mouseMoved:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     if (window->cursorMode == GLFW_CURSOR_DISABLED)
     {
         const double dx = [event deltaX] - window->ns.cursorWarpDeltaX;
@@ -477,6 +485,8 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)rightMouseDown:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputMouseClick(window,
                          GLFW_MOUSE_BUTTON_RIGHT,
                          GLFW_PRESS,
@@ -485,11 +495,15 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)rightMouseDragged:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     [self mouseMoved:event];
 }
 
 - (void)rightMouseUp:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputMouseClick(window,
                          GLFW_MOUSE_BUTTON_RIGHT,
                          GLFW_RELEASE,
@@ -498,6 +512,8 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)otherMouseDown:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputMouseClick(window,
                          (int) [event buttonNumber],
                          GLFW_PRESS,
@@ -506,11 +522,15 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)otherMouseDragged:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     [self mouseMoved:event];
 }
 
 - (void)otherMouseUp:(NSEvent *)event
 {
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputMouseClick(window,
                          (int) [event buttonNumber],
                          GLFW_RELEASE,
@@ -569,6 +589,8 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
     const int key = translateKey([event keyCode]);
     const int mods = translateFlags([event modifierFlags]);
 
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputKey(window, key, [event keyCode], GLFW_PRESS, mods);
 
     [self interpretKeyEvents:[NSArray arrayWithObject:event]];
@@ -600,6 +622,9 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 {
     const int key = translateKey([event keyCode]);
     const int mods = translateFlags([event modifierFlags]);
+
+    window->ns.lastEventTime = [event timestamp];
+
     _glfwInputKey(window, key, [event keyCode], GLFW_RELEASE, mods);
 }
 
@@ -1431,6 +1456,12 @@ void _glfwPlatformSetWindowFloating(_GLFWwindow* window, GLFWbool enabled)
         [window->ns.object setLevel:NSFloatingWindowLevel];
     else
         [window->ns.object setLevel:NSNormalWindowLevel];
+}
+
+double _glfwPlatformGetEventTime(_GLFWwindow* window)
+{
+    /* Windows events are stored in seconds */
+    return window->ns.lastEventTime;
 }
 
 void _glfwPlatformPollEvents(void)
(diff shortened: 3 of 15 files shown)


elmindreda:
How do you handle wrapping?

elmindreda:
Since the function is meant to be called from within callbacks and is limited to the main thread, there is no need for the window parameter. The time could be stored in the global struct.

elmindreda:
Style-wise it is excellent.

felipefs:
> How do you handle wrapping?

I didn't understand what you mean.

> Since the function is meant to be called from within callbacks and is limited to the main thread, there is no need for the window parameter. The time could be stored in the global struct.

Fixed that. Now it doesn't take the window as argument and the event time is stored in the global struct.

shurcooL:
I have a question. When I first read the PR description, and the use cases, I expected the implementation to bundle the event time with all events. So that if a key is pressed at time t0 and a mouse button is pressed at t1, you can tell exactly when those events happened.

However, looking at the implementation, I see it stores a single global time of the "latest" event.

Is it okay that each latest event overwrites the time of previous events? How does everything fit together?

felipefs:
> Is it okay that each latest event overwrites the time of previous events? How does everything fit together?
As far as I can tell, it's ok., because internall, GLFW will immediately call the respective callback of the event, and they will be processed in order of their arrival:

1. Receive the event
2. Store the event time
3. Do necessary work with the event properties
4. Call the callback
5. Receive next event...

Between 1 and 2, theoretically, shouldn't have any event, since they are collected in order from the event queue (part of the system in the case of Windows and MacOS, or the window sytem in the case of X11, Wayland or Mir).

shurcooL:
Thanks for clarifying, that makes sense if that's how it works.

elmindreda:
@ferreiradaselva I mean that milliseconds stored in a uint32 will overflow every 49 days or so, counted from system boot and not the start of the process. That applies to the Windows, X11 and Wayland implementations of this function. The Windows value is signed as well and will wrap to -2³¹. The code should instead probably accumulate the difference between timestamps into a uint64.

ferreiradaselva:
Got it. I will work on a fix for that.

elmindreda:
Won't the user need to also know the current time, according to whatever time source the message queue uses, for this feature to be fully useful? If so, what do we return on X11?

ghuser404:
@ferreiradaselva, @elmindreda, I would argue that returning time in `double` format is not the best/practical/useful solution. If this was retuning time in nanoseconds, it would be more useful, because you would be able to reliably compare those values.

ferreiradaselva:
@elmindreda As far as I know, the point is being able to compare the time that the event was pushed in the queue with the time returned by `glfwGetTime()`

@ghuser404 `glfwGetTime()` returns `double` that's why this function is returning `double`, too.

ghuser404:
@ferreiradaselva, makes sense. Although I think `glfwGetTime()` would be useful as a reference time if we could get current time of event handling clock, and set it via `glfwSetTime()`.

Too bad it's all using imprecise floating point values.

tombsar:
@ghuser404 I completely understand your instinct to use integral numbers to represent time, but the fact is that your program has to be running for a *really* long time before floating-point precision starts to cause any problems.

ghuser404:
@tombsar, see http://floating-point-gui.de/errors/comparison/
```
float a = 0.15 + 0.15
float b = 0.1 + 0.2
if(a == b) // can be false!
if(a >= b) // can also be false!
```
This is enough for me to not use floating point values for such problems.

tombsar:
@ghuser404 We are well aware of the limitations of floating-point representations. I have yet to see a single real-world bug caused by storing time as a `double`, and I can't think of any that could not be fixed in the user's code (e.g. by not comparing for exact equality with a constant that is known to not have an exact representation). If you concerned, you can always use `uint64_t millis = glfwGetTime() * 1000.0` or similar as your basis for time comparison.

elmindreda:
@ferreiradaselva I wish we could use `glfwGetTime` as a reference, but it's on at least some platforms backed by a different time source and it uses an offset that the user can change at any time with `glfwSetTime`.

I've looked into making this line up, but there's still the problem of reference time on X11 where we may not even be running on the same machine as the event time source.

include/GLFW/glfw3.h
```diff
@@ -3224,6 +3224,27 @@ GLFWAPI GLFWwindowmaximizefun glfwSetWindowMaximizeCallback(GLFWwindow* window,
  */
 GLFWAPI GLFWframebuffersizefun glfwSetFramebufferSizeCallback(GLFWwindow* window, GLFWframebuffersizefun cbfun);
 
+/*! @brief Returns the time of the last input event.
+ *
+ *  This function returns the time, in seconds, of the last event occurence. The
+ *  only events queried are the input events button-press, button-release,
+ *  key-press, key-release and cursor motion, and the proper place to call this
```
linkmauve: You should mention scroll as well.

src/wl_platform.h
```diff
@@ -141,6 +142,9 @@ typedef struct _GLFWlibraryWayland
     _GLFWwindow*                pointerFocus;
     _GLFWwindow*                keyboardFocus;
 
+    // The time of the last event
+    unsigned int                lastEventTime;
```
linkmauve: `uint32_t` please.

OlivierSohn:
Hello,

What I dislike about this design is that a user of the API has to know wether the event is timed or not.

In #1186 I propose an alternative approach where we pass the time in the callback parameters, so that there is no ambiguity.

OlivierSohn:
Also, regarding the time format, I think `double` is appropriate here both because the API uses it in `glfwGetTime()`, as @ferreiradaselva already mentionned, and also because on osx, the time of events is given using a double representing seconds, hence we could lose precision when converting back to nanoseconds. BTW it would be nice to quantify this, in the worst case where the time is "big", to see what nanoseconds precision is guaranteed.

OlivierSohn:
@elmindreda , you mentionned the consistency with current time, I would also raise this concern : on every platform, the clock used for event times should be consistent with how the time argument of _glfwPlatformWaitEventsTimeout is interpreted.

elmindreda:
Instead of introducing an additional time source as a reference for the event time and since no one seems to know how to get the current time from the X server, what about changing this to `glfwGetEventLatency`? On X11 we could return the less nonsensical 0.0 and since event times are measured in ms on several platforms, the time between the few lines of GLFW event handling code and the callback being called should be negligible.

OlivierSohn:
@elmindreda I don't understant your last comment, maybe it's because I lack some context. Do you mean just changing the name `glfwGetEventTime` to `glfwGetEventLatency ` ?

ghuser404:
I assume it means getting time difference between two events, as opposed to getting time in reference to some clock.

elmindreda:
Yes, exactly what @ghuser404 said.

OlivierSohn:
So I assume it's the time difference between the `n` and `n-1` events? I don't see a use case where this information would make sense to use, especially as here we have a global state for all events : if when reading a key press I know that the event before was `x` ms away, what information does it bring me? I don't even know if it was another key press, or a mouse movement, or else. Did I understand correctly?

ghuser404:
> Sorry, misread. What I meant was return the difference between the event time for that OS event and the then current time from its reference time source (usually epoch or boot time).

@elmindreda Don't you think it might end up negative on some OS this way?

OlivierSohn:
@elmindreda I think I see your idea. But : if event time and time are based on trully different clocks, the information on latency wouldn't help, I think, because it's like substracting apples from oranges. And if they are based on the same clock, then we might as well return the time like it is today, because it's less convoluted. Maybe one thing to do is to verify, platform per platform, if the clocks used are the same.

elmindreda:
@ghuser404 Potentially, if the time source can go backwards, but `glfwGetEventTime` would have the same problem.

@OlivierSohn Why wouldn't it help? What you want to know is the time from when the event was generated and the time it's received. The resulting duration in seconds is the same.

OlivierSohn:
@elmindreda There are different kinds of clocks : realtime, monotonic, cpu. They all have different semantics (see [here](http://chrisdone.com/posts/measuring-duration-in-haskell) and [here](http://hackage.haskell.org/package/clock-0.4.1.3/docs/System-Clock.html)).

> What you want to know is the time from when the event was generated and the time it's received.

In this example, `when the event was generated` is given by the os, using clock 1 and `the time it's received` is given by clock 2 (the clock used underneath glfwGetTime) which may or may not be the same as clock 1.

Let's say that clock 1 is a RealTime clock and clock 2 is a Monotonic clock ([these notions are defined here](http://hackage.haskell.org/package/clock-0.4.1.3/docs/System-Clock.html)), doing the difference between the two would be nonsense because they don't have the same reference (zero) and even, they don't increase at the same speed.

But maybe on some platforms, the 2 clocks are the same : we could test it empirically by writing the event time and the result of glfwGetTime in the event callback and seeing if we have close values all the time. Or, better, inspect documentations and implementations to see what underlying clock is used.

elmindreda:
> Or, better, inspect documentations and implementations to see what underlying clock is used.

I did that and brought up time source specific concerns above in this thread.

For example, the Windows ~event~ message time is based on `GetTickCount`, uint32 ms since boot that wraps every ~49 days. The GLFW timer (`glfwGetTimer*`) that `glfwGetTime` is a helper function on top of uses `QueryPerformance*`, a deliberately unspecified hardware time source that is guaranteed monotonic only on the same core.

> But maybe on some platforms, the 2 clocks are the same : we could test it empirically by writing the event time and the result of glfwGetTime in the event callback and seeing if we have close values all the time.

The result of `glfwGetTime` won't line up with any of the time sources discussed as it starts at `glfwInit` and can have its base changed at any time (from any thread) with `glfwSetTime`.

We could rebase the OS event times to the GLFW timer and return uint64 values of a frequency only known at run-time, no problem, but it doesn't seem very consistent or user friendly.

OlivierSohn:
In addition to my latest comment, I add this quote from the GLFW doc for `glfwGetTime` : 

> It uses the highest-resolution monotonic time source on each supported platform

so we just need to make sure that the os times events using the same clock.

OlivierSohn:
> We could rebase the OS event times to the GLFW timer and return uint64 values of a frequency only known at run-time, no problem, but it doesn't seem very consistent or user friendly.

Or remove the glfwSetTime functionality, and remove the rebase on zero at glfwInit?

elmindreda:
> Or remove the glfwSetTime functionality, and remove the rebase on zero at glfwInit?

We could do that, but not before 4.0. The advantage of `glfwGetEventLatency` is we can put it into 3.3.

OlivierSohn:
So I understand that the frequency is just about converting from the unit (nanosecond or milliseconds) to seconds, so we should do that if we return a double, in all cases.

Hence, your proposal to take the offset and frequency into account would be correct if we knew the 2 clocks are the same (the monotonic one from glfwGetTime, and the one of the os that times the events).

OlivierSohn:
I'm also having second thoughts on removing the rebase to 0 : if we do that the floating point operations precisions precision will decrease so when removing the rebase we could consider changing from double to uint64_t, or not change the behaviour.

OlivierSohn:
@elmindreda Concerning Windows, to followup on what you said, regarding [GetTickCount](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724408(v=vs.85).aspx) and [QueryPerformaceCounter](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644904(v=vs.85).aspx), it seems indeed that these 2 are monotonic, but they have different resolutions. And maybe different starts? GetTickCount starts at 0 at system boot, but it's not specified when is the zero of QueryPerformaceCounter, hence how would we compare them if we wanted to implement `glfwGetEventLatency ` ?

elmindreda:
@OlivierSohn `glfwGetEventLatency` would have nothing to do with `QueryPerformance*`. It would return `GetTickCount() - GetMessageTime()` as calculated when the event was received by GLFW, just before calling the corresponding event callback. Since it returns a duration and not an absolute time it doesn't matter what time source it used. It's just an estimate of how long it took from the time the OS recorded the event to when GLFW received it.

OlivierSohn:
@elmindreda thanks for clarifying, it makes sense now! So if for each platform we know how to get the current time with the same clock as the one the os used to time the event, it will be possible to implement `glfwGetEventLatency` indeed.

ghuser404:
> So if for each platform we know how to get the current time with the same clock as the one the os used to time the event, it will be possible to implement glfwGetEventLatency indeed.

@OlivierSohn I think this is exactly what @elmindreda wanted to make with the initial API.
> I wish we could use glfwGetTime as a reference, but it's on at least some platforms backed by a different time source and it uses an offset that the user can change at any time with glfwSetTime.
> 
> I've looked into making this line up, but there's still the problem of reference time on X11 where we may not even be running on the same machine as the event time source.

If `glfwGetEventLatency` is obliged to use the same clock, then I don't see how it's different from the initially proposed solution.

OlivierSohn:
@ghuser404 I think there is a slight misunderstanding on your side : I underlined that in `glfwGetEventLatency` we need the same clock _as the one the os uses to time events_, **not** the same clock _as the one used by glfwGetTime_. I hope this clarifies my point, which is btw the point @elmindreda made in the beginning of the discussion:

> Won't the user need to also know the current time, according to whatever time source the message queue uses, for this feature to be fully useful?

ghuser404:
@OlivierSohn I probably quoted not the most helpful piece of discussion. But your quote only emphasises what I wanted to say: this whole thing didn't get merged into GLFW 3.3 because we cannot query clock on which the OS delivers events. @elmindreda then suggested to solve this by introducing `glfwGetEventLatency` which wouldn't rely on querying this clock. But you reverted the discussion back to being able to query it :)

Why I quoted the `glfwGetTime` piece is because the problem with integrating this API is because it may be running on a different clock, either because it was initially different or because user set the offset.

OlivierSohn:
Well, I'm not sure if I reverted anything, re-reading the discussions, I tend to think @elmindreda had this idea in the beginning, when she suggested `glfwGetEventLatency`, and it took me a while to see that we were on the same page :)

mdsitton:
User side timestamps are imo are not a solution here. This would also not be making the exposed glfw api multithreaded but the backend platform implementations instead.

PR #1015, opened 2017-05-16
Fix travis builds

Squareys:
Hi @elmindreda !

I have no idea how this happened, but travis changed something, which resulted in the cmake version of the default image being 2.8.7 rather than 2.8.12, which is insufficient for building glfw.
This affected us with magnum aswell (see https://github.com/mosra/magnum/pull/203), and since we build glfw as a dependency on travis aswell, I thought I'd contribute our fixes here, too.

Cheers, Jonathan.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,17 @@ os:
     - linux
     - osx
 sudo: false
+dist: trusty
 addons:
     apt:
         sources:
             - kubuntu-backports
         packages:
             - cmake
+            - libxrandr-dev
+            - libxinerama-dev
+            - libxcursor-dev
+            - libxi-dev
 env:
     - BUILD_SHARED_LIBS=ON
     - BUILD_SHARED_LIBS=OFF


Squareys:
The build doesn't actually run fully through, it seems to be missing a header now:
```
[  1%] Building C object src/CMakeFiles/glfw.dir/context.c.o

In file included from /home/travis/build/glfw/glfw/src/context.c:28:

In file included from /home/travis/build/glfw/glfw/src/internal.h:183:

/home/travis/build/glfw/glfw/src/x11_platform.h:51:10: fatal error: 'X11/extensions/XInput2.h' file not found

#include <X11/extensions/XInput2.h>
```
(We're probably building the 3.2 release.)
https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty Will help with checking whether you are able to install certain packages on travis, in case this missing header is due to a missing package. (CMake configure should probably catch that, though?)

Squareys:
The missing package is `libxi-dev`, which is whitelisted! Will ammend that to my commit.

Squareys:
(It looks like the `libx11-xcb1` package may be superfluous, the magnum build ran through without, I will check that real quick, and then this PR should be safe to merge)

elmindreda:
The list of required header packages keeps changing as new APIs are used and old ones are made fully dynamically loaded, which is why I usually recommend installing the `xorg-dev` metapackage. I'm sure the Travis folks will thank you for being more specific, though.

Squareys:
The xorg-dev package is also whitelisted, shall I change it to that? :)

elmindreda:
No, let's be nice to the free build bots. The latest revision seems to work well.

elmindreda:
Thanks!

Squareys:
No problem, thanks for merging! 👍

PR #1021, opened 2017-05-21
Radial deadzone on sticks for better precision

suprstefan:
When testing the 360 gamepad input on windows I noticed a big spike in stick input when leaving the dead-zone. Since this isn't very nice I implemented "The Right Way" described here: http://www.third-helix.com/2013/04/12/doing-thumbstick-dead-zones-right.html

I have no idea if other platforms should be changed in the same way?

### DIFF
--- a/src/win32_joystick.c
+++ b/src/win32_joystick.c
@@ -560,6 +560,25 @@ void _glfwDetectJoystickDisconnectionWin32(void)
 	}
 }
 
+// Applies radial deadzone to a circular input
+static void applySignedRadialDeadzone(SHORT x, SHORT y, SHORT deadzone, float* dest)
+{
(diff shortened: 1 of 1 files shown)


elmindreda:
That's a nice deadzone variant and a nice implementation of it, but as the article itself says there's no one size fits all. Instead I think I'll remove the XInput circular deadzone and let the user implement whatever deadzone is desired.

With the addition of the [gamepad functions](http://www.glfw.org/docs/3.3/input_guide.html#gamepad) this is now possible to do in a sane way for all axes.

elmindreda:
The XInput deadzone logic has been removed with 75e44abf09621180b6c67dc44aaf60a2309f1253.

PR #1026, opened 2017-05-31
Move variable declarations to top of blocks

sebras:
This makes the project compile successfully
when -Wdeclaration-after-statement is enabled.

### DIFF
--- a/src/mir_monitor.c
+++ b/src/mir_monitor.c
@@ -138,33 +138,38 @@ GLFWvidmode* _glfwPlatformGetVideoModes(_GLFWmonitor* monitor, int* found)
         if (id != monitor->mir.outputId)
             continue;
 
-        MirOutputConnectionState state = mir_output_get_connection_state(output);
-        bool enabled = mir_output_is_enabled(output);
-
-        // We must have been disconnected
-        if (!enabled || state != mir_output_connection_state_connected)
         {
-            _glfwInputError(GLFW_PLATFORM_ERROR,
-                            "Mir: Monitor no longer connected");
-            return NULL;
+            MirOutputConnectionState state = mir_output_get_connection_state(output);
+            bool enabled = mir_output_is_enabled(output);
+
+            // We must have been disconnected
+            if (!enabled || state != mir_output_connection_state_connected)
+            {
+                _glfwInputError(GLFW_PLATFORM_ERROR,
+                                "Mir: Monitor no longer connected");
+                return NULL;
+            }
         }
 
-        int numModes = mir_output_get_num_modes(output);
-        modes = calloc(numModes, sizeof(GLFWvidmode));
 
-        for (*found = 0;  *found < numModes;  (*found)++)
         {
-            const MirOutputMode* mode = mir_output_get_mode(output, *found);
-            int width  = mir_output_mode_get_width(mode);
-            int height = mir_output_mode_get_height(mode);
-            double refreshRate = mir_output_mode_get_refresh_rate(mode);
-            MirPixelFormat currentFormat = mir_output_get_current_pixel_format(output);
+            int numModes = mir_output_get_num_modes(output);
+            modes = calloc(numModes, sizeof(GLFWvidmode));
 
-            modes[*found].width  = width;
-            modes[*found].height = height;
-            modes[*found].refreshRate = refreshRate;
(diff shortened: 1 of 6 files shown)


jcowgill:
Why do you need to enable that warning (which shouldn't be causing the build to fail anyway)? It's only relevant for C90 compatibility.

sebras:
A project I'm participating in uses GLFW and compiles both for win32 and Linux. To be backwards compatible as far back as is possible (to Windows XP?) the win32 port is normally compiled with an old versions of MSVC (2005 I believe). Due to us Linux developers unknowingly causing issues with declarations midway through functions (as gcc does not usually complain about this) we conciously enabled -Wdeclaration-after-statement for all the code in the project. Most of GLFW compiled without warnings (this is expected as the same code is also used in the win32 port), the warnings were limited to the changed three files. It would be nice to avoid the warnings completely as this allows us to consistenly compile with this flag enabled and not cause issues for users of older MSVC 2005 compiles by accident.

linkmauve:
I would actually like to change the other way for Linux-only backends.  On Linux we are lucky enough to not have to support MSVC’s legacy, and block-scoped variables help prevent common errors such as reusing a variable after the end of the block, or simply human understanding.

sebras:
Ok, so you are suggesting adding more blocks and thus new scopes to the source instead? That might would resolve the issue as well. Let me try that out.

sebras:
I attempted to just do a naïve implementation and while it silences the compiler warnings I'm not sure the commitdiff nor the resulting code looks more appealing…

elmindreda:
I'm sorry; I won't merge this.

C99 declarations is a lot to give up, especially on intermediate-heavy Cocoa, and having to maintain strict C89 discipline across the project for years to come until we drop support for VS 2010 and 2012 is a lot to ask to avoid making a compiler flag special case.

sebras:
Just so I understand you clearly. :) 

The project where we use GLFW is [MuPDF](https://github.com/ccxvii/mupdf), and since MuPDF is built for Windows using VS2005 to keep support for Windows XP onwards (we try to keep that). VS2005 complains if variables are not declared at the top of blocks.

But what you are arguing is that is not worth staying with C89 for GLFW, because you only support VS2010 onwards and C99 has a few features you like to use? So I imagine both the original patch (faa784de) and the currently proposed patch (2a23666e) are not acceptable?

> especially on intermediate-heavy Cocoa

I'm not sure I understand what you mean by this..? Care to explain a bit more? I'm not arguing against you, but even if you decide not to take either of the patches on I'd still like to understand the underlying reasons. Thanks you! :)

thedmd:
Visual Studio have a toolset which generate Windows XP compatible binaries. For 2015 it is v140_xp. We are using it in all games we releaseing.

elmindreda:
> VS2005 complains if variables are not declared at the top of blocks.

So does VS 2010, which we support. Any GLFW code that will compile on VS is already strict about declaring variables at the top of blocks. That includes `win32_*`, `wgl_*`, `egl_*`, `osmesa_*` and files without an underscore. Your patch is about extending that restriction to other platforms where system compilers all support intermingled (C99 style) declarations.

I'm arguing that introducing and maintaining this restriction for all GLFW code for years to come is disproportionately more work than making a special case in your build system so the `-Wdeclaration-after-statement` flag isn't applied to all GLFW code.

I agree that applying this flag to your application code is a good idea for the reason you stated (and I should apply it to the Windows-relevant parts of GLFW for the same reason), but it should be possible for you to apply it to your code without also applying it to all your dependencies. Also, if you want the flag applied to GLFW because you are maintaining your own local modifications of it then my adding the flag selectively via CMake should help with that bit as well.

>> especially on intermediate-heavy Cocoa

> I'm not sure I understand what you mean by this..? Care to explain a bit more?

Most APIs on macOS are fairly verbose and either OO Objective-C or OO-style C. It's common that you need to store intermediate values in long fetch sequences, either to use them directly later on or to break up what would otherwise be lines hundreds of characters long and nested in ways that are difficult to line-break clearly.

However, I see that your patch doesn't touch the macOS code so perhaps that's irrelevant.

elmindreda:
The flag has (mostly) been added with e27dc50689885736b0b79fde22c9d24046264f96.

PR #1028, opened 2017-06-04
Hide cursor instead of using blank image

### LINKED ISSUE #971: Cursor becomes visible with System Notifications in disabled cursor mode
Whenever there is a system notification, such as a "hey please update your OS" or a skype message notification in the top right, the grabbed mouse becomes visible and remains that way.

If the window is alt-tabbed, and then re-tabbed, the mouse becomes invisible again.

Even when the window is full-screened, and a system notification cannot be seen, this is still an issue.

tikhomirov:
When cursor isn't in normal mode and should be hidden, use `[NSCursor hide]`
method instead of setting it to blank image. This should prevent
situations when hidden cursor becomes visible after system notification
was shown.

Fixes #971

### DIFF
--- a/src/cocoa_init.m
+++ b/src/cocoa_init.m
@@ -352,9 +352,6 @@ void _glfwPlatformTerminate(void)
         _glfw.ns.listener = nil;
     }
 
-    [_glfw.ns.cursor release];
-    _glfw.ns.cursor = nil;
-
     free(_glfw.ns.clipboardString);
 
     _glfwTerminateNSGL();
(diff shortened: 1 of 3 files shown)


elmindreda:
Looks good! Are the calls to `[window display]` necessary, and if so why?

tikhomirov:
@elmindreda I can't remember why I added this call 😅 I checked and it works without `[window display]` so I removed this line.

elmindreda:
Thank you!

PR #1043, opened 2017-07-12
Fix use of internal joystick API in null and Win32 drivers.

thedmd:
I found two inconsistencies in joysticks API.
In null driver it is a matter of using different function signature than in declaration.
In win32 driver joystick disconnection code iterated over id's instead of _GLFWjoystick*, this can lead to crash.

### DIFF
--- a/src/null_joystick.c
+++ b/src/null_joystick.c
@@ -31,7 +31,7 @@
 //////                       GLFW platform API                      //////
 //////////////////////////////////////////////////////////////////////////
 
-int _glfwPlatformPollJoystick(int jid, int mode)
+int _glfwPlatformPollJoystick(_GLFWjoystick* js, int mode)
 {
     return GLFW_FALSE;
 }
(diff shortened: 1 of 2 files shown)


elmindreda:
Thank you!

I wonder how that ever got through CI.

thedmd:
You're welcome.

shurcooL:
> I wonder how that ever got through CI.

Is it because the CI does not run on Windows? See https://github.com/glfw/glfw/blob/e55e616f1f25a36caaa11c8ab7bcfed8f47c9b9c/.travis.yml#L8-L9.

elmindreda:
Windows builds are handled by [AppVeyor](https://github.com/glfw/glfw/blob/e55e616f1f25a36caaa11c8ab7bcfed8f47c9b9c/.appveyor.yml), but it seems it only [emitted a warning](https://ci.appveyor.com/project/elmindreda/glfw/build/1.0.465/job/gfk6s9n3t6sqa8c7#L82).

Time for some `-Werror`.

elmindreda:
Some `-Werror` has been applied with 4feede6dd5f6cae5a0e90b23966b90b5dc635771.

PR #1054, opened 2017-07-24
Added function to check if GLFW is initialized

liavt:
`glfwIsInit()` returns the current initialization state of GLFW. This would be useful to prevent repeated calls to `glfwInit()` and to ensure that GLFW is initialized before using any library functions.


shurcooL:
> to prevent repeated calls to `glfwInit()`

In what kind of situation would your program not know whether it already called `glfwInit` or not?

include/GLFW/glfw3.h
```diff
@@ -1553,6 +1553,8 @@ typedef struct GLFWgamepadstate
  *
  *  Additional calls to this function after successful initialization but before
  *  termination will return `GLFW_TRUE` immediately.
+ *	
+ *	To check whether GLFW has been initializd, call @ref glfwIsInit.
```
shurcooL: The indentation is off. Existing comments use spaces, you've used tabs.

include/GLFW/glfw3.h
```diff
@@ -1568,13 +1570,32 @@ typedef struct GLFWgamepadstate
  *
  *  @sa @ref intro_init
  *  @sa @ref glfwTerminate
+ *	@sa @ref glfwIsInit
```
shurcooL: Here too.

include/GLFW/glfw3.h
```diff
@@ -1568,13 +1570,32 @@ typedef struct GLFWgamepadstate
...
+ *	frees any remaining resources by GLFW and should be called at the end of the application.
+ *
+ *	@return `GLFW_TRUE` if initialized, `GLFW_FALSE` otherwise
+ *
+ *	@sa @ref intro_init
+ *	@sa @ref glfwInit
+ *	@sa @ref glfwTerminate
+ *	@since Added in version 3.2.1
+ *
+ *	@ingroup init
```
shurcooL: Here too.

include/GLFW/glfw3.h
```diff
@@ -1600,6 +1621,7 @@ GLFWAPI int glfwInit(void);
  *
  *  @sa @ref intro_init
  *  @sa @ref glfwInit
+ *	@sa @ref glfwIsInit
```
shurcooL: And here.

liavt:
A scenario in which that would occur is if an app supports multiple windows with no shared global state. For example, in C++, there could be a window class which needs to know whether to call glfwInit() before glfwCreateWindow().

shurcooL:
I see. If there's no shared global state, is there a risk of race conditions if multiple places decide to check if glfw is initialized and call `glfwInit` if not?

liavt:
'glfwInit' must only be called by the main thread - it is not thread safe. 'glfwIsInit' would therefore have follow the same restriction on thread safety, as it is intended to be used with 'glfwInit'

liavt:
I have updated the documentation to use spaces instead of tabs. It also now mentions that `glfwIsInit()` should only be called from the main thread.

examples/simple.c
```diff
@@ -162,4 +162,4 @@ int main(void)
     exit(EXIT_SUCCESS);
 }
 
-//! [code]
+//! [code]
```
shurcooL: This change doesn't appear to be intentional.

shurcooL:
Sounds reasonable to me (although I always hope these things don't need to be added). But, of course, I defer to @elmindreda to know whether this change makes sense.

examples/simple.c
liavt: It wasn't intentional, now fixed

tombsar:
-1 I see no reason to add this to the API. Either the client program should only call it once during startup (what I think is the best practice), or keep track of the init state itself.

A compromise might be to guarantee that `glfwInit` takes care of this internally, making it harmless to call it multiple times (at least from the main thread).

ferreiradaselva:
I forgot that `glfwInit` returns a value of success. So yeah, now I think it's not so needed. I can think of a situation where keeping track of the initialization is useful: when you use GLFW as part of an editor. Years ago I made an editor with GTK and GLFW (invisible windows). But, yeah, since `glfwInit` already returns a success value, I see no need for an extra function.

elmindreda:
If GLFW is already initialized then additional calls to `glfwInit` will [succeed immediately](http://www.glfw.org/docs/latest/intro_guide.html#intro_init_init) with no effect so there is no need for application code to avoid that. This has been the case at least since 2.x.

liavt:
I didn't realize that multiple calls to `glfwInit` return immediately , so this change is not really that neccessary

cubberstud:
i'm currently working on new bindings for c++ that make strong use of raii.
for this case it is mandatory that querying if glfwInit() has already been called since the dtor calls glfwTerminate(). so creating multiple such objects is an error, because it would call glfwTerminate() multiple times: 
i require a l-value glfw_context (as i call it) to exist for creating an window so i know that all windows will be destroyed before glfwTerminate is called (rvalues are forbidden and the respective function is deleted). however, if i create another glfw_context with a smaller scope (e.g. r-value category) it would call glfwTerminate before all windows were destroyed.

currently i need an ugly static member since i can not know if glfwInit() "returned immediately" or if this was simlply the first time i called this function. my goal is to solve this e.g. through an exception when the user should not be allowed to do so or - in the spirit of this disussion - allowing it and memorizing not to call glfwTerminate() for this instance. please reconsider accepting the proposal or share any thoughts on how i could solve this in a clean way otherwise!

tombsar:
@cubberstud The problem is that there is only one GLFW state, but your design allows for multiple `glfw_context` instances to coexist. Each instance does not "own" GLFW, and therefore should not be allowed to call `glfwTerminate`, as doing so interferes with other instances.

I don't know why you say that having a static member is "ugly". The question of whether or not GLFW has been initialized really is static data, so it makes sense to store it as such.

Have a think about what instances of your `glfw_context` class are supposed to represent, and why you allow multiple of them to exist simultaneously in the first place. Could you achieve the same thing with a singleton class representing GLFW's global "initializedness"?

cubberstud:
the static member works for the scope of my framework -although i do not like it since it seems redundant because glfw already tracks this bit of information- but consider this:
if someone is using calls glfwInitialize() without using my framework, it will get called twice and as soon as the glfw_context's dtor is run, glfwTerminate() is called since i have no way of knowing if e.g. another library is handling glfw. if i could just use glfwIsInit() in the ctor of glfw_context, it would detect that glfw is already initialized and will not attempt to call glfwTerminate() in the dtor. so this would allow for a thirdparty implementation to handle glfw. 

i know that this example might seem rather contrived, but the code would even work correctly in this case without further effort. 
as far as my framework is concerned: i do not want to use singletons, i implemented move and move-assign correctly and disabled copying. so having multiple glfw_context-instances around is no problem (i was throwing previously since i considered this a logical error but reading through this conversation i realized that this was considered fine so i changed my code such that the last instance that goes out of scope does all the cleanup).
note that i do not want to get rid of the concept of glfw_context since something like this is required for exception-saftiness and the concept of such a context gives me compile-time gurantees that the user attempted to call glfwInit() which i think is a nice way of approaching things.

Issue #894, opened 2016-11-03
No way to get/set the primary and secondary selection


kovidgoyal:
As per the docs here: http://www.glfw.org/docs/latest/input_guide.html#clipboard there does not seem to be any way to get/set the primary and secondary selections from the X11 clipboard. The X11 clipboard maintains two selections, unlike most other platform clipboards that maintain only one.

This can be worked around by using the xsel to manipulate the clipboard program, but this is less than ideal, for obvious reasons. It would be nice if glfw provided some API for this.

Maybe it already does, and I just cant find it?

elmindreda:
> As per the docs there does not seem to be any way to get/set the primary and secondary selections from the X11 clipboard.

That's correct. GLFW uses the CLIPBOARD selection, as it's the one most similar to the clipboard of other platforms. If I added functions for PRIMARY and SECONDARY selections, they wouldn't map to anything useful on platforms other than X11.

kovidgoyal:
Sure, but on X11 it is pretty important to be able to use the other selections. The ability to paste from and copy to the PRIMARY selection is widely used throughout the X11 ecosystem, usually bound to the mouse middle click to paste and select to copy. While I can appreciate it is not useful on other platforms, IMO, it is too important to leave out. Other cross platform toolkits, Qt, for example, include the ability to use it on X11, see https://doc.qt.io/qt-5/qclipboard.html#Mode-enum

elmindreda:
I'll consider it.

elmindreda:
Now that I know about kitty, I think I understand this issue better. Still not sure if or how to address it, but I'll keep an eye out when implementing #275.

kovidgoyal:
Thanks :)

knielsen:
Meanwhile, does anyone know if there is a work-around to set the primary selection?
Eg. using a different API/library, based on a GLFW-created window?

(My motivation is that I need primary selection support in mupdf-gl, which is based on GLFW).

kovidgoyal:
Just use a program like xsel. Look in the kitty source code for an example.

knielsen:
Ah, I see, nice idea. Thanks for the pointer!

knielsen:
I ended up implementing a facility in GLFW for accessing the X11 primary selection:

https://github.com/knielsen/glfw/commit/045fc5780edd215ddac70d0aad715854ac4ebd5d

[...]

PR #1056, opened 2017-08-02
Implement access to primary selection

knielsen:
This patch implements the access to X11 primary selection that was requested in issue #894.

New functions glfwSetSelectionString() and glfwGetSelectionString() are added similar to glfwSetClipboardString() and glfwGetClipboardString(). These functions do nothing on platforms with no X11-like primary selection.

There is some discussion in the issue 894 as to whether this is too X11 specific to be included in GLFW, or if it should perhaps be in the platform-specific api. Win32 does not have any similar concept to primary selection, to my knowledge. On the other hand, there are some references to primary selection in Wayland, eg. https://wiki.gnome.org/Initiatives/Wayland/PrimarySelection. So this might be non-X11 specific in the future.

In any case, I decided to create this pull request, which is rebased against latest GLFW master, in case it is useful.

I added stubs for other platforms; howver patch is currently only tested on Linux/X11 due to my lack of available other platforms.

Signed-off-by: Kristian Nielsen <knielsen@knielsen-hq.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,9 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+- Added `glfwSetX11SelectionString` and `glfwGetX11SelectionString`
+  native functions for accessing X11 primary selection as a supplement to
+  the clipboard on the X11 platform (#894)
 - Added `glfwGetError` function for querying the last error code and its
   description (#970)
 - Added `glfwUpdateGamepadMappings` function for importing gamepad mappings in
--- a/docs/input.dox
+++ b/docs/input.dox
@@ -868,6 +868,12 @@ The clipboard functions take a window handle argument because some window
 systems require a window to communicate with the system clipboard.  Any valid
 window may be used.
 
+X11 also has the concept of primary selection, which is similar to,
+but distinct from, the clipboard. If this is needed on X11, it can be
+accessed from the X11 native interface with the functions @ref
+glfwGetX11SelectionString and @ref glfwSetX11SelectionString. See
+@ref native.
+
 
 @section path_drop Path drop input
 
--- a/docs/news.dox
+++ b/docs/news.dox
@@ -5,6 +5,15 @@
 @section news_33 Release notes for 3.3
 
 
+@subsection news_33_native_x11_selection X11 native Primary Selection access
+
+GLFW now supports X11 platform specific native functions for accessing
+the X11 primary selection, as a supplement to the clipboard on this
+platform. See @ref clipboard.
+
+@see @ref error_handling
+
+
 @subsection news_33_geterror Error query
 
 GLFW now supports querying the last error code for the calling thread and its
--- a/include/GLFW/glfw3native.h
+++ b/include/GLFW/glfw3native.h
@@ -289,6 +289,56 @@ GLFWAPI RROutput glfwGetX11Monitor(GLFWmonitor* monitor);
  *  @ingroup native
  */
 GLFWAPI Window glfwGetX11Window(GLFWwindow* window);
+
+/*! @brief Sets the current primary selection to the specified string.
+ *
+ *  @param[in] string A UTF-8 encoded string.
+ *
+ *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED and @ref
+ *  GLFW_PLATFORM_ERROR.
+ *
+ *  @pointer_lifetime The specified string is copied before this function
+ *  returns.
+ *
+ *  @thread_safety This function must only be called from the main thread.
+ *
+ *  @sa @ref clipboard
+ *  @sa glfwGetX11SelectionString
+ *  @sa glfwSetClipboardString
+ *
+ *  @since Added in version 3.3.
+ *
+ *  @ingroup native
+ */
+GLFWAPI void glfwSetX11SelectionString(const char* string);
+
+/*! @brief Returns the contents of the current primary selection as a string.
+ *
+ *  If the selection is empty or if its contents cannot be converted, `NULL`
+ *  is returned and a @ref GLFW_FORMAT_UNAVAILABLE error is generated.
+ *
+ *  @return The contents of the selection as a UTF-8 encoded string, or `NULL`
+ *  if an [error](@ref error_handling) occurred.
+ *
+ *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED and @ref
+ *  GLFW_PLATFORM_ERROR.
+ *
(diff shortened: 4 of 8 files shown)


elmindreda:
This is a beautiful patch. I don't know yet whether to merge it and if so in what form, but that's because it's right on the edge of the scope of GLFW. Seems it's time to define that edge more clearly.

knielsen:
Yes.

On X11, primary selection is really important. Traditional X11 uses *only*
the primary selection. I cannot think off-hand of a program that does not
support copy-paste just by double-clicking a word in one window and
middle-clicking to paste in another. The odd program encountered that does
not support it feels very awkward.

Hence the motivation for this patch, to get proper copy-paste support in
mupdf-gl on X11. Without some support in GLFW, the only option seems to be
to spawn an external xsel program for each copy operation, which is
inefficient and requires that xsel be made available somehow.

Primary selection is now in Wayland, so this issue should eventually come up
there as well. Its implementation was made a blocker for having Wayland as
default in Fedora:

1. https://fedoraproject.org/wiki/Wayland_features#BLOCKER:_primary_selection
1. https://www.phoronix.com/scan.php?page=news_item&px=Wayland-Primary-Selection
1. https://bugzilla.redhat.com/show_bug.cgi?id=1214655

On the other hand, GLFW seems to very successfully have hidden all platform
differences behind its API. As far as I know (not being a Win32 or OS X
user), there is no concept of primary selection on those platforms. It is
ackward to have a window system feature that is important on some platforms,
and completely missing on others. I am not sufficiently familiar with GLFW
to know if a similar situation exists in another part of the GLFW API and if
so, how it was handled.

Arguments for support in GLFW:

* Primary selection is important in some places, and awkward to support without GLFW assistance.
* Default behaviour on eg. Win32 is reasonable (works the same as for an X11 user that does not use primary selection).
* Some other cross-platform frameworks also seem to support primary selection (see below).

I did some background research on how this issue is dealt with in other
cross-platform frameworks:

### Java Swing

There is getSystemSelection() to access the primary selection in addition to
getSystemClipboard(). The function does nothing on systems without the
concept of primary selection:

1. https://docs.oracle.com/javase/7/docs/api/java/awt/Toolkit.html#getSystemSelection()

### QT

Primary selection is supported through the QClipboard::Selection mode.
Interestingly, it also supports some OS X specific QClipboard::FindBuffer
mode for the current search string:

1. http://doc.qt.io/qt-5/qclipboard.html#Mode-enum

### Javascript

I could not find a way to access the primary selection from Javascript (in
browsers).

Another option for an API is some kind of hint to which selection to use -
and only the CLIPBOARD hint would be available on all platforms. Maybe this
is more in the spirit of the GLFW API. This could even be used to support
the OS X FindBuffer as in QT, perhaps, or X11 SECONDARY selection, though
not sure if those are commonly used.

I hope the above helps us end up with the right decision wrt. primary
selection support in GLFW.

elmindreda:
I don't think it's entirely fair to compare GLFW with Qt. They have different aims and very different scopes. The primary selection was deliberately excluded when clipboard support was added, as GLFW aims to provide a reliable common denominator, not like Qt enable the creation of native-like applications.

Neither SDL nor SFML provide explicit support for the primary selection. SDL does implicitly grab it when you call their equivalent of `glfwSetClipboardString', but from what I understand that wouldn't be helpful to you.

> On the other hand, GLFW seems to very successfully have hidden all platform differences behind its API. [...] I am not sufficiently familiar with GLFW to know if a similar situation exists in another part of the GLFW API and if so, how it was handled.

Until recently it's been possible to handle via by taking it into account in the initial design or via exceptions within the current API. For example, `glfwSetWindowIcon` is a no-op on macOS where window icons there are for document icons, not application icons. This release adds platform-specific hints, but using these look more like in-code metadata than new logic.

I'm very wary of adding whole functions to the public API (any function in `glfw3.h`) that only function on a single platform (or even that platform and its successor), precisely so that code written by someone only familiar with one platform will work as expected on the others.

That said, because this is low-impact, relatively small and has real applications waiting to use it I will merge this if reworked into a pair of X11-specific native access functions. There is some precedence for this with `glfwGetOSMesaColorBuffer` and any use of the primary selection will require X11-specific logic anyway.

knielsen:
Yes, native API seems a good option here. There is value in having a clean
platform neutral main API. Thanks for taking the time to think this over.

I have re-done the patch as a native X11 API now. I hope this is what you
had in mind (else just let me know).

It turns out that the window argument of the functions is not needed in
newest GLFW (it uses _glfw.x11.helperWindowHandle internally, and the X11
selection is global to the display). So I omitted the window argument from
the functions:
```c
void glfwSetX11SelectionString(const char* string);
const char* glfwGetX11SelectionString(void);
```
This is somewhat inconsistent with glfw[GS]etClipboardString() which does
take a window argument, but should be appropriate for a native API.

The existing GLFW has no tests of native APIs. I kept the test for primary
selection in the patch in tests/clipboard.c under #ifdef
GLFW_EXPOSE_NATIVE_X11. So it can be manually enabled by passing
-DCMAKE_C_FLAGS=-DGLFW_EXPOSE_NATIVE_X11 to cmake (in addition to
-DGLFW_BUILD_TESTS=ON).

For applications, this still leaves the problem of detecting (at compile
time) whether there is GLFW_EXPOSE_NATIVE_X11 support in the instance of the
GLFW library being used. That can be worked around.

It could be useful if GLFW made available some GLFW_HAVE_NATIVE_X11 or
similar - so each application does not have to redo the platform check, and
end up with linker error if it gets it wrong. I did not see any such current
mechanims. But I think that is a separate issue from this pull request.

knielsen:
Hm, I rebased this onto latest master. The Travis failure seems to have been inherited from latest master.

elmindreda:
> I have re-done the patch as a native X11 API now. I hope this is what you had in mind (else just let me know).

Looks good!

> It turns out that the window argument of the functions is not needed in newest GLFW [...]. So I omitted the window argument [...] This is somewhat inconsistent with glfw[GS]etClipboardString() which does take a window argument, but should be appropriate for a native API.

I think that's the right choice. The window parameter of the clipboard functions is a design mistake that will be removed in the next API revision.

> It could be useful if GLFW made available some GLFW_HAVE_NATIVE_X11 or similar - so each application does not have to redo the platform check, and end up with linker error if it gets it wrong. I did not see any such current mechanims. But I think that is a separate issue from this pull request.

Yeah. Once run-time selection between X11 and Wayland is in it should be possible to do using OS-specific macros.

The warnings that caused CI to fail on macOS have been fixed.

Issue #197, opened 2013-12-28
Option for transparent windows


nezumisama:
On X11, when using a compositor/compositing WM, windows can have alpha-enabled visuals, allowing the application to specify opacity of each pixel. This is not however supported by GLFW now.

Here's some example code from Nvidia (note this works on Mesa as well): ftp://download.nvidia.com/XFree86/Linux-x86/177.70/README/chapter-23.html

I hacked together a version of glxgears with transparent background, for reference: https://gist.github.com/nezumisama/8161902

andrewrk:
I'm interested in this feature for the purpose of creating a GUI toolkit, for example a menu with a drop shadow.

r-lyeh:
same here, GUI purposes

andsve:
Funny thing, I'm also in need of it myself at the moment. I'll see if I have time to look into it this weekend, I think my branch is way out of sync with GLFW3 etc...

andsve:
Hey, if anyone is still interested in this, I just pushed a fix for Windows on this branch: https://github.com/andsve/glfw/tree/transparent-windows

Interested in hearing if it works or not for everyone else.

Haven't tested my OS X chnages since I last worked on this, so will do that next. Then maybe try integrating the Linux stuff from above in this thread.

tombsar:
> Interested in hearing if it works or not for everyone else.

Works perfectly for me. Windows 8.1 64 bit (but building glfw as 32 bit) with Nvidia drivers.

andsve:
Tried to use the supplied X11 related changes that @nezumisama posted above, but I can't get it to work. I'm however running a debian install inside VirtualBox so I can't say for sure... Anyone with a Linux box that would like to apply this (using `transparent-windows` branch) and give it a try?
```diff
diff --git a/src/glx_context.c b/src/glx_context.c
index ec34e8f..28f0f7c 100644
--- a/src/glx_context.c
+++ b/src/glx_context.c
@@ -60,6 +60,8 @@ static GLboolean chooseFBConfig(const _GLFWfbconfig* desired, GLXFBConfig* resul
     int i, nativeCount, usableCount;
     const char* vendor;
     GLboolean trustWindowBit = GL_TRUE;
+    XVisualInfo *vi;
+    XRenderPictFormat *pf;
 
     vendor = glXGetClientString(_glfw.x11.display, GLX_VENDOR);
     if (strcmp(vendor, "Chromium") == 0)
[...]

PR #1078, opened 2017-09-16
transparent windows

cosier:
Hello! Great library!
I have come across the need for transparent windows _(in dealing with rounded borders for a undecorated window)_

I have looked at the existing issue for transparency options #197 
However it was incomplete, did not compile on OSX, and didn't support X11.

My pull request provides working transparency for both OSX and X11 (not aware of the current state for transparency on Windows **but am willing to work on it next**).

What I have done:

1. Checkout'd @ckaran branch ckaran/transparent_window and made it compile and added some alphaMask struct members here and there for OSX, 
2. Synced it up with the current master.
2. Added some use of XRender _(requires additional linkage on cmake)_ to detect pict_format for alphaMask (maybe there is a better alternative that doesn't require XRender?.

Transparency can be enjoyed with the following glfw hint
`glfwWindowHint(GLFW_TRANSPARENT, GLFW_TRUE)`

This is my first dive in GLFW and the community here, please let me know what I am missing or if anything is needs to be done.

There is an example application called "transparent" in examples/

transparent_window_example_linux
transparent_window_example_osx


cosier:
Looks like the windows builds are failing on AppVeyor. 
I shall look into it.

cosier:
There was some minor linkage issue with building the transparency demo on Windows.
That has been fixed, however it is lacking transparency _(the most important part!)_.

My guess is that there is no transparency implementation for wgl.
I shall look into that.

cosier:
Luckily, I was pointed to #723 , which already implemented effecient transparency for windows via DWM.
I have merged that into this branch, and we now have windows support as well.

I have tested it on Windows 7 inside a VM.
transparency_demo_windows

cosier:
It seems older MSVC (ie. 2013) does not support the c++ keyword "inline", but rather a vendor specific "__inline". So I just removed it and let the compiler deal with it.

Also after merging with _glfw:transparent-window_ branch, it already contained a transparentified gears.exe demo, thus rendering the transparent.exe demo from another dev branch **redundant**.

I have removed the transparent demo to slim this pull request. 

Currently seeking any other issues...

cosier:
Did some cleanup around duplicant transparent flags for wndconfig and fbconfig.

Since GLFW_TRANSPARENT is a framebuffer flag, I removed the other ones (as it caused additional internal signature changes to pass around the wndconfig).

Merge should slimmer now, and ready for review.

elmindreda:
(This is not the review but) this looks very promising! Thank you!

elmindreda:
@cosier I'm afraid I couldn't merge this directly as the original author metadata of the code you rebased is missing. Instead I put together the (current state of the) `transparent-window` branch, using your work as a guide through all the various revisions and adding your cleanups on top.

Does this look all right? Did I miss or mis-attribute anything?

cosier:
Ah the good ol' git rebase trick.
Looks good, I guess we can close this pull request then :-)

bit2shift:
Wouldn't it be interesting if one were to use the DWM API to extend the frame (`DwmExtendFrameIntoClientArea` with a padding of -1) all way and make the window fuly Aero Glass?
Just dropping my 2 cents (actually 2 function calls, the other being `DwmIsCompositionEnabled()`, which is required for Vista and 7 only).

elmindreda:
@cosier Thank you again so much for this PR! It honestly was immensely helpful for making sense of the backlog.

@bit2shift We can look into that once 3.3 is out. There are platform specific window hints now.

PR #1079, opened 2017-09-21
Transparent hint check

cosier:
Currently the outcome of setting window transparency is unknown.

This pull request adds the ability to get the result of the GLFW_TRANSPARENT hint window outcome by using `glfwGetWindowAttrib()`.

I have also updated the docs detailing this graceful fallback and how to check it's outcome.
Along with adding it to the gears example, as to show how to catch it.

### DIFF
--- a/examples/gears.c
+++ b/examples/gears.c
@@ -322,6 +322,11 @@ int main(int argc, char *argv[])
         exit( EXIT_FAILURE );
     }
 
+    if (!glfwGetWindowAttrib(window, GLFW_TRANSPARENT))
+    {
+        fprintf( stderr, "Failed to set GLFW_TRANSPARENT flag\n" );
+    }
+
     // Set callback functions
     glfwSetFramebufferSizeCallback(window, reshape);
     glfwSetKeyCallback(window, key);
--- a/docs/window.dox
+++ b/docs/window.dox
@@ -289,7 +289,9 @@ constraint.  Possible values are `GLFW_TRUE` and `GLFW_FALSE`.
 
 @anchor GLFW_TRANSPARENT
 __GLFW_TRANSPARENT__ specifies whether the framebuffer will support transparency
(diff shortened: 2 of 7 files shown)


examples/gears.c
```diff
@@ -322,6 +322,10 @@ int main(int argc, char *argv[])
         exit( EXIT_FAILURE );
     }
 
+    if (!glfwGetWindowAttrib(window, GLFW_TRANSPARENT)) {
```
shurcooL: Style-wise, is it okay this places `{` on same line as `if`, while the rest of the code places it next line?

cosier: Sorry that is an IDE mistake, I will fix that.

cosier:
@elmindreda  I found a bug though, regarding windows framebuffer implemention.

In order for this pull request to work, it requires detection of transparency on WGL, but that never happens.

This works on EGL and GLX context's because we are checking pixel format:
 `pf->direct.alphaMask` and setting the `_GLFWfbconfig->transparent` appropiately.

But on wgl_context implementation, this never happens.

thomthom:
Just want to add my +1 to this feature. Knowing the success of the transparent feature would be very useful for what I want to build.

cosier:
I have just pushed a fix for win32 transparency checks.
Since we are doing transparency with DWM, I had to restructure the checks around a bit.

Now with this pull request, using `glfwGetWindowAttrib()`:

1. X11: If transparency was available via Xrender and requested returns True
2. OSX: will always support transparency, the same mechanism will return True
3. Windows: If compositing mode is enabled and transparency request, returns True

I have successfully tested these checks by toggling windows DWM on and off and running the Gears example (turning off DWM by using aero basic).

I have also tested the latest commit on OSX and Linux.
Should be ready for review.

elmindreda:
I agree that this window attribute should be added.

If the semantics are "is it currently transparent?" then it will additionally need to check on both Windows and X11 whether a desktop compositor is currently running. It will also slightly complicate #25.

PR #1090, opened 2017-10-11
Send control characters to CharMod callback on X11.

ccxvii:
Control A-Z were being inhibited, but other control characters were being
passed through. With this change, we pass all control characters that are
mapped via X*LookupString to the CharMod callback.

### DIFF
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -1130,6 +1130,14 @@ static void releaseMonitor(_GLFWwindow* window)
     }
 }
 
+static void _glfwInputCharX11(_GLFWwindow* window, unsigned int codepoint, int mods, GLFWbool plain)
+{
+    /* Reverse the Ctl+[A-Z] bit-shifting that XLookupString has applied */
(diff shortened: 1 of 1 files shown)


src/x11_window.c
```diff
@@ -1130,6 +1130,14 @@ static void releaseMonitor(_GLFWwindow* window)
     }
 }
 
+static void _glfwInputCharX11(_GLFWwindow* window, unsigned int codepoint, int mods, GLFWbool plain)
```
linkmauve: You should rename this function `inputChar()` or something similar, the `_glfwInput` prefix is reserved for message passing from the backend to the core of the library.

src/x11_window.c
```diff
@@ -1130,6 +1130,14 @@ static void releaseMonitor(_GLFWwindow* window)
...
+{
+    /* Reverse the Ctl+[A-Z] bit-shifting that XLookupString has applied */
```
linkmauve: Ctrl, not Ctl.

elmindreda:
The charmods callback has been deprecated and no further fixes will be made to it.

PR #1094, opened 2017-10-14
Pass shared context and parent window as window hints.

sagaceilo:
This CL doesn't include other platform since I'm not familiar with them.
Nevertheless, it would be cool to be able to pass shared context and
parent window as optional window hints.
CreateWindow no more need extra parameter to be passed (I assume 99% of
time, no one even use this). hints.context.share already exists. Its
just set in glfwChildWindowHind (the name can be changed, I really
didn't come with anything better).


tombsar:
I'm cautiously in favour of this, as I agree that most users never set these to anything other than `NULL`.

Something to be looked into for version 4.

elmindreda:
Definitely agree that share context should not be a parameter to `glfwCreateWindow` in the next API revision. Contexts and windows need to be decoupled in many ways.

There are no plans to support creating child GLFW windows of other GLFW windows, however (as opposed to attaching existing native child windows as GLFW window object, which likely will happen).

PR #1115, opened 2017-11-01
Improve DPI change behavior on Windows

Spasi:
When a window transitions between monitors with different scale/dpi settings:

1. It's automatically resized and repositioned by handling the `WM_DPICHANGED` event (Windows 8.1 or newer)
2. Its non-client area is automatically updated to match the new DPI (Windows 10 Anniversary Update or newer)

In Windows 10 Creators Update the [DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2](https://msdn.microsoft.com/library/windows/desktop/mt791579(v=vs.85).aspx) mode was introduced which includes the `EnableNonClientDpiScaling` behavior. However, I don't think we need all its features and supporting it in GLFW would be more complicated (required calls: `SetProcess/ThreadDpiAwarenessContext` and `IsValidDpiAwarenessContext`), so I left it out for now. Calling `EnableNonClientDpiScaling` is harmless, if an application sets the `_V2` context manually.

I recorded a video of the improved behavior, [here](https://www.youtube.com/watch?v=ZkHp1I06YTk).

### DIFF
--- a/src/win32_init.c
+++ b/src/win32_init.c
@@ -98,6 +98,8 @@ static GLFWbool loadLibraries(void)
 
     _glfw.win32.user32.SetProcessDPIAware_ = (PFN_SetProcessDPIAware)
         GetProcAddress(_glfw.win32.user32.instance, "SetProcessDPIAware");
+    _glfw.win32.user32.EnableNonClientDpiScaling_ = (PFN_EnableNonClientDpiScaling)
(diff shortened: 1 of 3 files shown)


elmindreda:
With the addition of `GLFW_SCALE_TO_MONITOR` I believe this is done.

Spasi:
Yes, thank you, dpi transitions work great with `GLFW_SCALE_TO_MONITOR`.

PR #1123, opened 2017-11-09
Add occlusion callback and window attribute

GreenLightning:
Adds `glfwSetWindowOcclusionCallback(GLFWwindow* window, GLFWwindowocclusionfun cbfun)` and `GLFW_OCCLUDED` window attribute as a first step towards #680.

Only implemented on macOS for now.

### DIFF
--- a/docs/window.dox
+++ b/docs/window.dox
@@ -1150,6 +1150,12 @@ void window_occlusion_callback(GLFWwindow* window, int occluded)
 }
 @endcode
 
+You can also get the current occlusion state with @ref glfwGetWindowAttrib.
+
+@code
+int occluded = glfwGetWindowAttrib(window, GLFW_OCCLUDED);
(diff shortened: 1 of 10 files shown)


elmindreda:
Thank you, will review!

Miigon:
This feature is so helpful. Please review & merge. Thanks.

elmindreda:
As a side-effect of fixing #1337, OpenGL vsync now uses a display link and is not affected by window occlusion. This idea of window occlusion doesn't seem to exist on platforms other than macOS so it would be a macOS-only window attribute and I would prefer not adding that.

GreenLightning:
The intention was to help users build custom workarounds for the high cpu usage issue caused by the weird vsync behavior of macOS. Since that is all fixed now, this pull request is not needed any more.

PR #1131, opened 2017-11-16
Add Wayland to the CI

linkmauve:
This adds two new CI platforms, Wayland with shared libs and with static libs.

There are a few fixes for the Wayland platform bundled, notably it will disable the compose key when xkbcommon is too old to support it.


linkmauve:
This should now be good to go. :)

Issue #893, opened 2016-11-02
Init hints for ICCCM WM_CLASS components


kovidgoyal:
I'd like a small enhancement to the hack for setting WM_CLASS. Currently both parts of the WM_CLASS are set to the initial title. Instead, could the title be split on the 30 character (ASCII record separator) so that the name and class parts of WM_CLASS can be set separately? Thanks.

elmindreda:
Thanks! Will look into it.

elmindreda:
This would be better served via platform specific hints than introducing magic into window titles.

kovidgoyal:
Sure, if you are willing to do the work to setup the infrastucture for that, great. That would help with other things as well, such as support for the different types of clipboards in X11.

However, my suggested modification is much less work. Which approach you choose is entirely up to you, since you are the one who will be maintaing this codebase.

elmindreda:
Which convention are you planning to use for WM_CLASS?

kovidgoyal:
Not sure what you are asking? I am fine with any approach that allows me to set both parts of WM_CLASS independently. Either use a record separator in the title hack or setup some api for adding platform specifi properties to windows.

elmindreda:
I'll be adding the `glfwInitHint` function similar to the current `glfwWindowHint` for (mostly platform specific) options that must be known at initialization to move them from compile-time to run-time and cut down on the need for people to (re-)compile their own binaries. Since WM_CLASS is used to expose application-global information, it seems to be a good fit for that mechanism.

I propose adding two X11 specific hints, one for each component of the property. These strings would then be used for all top-level X11 windows created by GLFW. Would this be sufficient for your needs?

kovidgoyal:
Yes, that is fine, thank you :)

PR #1139, opened 2017-11-20
Support for setting WM_CLASS on a per-window basis

kovidgoyal:
The current methods of setting WM_CLASS are:

1) Use the title set a creation time. Does not allow setting the name and class parts of WM_CLASS
2) Use an init hint. Does nto allow setting different WM_CLASS for different windows in the same application.

My proposed method addresses both od these shortcomings, by encoding the WM_CLASS and title values into a single string demarcated by ASCII start-of-header and record separator bytes (which do not occur in the utf-8 encoding). If the title does not start with start-of-header, the old behavior from (1) is used.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -2343,7 +2343,12 @@ GLFWAPI void glfwWindowHint(int hint, int value);
  *  @remark @x11 The name and class of the `WM_CLASS` window property will by
  *  default be set to the window title passed to this function.  Set the @ref
  *  GLFW_X11_WM_CLASS_NAME and @ref GLFW_X11_WM_CLASS_CLASS init hints before
- *  initialization to override this.
+ *  initialization to override this. You can also set the title int he following
+ *  special format, which allows setting the two parts of the WM_CLASS property
+ *  and the window title independently:
+ *  <01> WM_CLASS name <30> WM_CLASS class <30> title
+ *  Here <01> refers to the byte value 01 (ASCII start-of-header) and <30> refers
+ *  to the byte value 30 (ASCII record separator).
  *
  *  @remark @wayland The window frame is currently unimplemented, as if
  *  [GLFW_DECORATED](@ref GLFW_DECORATED_hint) was always set to `GLFW_FALSE`.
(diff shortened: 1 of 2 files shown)


elmindreda:
I believed we had [agreed to](https://github.com/glfw/glfw/issues/893#issuecomment-272943483) the current `glfwInitHintString` solution. If that's insufficient then I need to remove it before 3.3 release and replace it with something else.

What are you planning to set `WM_CLASS` to that will vary between windows?

kovidgoyal:
I just finished adding support for daemon mode to kitty. In which the first kitty window becomes the master and subsequent invocations cause the master to open new windows. The new windows can have different name/class values. I apologize for the change, but at the time that I thought the original fix was sufficient, I had not yet conceived of the daemon mode.

include/GLFW/glfw3.h
```diff
@@ -2343,7 +2343,12 @@ GLFWAPI void glfwWindowHint(int hint, int value);
  *  @remark @x11 The name and class of the `WM_CLASS` window property will by
  *  default be set to the window title passed to this function.  Set the @ref
  *  GLFW_X11_WM_CLASS_NAME and @ref GLFW_X11_WM_CLASS_CLASS init hints before
- *  initialization to override this.
+ *  initialization to override this. You can also set the title int he following
```
ghuser404: > in the

elmindreda:
@kovidgoyal Would 79295048b2d049d54261eeffbd535749470908a0 work for you?

kovidgoyal:
Looks good to me, thank you.

elmindreda:
@Spasi How bad an idea would it be for me to merge 79295048b2d049d54261eeffbd535749470908a0 considering LWJGL is already shipping and exposing `glfwInitHintString`?

Spasi:
@elmindreda It's not a problem. GLFW functions are [annotated](https://github.com/LWJGL/lwjgl3-generated/blob/070102c64701cdbea00f3472d8cf158409786e1d/java/org/lwjgl/glfw/GLFW.java#L902) with the version they were introduced. We're shipping a pre-release build of GLFW 3.3 and `glfwInitHintString` was added in that version, so we are not bound by any compatibility requirements.

PR #1141, opened 2017-11-20
Implement glfwWindowBell

kovidgoyal:
Exposes the platform specific Bell/Beep API to make the default
system beep sound.


kovidgoyal:
Isn't mir dead?

ghuser404:
Perhaps. But I thought things were still being added to those files, to make it compile.

kovidgoyal:
Aren't backend files conditionally compiled? Anyway, if @elmindreda is OK with actually merging this functionality into glfw, then adding a dummy function to mir_window.c is trivial. But I dont want to spend the time until I know this PR is something she actually wants.

elmindreda:
This could be a useful thing to have. Will have a proper look after 3.3 is out.

I don't agree with it having a platform-specific "ioctl" parameter and would rather have a simple interface like `NSBeep`.

The Mir backend might be removed now that Mir is implementing the Wayland client API, so depending on timing it may not need a Mir stub at merge time.

kovidgoyal:
The patform specific parameter is there to control the volume/pitch of the beep, which is functionality I need.

elmindreda:
Why do you only need to control this on a single platform?

kovidgoyal:
I would like to control it on other platforms as well, I just dont know of an API to do so. I dont want to deprive X11 users of the ability to control the bell just because other OSes dont provide that ability.

ghuser404:
The same could be said about `glfwRequestWindowAttention`, Windows gives more control then what GLFW provides us with.

kovidgoyal:
Yes, but I dont need the extra control for request attention, I do need it for bell. Anyway, I dont really want to argue about it. Since I maintain my own fork of glfw in any case, I dont really care if you accept this PR or not.

ghuser404:
I'm not arguing with you, I'm adding bits to the discussion. It's up to @elmindreda to decide what goes into GLFW.

jadahl:
The "window bell" is not strictly an audio feature, it's an accessibility feature. If you need a beep and control the volume and other details about it, just play a sound instead. The bell may just as well be a visual bell on some systems and not have a sound at all.

I think adding any parameter to controlling any "bell" is a bad idea, and it will be completely ignored on Linux (as on modern systems, XBell just ends up as a libcanberra sound with all parameters from XBell being ignored IF the bell even happen to be audioable).

r00ster91:
Is this really useful? What is the difference between this (`glfwWindowBell(window);`) and `printf("\a");`? Don't they both do the same? I assume the `\a` escape sequence is very portable since it's provided by C itself so I don't really see the need in this.

jadahl:
> `printf("\a");`?

I assume this needs the glfw application to run in a terminal, which is probably not very common.

jadahl:
> I assume this needs the glfw application to run in a terminal, which is probably not very common.

Also, when the system bell is a visual bell (e.g. used by someone deaf), it'd flash the terminal, instead of the actual GLFW window.

PR #1143, opened 2017-11-21
Fix uninitialized variable use

kovidgoyal:
Pointed out by compiling with -Og -g3 (gcc)

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -131,7 +131,7 @@ static void pointerHandleAxis(void* data,
 {
     _GLFWwindow* window = _glfw.wl.pointerFocus;
     double scrollFactor;
-    double x, y;
+    double x = 0.0, y = 0.0;
 
     if (!window)
         return;


linkmauve:
This should have been fixed already by https://github.com/glfw/glfw/pull/1131/commits/36a5f8c22ae1d7ffdb0edee3f03660f751a5ecb8

Can you reproduce your original issue on master?

kovidgoyal:
At the time I sent this PR it was against master.

elmindreda:
What do you think of 43801d1c736c2e4837f88270472b90ffad62af31?

kovidgoyal:
Looks OK to me.

elmindreda:
@linkmauve Is 43801d1c736c2e4837f88270472b90ffad62af31 acceptable?

linkmauve:
It looks fine, but the `(x, y) == (0.0, 0.0)` case should never be possible to reach with a conform compositor, hence the previous assert.

Is there still any reason for this btw?  On both clang and gcc, the assert removed the warning that led to this PR.

elmindreda:
@linkmauve Only that I don't feel entirely comfortable with `assert(false)`.

Would it be okay to instead of the default case (and instead of the above patches) start the function with something like
```c
assert(axis == WL_POINTER_AXIS_HORIZONTAL_SCROLL || axis == WL_POINTER_AXIS_VERTICAL_SCROLL);
```
to better capture the intent?

elmindreda:
This has been addressed by 36a5f8c22ae1d7ffdb0edee3f03660f751a5ecb8 and 8dab9f6ab1388c8ec73d7b341abbe69e2164308c.

PR #1152, opened 2017-12-03
Implement xdg-shell in addition to wl_shell

linkmauve:
This protocol matches desktops much better than the deprecated wl_shell, fixing a bunch of race conditions, removing undefined behaviour, adding missing features, and generally providing a much more user-friendly experience.

Since most compositors don’t support it yet, the wl_shell_surface role is kept as fallback for now.

I haven’t tested this new code at all, since my compositor doesn’t support it yet, hence the help wanted label.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ script:
           sudo dpkg -i extra-cmake-modules_5.38.0a-0ubuntu1_amd64.deb;
           git clone git://anongit.freedesktop.org/wayland/wayland-protocols;
           pushd wayland-protocols;
-          git checkout 1.6 && ./autogen.sh --prefix=/usr && make && sudo make install;
+          git checkout 1.12 && ./autogen.sh --prefix=/usr && make && sudo make install;
           popd;
       fi
     - cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} -DGLFW_USE_WAYLAND=${USE_WAYLAND} ..
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -277,7 +277,7 @@ if (_GLFW_WAYLAND)
 
     find_package(Wayland REQUIRED Client Cursor Egl)
     find_package(WaylandScanner REQUIRED)
-    find_package(WaylandProtocols 1.6 REQUIRED)
+    find_package(WaylandProtocols 1.12 REQUIRED)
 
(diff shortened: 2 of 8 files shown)


jadahl:
@linkmauve Just pushed http://github.com/jadahl/mutter/commits/wip/xdg-shell . It's mostly complete, but still has the https://bugzilla.gnome.org/show_bug.cgi?id=791865 bug as far as I can tell. Anyhow, if you need some xdg-shell implementation test with, you can give that branch a try.

linkmauve:
I have now tested against @jadahl’s Mutter branch, it works as well as the `wl_shell` codepath there!

Two small notes though:
- There is still somewhat of a race between the `xdg_surface::configure` and the `wl_output::enter` events, the only potential issue caused by that is that on HiDPI monitors we will draw the first two frames at scale=1 instead of the monitor’s.
- The autoIconify hint defaults to `GLFW_TRUE`, which is quite surprising as it requires the user to alt-tab to the newly open window instead of it being displayed right away.  It may be the expected behaviour on other systems, but on Linux I’ve never encountered a program doing this.

jadahl:
> There is still somewhat of a race between the xdg_surface::configure and the wl_output::enter events, the only potential issue caused by that is that on HiDPI monitors we will draw the first two frames at scale=1 instead of the monitor’s.

This is an issue with Wayland and should be addressed there. Nothing GTK+ has the same issue, and there is nothing GLFW can do about it yet really.

> The autoIconify hint defaults to GLFW_TRUE, which is quite surprising as it requires the user to alt-tab to the newly open window instead of it being displayed right away. It may be the expected behaviour on other systems, but on Linux I’ve never encountered a program doing this.

Auto-iconify by default? That doesn't sound sane, why would one start a GLFW application minimized?

elmindreda:
Auto-iconify doesn't mean the window is created iconified; only that it will be iconified when it later loses input focus. Full screen windows are given input focus on creation on other platforms, to the extent that the `GLFW_FOCUSED` hint applies only to windowed mode windows (a weirdness that maybe should go away in 4).

I realize very little of this makes sense on Wayland.

However, the intent of auto-iconify has little to do with iconification and more to do with having the topmost undecorated window go away politely at Alt+Tab, mostly on Windows where there's no real high-level full screen window concept (not counting DXGI swapchain hacks in OpenGL ICDs).

linkmauve:
So that means I should special-case the first configure event received to avoid iconifying the window when it just got created and not yet activated?

elmindreda:
@linkmauve If possible, yes.

linkmauve:
This is now fixed.

PR #1158, opened 2017-12-09
[GLES] Include proper `glext` header with GL ES 3.x

Tuetuopay:
When using OpenGL ES 3.x headers, setting the `GLFW_INCLUDE_GLEXT` macro would include `GLES2/gl2ext.h` instead of the version 3 (`GLES3/gl3ext.h`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+- [GLES] Include proper `glext` headers when required with OpenGL ES 3.x
 - Added `glfwGetError` function for querying the last error code and its
   description (#970)
 - Added `glfwUpdateGamepadMappings` function for importing gamepad mappings in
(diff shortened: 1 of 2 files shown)


elmindreda:
In what environment is this file located at `GLES3/gl3ext.h`?

Tuetuopay:
@elmindreda my test environment is Ubuntu 16.04 on a WandBoard Quad (ARM SoC). However, I just checked the norm and it specifies `GLES2/gl2ext.h`. Bummer. Especially since the `glfw` package includes ... `GLES3/gl2ext.h` o_O

elmindreda:
Ah, yes, sorry about that. That was fixed with ca21a5c7b40310bab420818c827c986865dbbd0d in GLFW 3.2. I see that 16.04 only provides 3.1.2, which is the release _just_ before that.

PR #1161, opened 2017-12-19
Implement window decorations

### LINKED ISSUE #1068: No window decorations or close callback on Wayland
I'm aware Wayland is currently experimental but I figured I'd bring up the issues I'm having with it since I haven't seen many posts regarding it. When using the wayland backend I have no window decorations regardless of what the GLFW_DECORATIONS hint is set to and my close callback never seems to get called and as far as I've found the only way for me to close the program is by ending the process directly, using Ctrl+C if I'm running from shell, or by adding a listener to escape or some other key that performs a close. The mouse also seems to be larger when hovering over my glfw window when compared with what it should be. I'm using gnome 3.24 on Debian sid as my wayland server.

linkmauve:
This PR implements window decorations on Wayland.

The current design is to allocate a single buffer mapped from a single pixel, and to scale it using a `wp_viewport` on each of the four surfaces.  This prevents buttons and text from being drawn there, but at least for text I’m not sure we want that inside of GLFW.

Resizing from the left, top-left or top side moves the right and bottom sides while they should stay in place, I’m not sure why, but I wouldn’t consider that a blocker either.

Disabling decorations, fullscreen, maximizing, and hidpi have all been tested now.

Here is how it looks on Weston: 
wayland-screenshot

Fixes #1068.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -2453,10 +2453,11 @@ GLFWAPI void glfwWindowHintString(int hint, const char* value);
  *  GLFW_X11_CLASS_NAME and @ref GLFW_X11_INSTANCE_NAME window hints to override
  *  this.
  *
- *  @remark @wayland The window frame is currently unimplemented, as if
- *  [GLFW_DECORATED](@ref GLFW_DECORATED_hint) was always set to `GLFW_FALSE`.
- *  A compositor can still emit close, resize or maximize events, using for
- *  example a keybind mechanism.
+ *  @remark @wayland The window frame is currently very simple, only allowing
+ *  window resize or move.  A compositor can still emit close, maximize or
+ *  fullscreen events, using for example a keybind mechanism.  Additionally,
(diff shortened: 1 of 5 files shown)


linkmauve:
This has been updated to work on xdg-shell too.

gracicot:
I have a question. I use KDE desktop, which implements server side decorations on wayland. Will glfw use those provided by the wayland compositor in that case? Thanks.

linkmauve:
No, it will neither use them on current master nor with this PR.

I will add support for it once it reaches wayland-protocols, see the discussion on wayland-devel@ in October, it got a few comments to be resolved before its inclusion.

emersion:
Note that not all compositors implement the viewporter protocol.

Cloudef:
I just hope compositors implement SSD instead.

emersion:
Update: here's a v2 proposal for a decorations protocol: https://lists.freedesktop.org/archives/wayland-devel/2018-February/037119.html

linkmauve:
Support for server-side decorations has been added in #1341, please test.

PR #1196, opened 2018-01-27
Add support for kernels < 2.6.39

a3f:
Broke build for me on a 2.6.32 machine:
http://www.cpantesters.org/cpan/report/d956d128-0339-11e8-b0d1-b6c4abd39192

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,7 @@ information on what to include when reporting a bug.
 - [X11] Bugfix: Latin-1 text read from selections was not converted to UTF-8
 - [X11] Bugfix: NVidia EGL would segfault if unloaded before closing the display
 - [Linux] Moved to evdev for joystick input (#906,#1005)
+- [Linux] Added manual definition of joystick `SYN_DROPPED` to support kernels < v2.6.39 (#1196)
 - [Linux] Bugfix: Event processing did not detect joystick disconnection (#932)
 - [Linux] Bugfix: The joystick device path could be truncated (#1025)
 - [Linux] Bugfix: `glfwInit` would fail if inotify creation failed (#833)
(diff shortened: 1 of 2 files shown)


xCuri0:
Isn't Linux 2.6 EOL ?

a3f:
What's the minimum Linux version supported by GLFW?

elmindreda:
@a3f Wow, that's really old.

1. Does GLFW behave correctly on that kernel with this fix or does it just compile okay? Have you checked the documentation for the corresponding version of the API?
2. Are you willing to maintain compatibility with this kernel version in future releases of GLFW so it won't rot?

a3f:
I've a Perl module with GLFW as dependency, which was smoke-tested by said v2.6.32 Linux (CentOS-6 supported till 2020). I know it builds correctly with patch applied, but I haven't tested functionality and can't commit to maintaining compatibility in future.

Kernel [v2.6.39](https://elixir.free-electrons.com/linux/v2.6.39/ident/SYN_DROPPED) was the first to indicate `SYN_DROPPED`, so in any lower version that event code would never be read, branch never taken and all is well. So, I thought it looks like innocuous enough of a fix to be actually considered.

README.md
```diff
@@ -124,6 +124,7 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+- Added manual defintion of joystick `SYN_DROPPED` for backwards compatibility with Linux < 2.6.39 (#1196)
```
elmindreda: This is a Linux-specific workaround and should be listed under `- [Linux]` (without `Bugfix:`) below.

src/linux_joystick.c
```diff
@@ -38,6 +38,10 @@
 #include <string.h>
 #include <unistd.h>
 
+#ifndef SYN_DROPPED // < 2.6.39 kernel headers
```
elmindreda: Also mention which dist and version requires this and when support for it ends so this workaround can be removed.

Issue #1197, opened 2018-01-29
Compiler warnings in wayland code


kovidgoyal:
```
wl_window.c: In function ‘xdgToplevelHandleConfigure’:
wl_window.c:299:14: error: variable ‘resizing’ set but not used [-Werror=unused-but-set-variable]
     GLFWbool resizing = GLFW_FALSE;
```

And the issue I tried to fix in #1143 is still not fixed

```
wl_init.c: In function ‘pointerHandleAxis’:
wl_init.c:167:5: error: ‘y’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
     _glfwInputScroll(window, x, y);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
wl_init.c:167:5: error: ‘x’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
```

Is initializing x and y to zero, really such a big deal? The asserts dont make the warning go away when compiled as:

```
gcc -MMD -Wno-missing-field-initializers -Wall -std=c99 -D_XOPEN_SOURCE=700 -Werror -g3 -Og -DDEBUG -fwrapv -fstack-protector-strong -pipe -fvisibility=hidden -fPIC -pthread -fpic -D_GLFW_WAYLAND -D_GLFW_BUILD_DLL ...
```

kovidgoyal:
```
gcc -v                                                                                                                                                                                                        1   master 
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
[...]

PR #1203, opened 2018-01-30
Fix -Werror in Travis, env gets overriden

linkmauve:
The global env wasn’t global enough.

Prevents #1197 from happening again.


elmindreda:
Is this ready to merge?

linkmauve:
Yes, I just wanted to merge #1202 before, in order to avoid failing every build due to master not checking anymore.

PR #1220, opened 2018-02-24
Add a FreeMonitor platform API

linkmauve:
This fix the leak of a wl_output object everytime a monitor is freed.

I haven’t checked whether other winsys need a similar custom `free()` (valgrind is helpful at that), but it shouldn’t hurt to add them later.


elmindreda:
Looks good.

PR #1228, opened 2018-03-05
Compile platform-specific hints conditionally

linkmauve:
This prevents them from being compiled on unrelated platforms.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -51,10 +51,12 @@ static GLFWerrorfun _glfwErrorCallback;
 static _GLFWinitconfig _glfwInitHints =
 {
     GLFW_TRUE,      // hat buttons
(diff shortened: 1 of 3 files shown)


elmindreda:
I avoided doing this to keep the number of platform-specific ifdefs in shared code to a minimum, and to not add complexity to what is only a few lines of code.

elmindreda:
This change would also need to avoid emitting `GLFW_INVALID_ENUM` for the platform-specific hints, as setting them is [valid on any platform](http://www.glfw.org/docs/3.3/window_guide.html#window_hints).

elmindreda:
I would prefer dropping this, though I agree the current design in `master` isn't ideal.

Issue #1219, opened 2018-02-24
Test GLFW’s Wayland backend on non-Linux POSIX OSes


linkmauve:
Since #1210 got merged, a dependency on timerfd has been added which is a Linux extension.

We should have a CI pass to guarantee that any such non-POSIX extension isn’t used, or at least a way to test it somehow.

myfreeweb:
Hi! I just compiled 3.2.1 on FreeBSD with Wayland. Works great. Haven't tested master yet.

But we do have a timerfd implementation (emulation) in [libepoll-shim](https://github.com/jiixyj/epoll-shim), which is a library libwayland itself uses anyway. So you can just add `-lepoll-shim` on FreeBSD and DragonFly at least (and I'm not sure Wayland works on any other non-Linux OS…)

linkmauve:
That’s great news! The timerfd usage got merged recently so it will only be available on master.

I think I’m fine with users of other OSes reporting back issues whenever they find it, rather than me trying to figure out which parts of POSIX and extensions they may be using.

Do you know of a FreeBSD CI system we could plug here, btw?

myfreeweb:
I have a buildbot at home, it's not very public though :D

You can [cross-compile like Rust does it](https://github.com/rust-lang/rust/blob/master/src/ci/docker/dist-x86_64-freebsd/Dockerfile) I guess.

linkmauve:
Fyi, apparently Travis now supports FreeBSD: https://github.com/mpv-player/mpv/pull/7754

Someone who cares could set it up.

Issue #1174, opened 2018-01-05
Load libwayland-client at runtime


linkmauve:
Once #1173 is merged, libwayland-client will be the last Wayland-specific shared object GLFW links against.

The main issue with this library is that protocol headers, generated by `wayland-scanner` and provided either alongside libwayland for `wayland-client-protocol.h`, or built by GLFW (see e.g. `build/src/wayland-idle-inhibit-unstable-v1-client-protocol.h`), have `static inline` functions calling into it.

I haven’t been able to figure out how to make those functions defined in the protocol header call our own function aliases, the ones defined in `_glfw.wl` and loaded during `glfwInit()`.

To make it worse, these protocol headers also depend on `wl_interface` symbols which aren’t functions, exported by libwayland-client.so or internal to GLFW for other protocols, and I haven’t been able to change them using our usual `#define` dance either.

jadahl:
You can do what I made SDL do here: http://hg.libsdl.org/SDL/rev/b326f93e1bc5

In short, you don't use the protocol symbols from libwayland-client, but generate your own that can be bundled. The only libwayland-client symbols you load on demand are the non-generated one.

PR #1235, opened 2018-03-19
Add support for Wayland on FreeBSD and other OSes

myfreeweb:
re: #1219.

Mostly just ifdefs Linux joystick support to Linux. I also added linking to libepoll-shim in CMake though.

BTW, any plans for selecting between Wayland and X11 dynamically (on the fly) like the big toolkits (incl. SDL) do? [Some projects](https://github.com/kovidgoyal/kitty) are already implementing this by themselves — by building two copies of bundled glfw and dlopen'ing one of them depending on the desired backend…

### DIFF
--- a/CMake/modules/FindEpollShim.cmake
+++ b/CMake/modules/FindEpollShim.cmake
@@ -1,0 +1,17 @@
+# Find EpollShim
+# Once done, this will define
+#
+#   EPOLLSHIM_FOUND - System has EpollShim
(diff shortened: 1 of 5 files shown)


linkmauve:
Selecting the backend dynamically depends on #1174 (and similar ones for X11), after which we will have to move `_glfwPlatform*` functions to be static and put into a “backend struct” which every function will call into.

That probably won’t end up in 3.3, but is definitely planned.

And LGTM, I also tested on Linux and found no regression, so I’ll merge it. :)

PR #1261, opened 2018-05-02
Single GLRC for multiple windows

LukasBanana:
In my branch `single-GLRC` I added a new CMake option `GLFW_OPENGL_SINGLE_GLRC` (OFF by default), to let GLFW create only a single OpenGL context but to be used for multiple windows when a shared context is specified.

I'm using it to have multiple windows but avoiding resource management of OpenGL objects that are not sharable between contexts: [Vertex Array Objects](https://www.khronos.org/opengl/wiki/Vertex_Specification#Vertex_Array_Object) and [Query Objects](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_occlusion_query.txt) for instance.

For this implementation, I added the global macro `_GLFW_OPENGL_SINGLE_GLRC` (when the respective CMake option is ON) and the following member to the `_GLFWcontext` struct:
```c
#ifdef _GLFW_OPENGL_SINGLE_GLRC
    // Specifies whether this _GLFWcontext has its own GLRC object.
    // If not, it must not delete the GLRC object on destruction.
    GLFWbool            customctx;
#endif
```

I implemented it for Windows (WGL), Linux (GLX), and MacOS (NSGL) contexts, but not for EGL.
I tested the implementation with the "sharing" Example on _Windows 10_, _Kubuntu 17.10_ (VirtualBox), and _macOS High Sierra_.

The functions `wglMakeCurrent` and `glXMakeCurrent` functions allow to specify a new drawable but the previous GL context can remain. On MacOS, the view is exchanged:
```c
static void makeContextCurrentNSGL(_GLFWwindow* window)
{
    ...
#ifdef _GLFW_OPENGL_SINGLE_GLRC
        [window->context.nsgl.object setView:window->ns.view];
#endif
        [window->context.nsgl.object makeCurrentContext];
    ...
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ option(GLFW_BUILD_TESTS "Build the GLFW test programs" ON)
 option(GLFW_BUILD_DOCS "Build the GLFW documentation" ON)
 option(GLFW_INSTALL "Generate installation target" ON)
 option(GLFW_VULKAN_STATIC "Use the Vulkan loader statically linked into application" OFF)
+option(GLFW_OPENGL_SINGLE_GLRC "Build the GLFW to use only a single OpenGL render context" OFF)
 
 if (UNIX)
     option(GLFW_USE_OSMESA "Use OSMesa for offscreen context creation" OFF)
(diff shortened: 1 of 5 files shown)


tombsar:
Are there any downsides to having a single GLRC across multiple windows? Unless there's something I'm overlooking, it seems that this would be a sensible default behaviour so long as it is still possible for the user to request separate shared contexts if they need them for some reason.

Possible candidate for 3.4.

LukasBanana:
It's been a while that I my GLFW fork at work, but I remember that we had an issue on one platform. I don't remember if it was Mac or Linux, but we had a customized Linux, so maybe it was that.
However, I would still recommend to use it as an option but not as the default one.

PR #1271, opened 2018-05-20
Use HiDPI cursors on scale>1 monitors on Wayland

linkmauve:
This prevents them from looking blurry.

Note that only the NULL cursor and the ones used for decorations (resize, etc.) are rendered at scale=2 currently.  The cursors created by glfwCreateStandardCursor() and glfwCreateCursor() contain a single image which is specified to be scale=1.

In the future, GLFWcursor should be extended to support that usecase, as well as animations.


linkmauve:
Thanks to the improvements from #1323, standard cursors now support HiDPI too, only user-defined ones still don’t.

PR #1285, opened 2018-06-08
Fix a race condition between glfwPostEmptyEvent and glfwWait***Events on X11 and Wayland

### LINKED ISSUE #1281: glfwPostEmptyEvent sometimes fails to unblock glfwWaitEvents on Ubuntu 18.04
From times to times, the empty event posted by a call to `glfwPostEmptyEvent` won't unblock a call to `glfwWaitEvent` in the main thread.

A minimal program reproducing the issue [is available here](https://github.com/OlivierSohn/repro-evt)
It can be built and run on linux like so:
```g++ main.cpp -std=c++14 -lglfw -lX11 -ldl -lpthread && ./a.out```

If the calls of `glfwPostEmptyEvent` and `glfwWaitEvent` don't overlap in time, then the bug will never happen, but if the 2 calls are concurrent, the bug happens, "sometimes".

Note that on OSX, this bug never happens.

GLFW Version (from glfwGetVersionString()):
`3.2.1 X11 GLX EGL clock_gettime /dev/js Xf86vm shared`

Ubuntu version
`18.04 LTS`

- Note that I have tried posting "several" empty events at the same time: it doesn't help either, it just makes the bug a little less reproducible but doesnt fix it entirely.

Edit : added the build command
Edit2 : fixed the build command

OlivierSohn:
This PR closes #1281 by using a self pipe to write when an empty even has been posted, and by using `ppoll` in place of `select` to wait on events.

Benchmarks on Ubuntu/x11 have shown that the timeout precision is unchanged.

@kovidgoyal tested a prior version on wayland, it was working. Since then, I had to make changes while cherrypicking from his fork. It compiles, but I couldn't test it. Could someone run this program to verify that the bug is fixed on wayland too?
https://github.com/OlivierSohn/repro-evt/blob/master/main.cpp

Thanks!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,13 +18,13 @@ elseif (_GLFW_WIN32)
                      win32_monitor.c win32_time.c win32_thread.c win32_window.c
                      wgl_context.c egl_context.c osmesa_context.c)
 elseif (_GLFW_X11)
-    set(glfw_HEADERS ${common_HEADERS} x11_platform.h xkb_unicode.h posix_time.h
+    set(glfw_HEADERS ${common_HEADERS} x11_platform.h xkb_unicode.h unix_commons.h posix_time.h
                      posix_thread.h glx_context.h egl_context.h osmesa_context.h)
     set(glfw_SOURCES ${common_SOURCES} x11_init.c x11_monitor.c x11_window.c
                      xkb_unicode.c posix_time.c posix_thread.c glx_context.c
                      egl_context.c osmesa_context.c)
 elseif (_GLFW_WAYLAND)
-    set(glfw_HEADERS ${common_HEADERS} wl_platform.h
+    set(glfw_HEADERS ${common_HEADERS} wl_platform.h unix_commons.h
                      posix_time.h posix_thread.h xkb_unicode.h egl_context.h
                      osmesa_context.h)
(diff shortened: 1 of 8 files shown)


linkmauve:
I see you are using `ppoll()`, according to my manpages this is Linux-specific, so you may want to keep using `select()` like the rest of the code.

OlivierSohn:
@linkmauve do you have in mind a platform that has not this function ? I'm quoting @kovidgoyal here:

> As far as i can see, ppoll is available on linux and freebsd and openbsd and on netbsd under the name pollts, so it's probably fine to use it.

kovidgoyal:
Not to mention that since @linkmauve did not accept my code to not use timerfd in the wayland backend anyway, it's not like the code is portable currently in any case.

OlivierSohn:
@linkmauve I'm a bit reluctant to use select now, that would imply a lot of changes to the code (and make the code less "nice" I think). But if it is a must-have, I'll happily let someone add a commit to this PR to do the change.

Note that we could also use `poll` but it exhibits a coarser timeout granularity (1 millisecond vs 80 microseconds according to my benchmarks), this is the reason why I preferred using `ppoll`.

OlivierSohn:
This would need a rebase, I tried to give myself 5 minutes to fix the conflicts but it's a bit more complicated than I first thought...

PR #1307, opened 2018-07-29
Facilitate embedding GLFW objects into user static libraries

a3f:
Some projects (e.g. [raylib](https://github.com/raysan5/raylib/commit/b588af5ce22691bb143dcd784858e3c331c4a1e6#diff-f4de6a139c953f4ef27acdbc4f3d5d68), [nanogui](https://github.com/wjakob/glfw/commit/695b66d104e8a5ece8e6e55f0121ea24e9124d03#diff-95e351a3805a1dafa85bf20b81d086e6)) ship a static library that has GLFW
included (and reexports its symbols). Unfortunately, it's not possible to do so portably with the
current CMake setup.

[The proper  (CMake 2.8.8+) way](https://cmake.org/pipermail/cmake/2014-August/058235.html) seems to be to define a [CMake `OBJECT` library](https://cmake.org/cmake/help/v3.5/command/add_library.html#object-libraries) target and use
that as source for the `glfw` target.

- User code that is interested in the `glfw` target is unaffected and can
keep using it normally.
- User code that wants the objects to link into a static library just
needs to supply `$<TARGET_OBJECTS:glfw_objlib>` as part of the source.

After building the static library, it might be desirable to include it in the generated
pkg-config file. For that the other commit exports `GLFW_PKG_DEPS` and `GLFW_PKG_LIBS`
for use in client code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,10 +304,12 @@ endif()
 # Export GLFW library dependencies
 #--------------------------------------------------------------------
 foreach(arg ${glfw_PKG_DEPS})
-    set(GLFW_PKG_DEPS "${GLFW_PKG_DEPS} ${arg}")
+    set(GLFW_PKG_DEPS "${GLFW_PKG_DEPS} ${arg}" CACHE INTERNAL
+        "GLFW pkg-config Requires.private")
 endforeach()
 foreach(arg ${glfw_PKG_LIBS})
-    set(GLFW_PKG_LIBS "${GLFW_PKG_LIBS} ${arg}")
+    set(GLFW_PKG_LIBS "${GLFW_PKG_LIBS} ${arg}" CACHE INTERNAL
+        "GLFW pkg-config Libs.private")
 endforeach()
 
 #--------------------------------------------------------------------
--- a/README.md
+++ b/README.md
@@ -148,6 +148,8 @@ information on what to include when reporting a bug.
  - [NSGL] Removed enforcement of forward-compatible flag for core contexts
  - Export CMake `GLFW_PKG_DEPS` and `GLFW_PKG_LIBS` to parent scope for use
   in client pkg-configs (#1307)
+- Added a `glfw_objlib` CMake OBJECT library target for embedding into static
+  libraries (#1307)
 
 
 ## Contact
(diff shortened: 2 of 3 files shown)


a3f:
Oh, I should have tested this more throughly first...
I will fix them and push a new version.
Till then, I would appreciate feedback whether inclusion of this feature into upstream is desirable.

AndySomogyi:
How does this resolve the stb_image issue? 

GLFW includes  ```#inlclude stb_image.h```, and defines an implementation for these image functions by defining  ```STB_IMAGE_IMPLEMENTATION```. 

The issue I've ran into is other programs also include stb_image, and you end up with duplicate symbols. 

It would be nice is there was a clean way around this. 

I don't know, maybe a macro that renames the stb_image functions???

a3f:
> It would be nice is there was a clean way around this.

Well, the other programs could omit defining `STB_IMAGE_IMPLEMENTATON` if GLFW already provides it? Why would you want to keep two versions of `stb_image.h` functions?

a3f:
@elmindreda Did you have time to review whether you would want this PR?

AndySomogyi:
You most certainly would not want to keep two versions of ```stb_image.h```, but that's the problem with these header only libraries. 

Frequently, existing programs include ```str_image.h```, and use functions out of it, but when the link with glfw statically, you get duplicate symbols. 

That means existing programs need to change their implementation. 

Furthermore, what if an existing program uses a different version of ```stb_image.h``` than glfw, then in this case, you could end up with a different function implementing some stb_image function. 

I don't know a clean solution off hand, but this is a conversation we should really have to try to figure out a clean solution that does not break existing programs.

tombsar:
Forgive me if I'm misunderstanding, but what is the stb_image issue? I wasn't aware that GLFW did include `stb_image.h` (with or without implementations), and Github search isn't showing anything except in one of the example programs.

AndySomogyi:
Nevermind, it's all good. 

I was trying to statically link nanogui, and it itself include str_image, along with some example programs which also include str_image. I was concerned about the wrong thing. 

Note, this is why I'm leery about header-only libraries.

a3f:
Any news on this?

a3f:
I rebased again on top of master. Any feedback what's holding the merge of this pull request back?

elmindreda:
Thank you for this PR and for your patience.

The new optional CMake variable `GLFW_LIBRARY_TYPE` added with d83849792b0cf3e40c402f299c867ac6e1fd0e20 will let you make the `glfw` CMake target an object library. I believe this will do what you need to incorporate GLFW in a larger library. If not, that's something I want to fix.

I also added CMake cache variables with the pkgconfig information with 244948e174c0b698ac492d5aa8e379d5ccbc1323, similar to yours but with longer names.

PR #1316, opened 2018-08-24
Win32 support for GLFW_KEY_KP_EQUAL fixes #1315

### LINKED ISSUE #1315: Numpad equals key is unknown
My numpad equals key reports "Unknown" when pressed (or released)
The scancode  89 (0x059) is not present in the windows init source code (according to dougbinks). 

Original issue post: [http://discourse.glfw.org/t/numpad-equals-not-working/1157/2](url)

dougbinks:
Added support for GLFW_KEY_KP_EQUAL according to information on the relevant scancode from @RobertK92. Note that I do not have this key so can't test.


elmindreda:
> Note that I do not have this key so can't test.

I don't either but I when translate it to a VK it ends up in a small OEM reserved range, so 🤷 `thisisfine.gif`

Thank you!

PR #1346, opened 2018-10-04
Fix for macOS mojave not rendering the OpenGL view on window start

### LINKED ISSUE #1334: [macOS]GLFW can't draw any content automatically from beginning on macOS 10.14
These codes are most from glfw guide "Getting started". It works fine in macOS 10.13, but it draw nothing from beginning on macOS 10.14. If you move / resize the window manually, your content will write to screen finally.
[main.txt](https://github.com/glfw/glfw/files/2422731/main.txt)

nullcatalyst:
Fixes #1334 

Note that this just a simplified version of the commit kovidgoyal/kitty@b82e74f as mentioned by @kovidgoyal in #1334.

I really wanted there to be a more elegant solution than this, but this seemed to be the simplest fix I could come up with that still seemed to work.

I don't mind if this isn't used at all. Just wanted to provide an option for a potential solution.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -343,6 +343,10 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 - (void)windowDidBecomeKey:(NSNotification *)notification
 {
+    // Fix for macOS mojave not rendering the OpenGL view on window start
+    if (window->context.client != GLFW_NO_API)
+        [window->context.nsgl.object update];
+
     if (_glfw.ns.disabledCursorWindow == window)
         centerCursor(window);
 


kovidgoyal:
wouldn't this fail if the window is a background window?

yairchu:
I've tried this fix and on my machine (`MacBookPro13,2`) this had no effect - like before, the "Boing" example for example still wouldn't show anything until the window is moved.

nullcatalyst:
@yairchu I could reproduce what you were saying. Apparently I did all of my testing on naked executables (as in not .app files) where it was working fine.

I have now updated the proposed solution to work with all* of the examples.

I add an asterisk to all, because the _splitview_ example still does not render when the window is opened. But that is because it does not try to redraw the frame unless something has changed. And since the tiny snippet of code I added does not fix the fact that the view is not displayed until the third frame on my machine (`MacBookPro11,3`), it ends up not displaying anything.

yairchu:
Now it works 👍

nullcatalyst:
@kovidgoyal I didn't respond to your comment at first, because my original interpretation was that you were trying to render to a hidden window, to which my naive first thought was "does it matter if it doesn't render if it's not being displayed anywhere". Upon rereading it now, I think I now understand that you were talking about creating a window while the app is being run in the background. For which I suspect you are 100% right - the window wouldn't be made the "key" window and wouldn't get updated. My apologies for misunderstanding you.

That being said, I _believe_ the second commit I added _should_ also work in that scenario as well. But I have not tested that.

nullcatalyst:
What you're saying makes sense.

My personal assumption was that calling `update` needed to happen sometime after a specific event in the lifecycle of the window/view. I had assumed that that was why your original code needed to try it on a couple of frames, because the event in question hadn't happened yet. And that is why I was opting to find an event that happened as late as possible and just call update at that point. ie: call it when the window was made key, but since that didn't seem to work, just push a new event onto the dispatch queue, which should happen after all other prior events are dequeued.

nullcatalyst:
I currently think there might be another bug in play though, as Apple's [GLEssentials](https://developer.apple.com/library/archive/samplecode/GLEssentials/Introduction/Intro.html) example doesn't seem to have the 2-3 frame delay.

kovidgoyal:
yeah I doubt that calling update is the correct fix, just one that works, with relatively low overhead. The correct fix is going to need someone with more knowledge of cocoa internals.

nullcatalyst:
Lol, yeah, I'm not the guy for that either

elmindreda:
Please see if the `mojave-fixes` branch resolves this for you.

PR #1352, opened 2018-10-10
wayland: add support for XCURSOR_THEME and XCURSOR_SIZE

emersion:
These can be set by the compositor or the user to configure the
xcursor theme and size.

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -27,6 +27,8 @@
 #include "internal.h"
 
 #include <assert.h>
+#include <errno.h>
+#include <limits.h>
 #include <linux/input.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -958,6 +960,12 @@ static void createKeyTables(void)
 
 int _glfwPlatformInit(void)
 {
+    const char *cursorTheme;
+    const char *cursorSizeStr;
+    char *cursorSizeEnd;
+    long cursorSizeLong;
+    int cursorSize;
+
     _glfw.wl.cursor.handle = _glfw_dlopen("libwayland-cursor.so.0");
     if (!_glfw.wl.cursor.handle)
     {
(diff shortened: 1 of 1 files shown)


src/wl_init.c
```diff
@@ -1078,7 +1081,10 @@ int _glfwPlatformInit(void)
 
     if (_glfw.wl.pointer && _glfw.wl.shm)
     {
-        _glfw.wl.cursorTheme = wl_cursor_theme_load(NULL, 32, _glfw.wl.shm);
+        cursorTheme = getenv("XCURSOR_THEME");
+        cursorSize = getenv("XCURSOR_SIZE") ? atoi(getenv("XCURSOR_SIZE")) : 32;
```
linkmauve: There is a race condition here, please load the env var in a variable, and only then do the check for `NULL` and the conversion.

Also, please use `strtol()` instead of `atoi()` in order to check for errors.

src/wl_init.c
```diff
@@ -1078,7 +1081,10 @@ int _glfwPlatformInit(void)
...
+        _glfw.wl.cursorTheme =
+            wl_cursor_theme_load(cursorTheme, cursorSize, _glfw.wl.shm);
```
linkmauve: There is another `wl_cursor_theme_load()` a few lines below, you can use `cursorSize * 2` there for the size, as we only support outputs with `scale=1` and `scale=2`.

emersion:
Thanks for your review. Pushed a new version addressing your comments.

>Otherwise, are you sure it’s a good idea to start with this hacky way, while we could design a proper protocol and implement it in a few days?

I'm sure standardizing a new protocol will take at least months. I'm also not yet sure creating a whole new protocol is worth it. Last, I want to hear from GNOME and KDE to know if they're interested in such a protocol.

This solution has the upside of being very simple and it always works, regardless of compositor support. IMHO it's a good short-term solution, it's better than hardcoding these values.

linkmauve:
I agree.

PR #358, opened 2014-09-28
Clipboard

ricardomv:
Hi, i have the clipboard working but i have some issues i would like to understand better. First question is @jadahl do i really need to get the serial from all input to use it to set a selection?
And_glfwPlatformGetClipboardString _glfwPlatformSetClipboardString are not thread safe and i can't find anywhere in the documentation if they should be or not.
BTW @jadahl can you review the patch?

PR #1353, opened 2018-10-10
Implement clipboard support

linkmauve:
Clipboard support was one of the last missing features in the Wayland backend.

The current API is not adapted to a highly asynchronous protocol such as Wayland, but fixing that will have to wait 4.0 so I did what I could. The main issue is in `glfwGetClipboardString()`, which has to dispatch the request and block until the client sending the clipboard data is done, which could potentially take a very long time. A good way to reach this state is to copy from another Wayland client, `kill -STOP` it, and paste in the GLFW program, it will then block until we `kill -CONT` the other one. A possible solution against this issue would be to set a timer and place a handler for SIGALRM, but in the general case we can expect other clients to be well-behaved.

Supersedes #358, from which one commit has been extracted but the rest has been written from scratch.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -1569,11 +1569,126 @@ void _glfwPlatformSetCursor(_GLFWwindow* window, _GLFWcursor* cursor)
     }
 }
 
+static void dataSourceHandleTarget(void* data,
+                                   struct wl_data_source* dataSource,
+                                   const char* mimeType)
+{
+    if (_glfw.wl.dataSource != dataSource)
+    {
+        _glfwInputError(GLFW_PLATFORM_ERROR,
+                        "Wayland: Unknown clipboard data source");
+        return;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


src/wl_window.c
```diff
@@ -1565,24 +1565,195 @@ void _glfwPlatformSetCursor(_GLFWwindow* window, _GLFWcursor* cursor)
...
+        {
+            if (!growClipboardString())
+            {
+                close(fds[0]);
+                return NULL;
+            }
+        }
+
+        // Then read from the fd to the clipboard, handling all known errors.
+        ret = read(fds[0], _glfw.wl.clipboardString, 4096);
```
emersion: Shouldn't this be `_glfw.wl.clipboardString + len`?

linkmauve: Good catch!

linkmauve:
I’ll merge as is in order to get more testing done, please report if anything is going wrong.

PR #1356, opened 2018-10-13
Fix missing check for _NET_WM_STATE

cyclopsian:
I am using a tiling WM (XMonad) which doesn't support the NET_WM_STATE atom out-of-the-box, and whenver I call glfwGetWindowAttrib(window, GLFW_MAXIMIZED), I get a crash:

```                                                
X Error of failed request:  BadAtom (invalid Atom parameter)                                                                                                                                                                      
  Major opcode of failed request:  20 (X_GetProperty)                                                                                                                                                                               
  Atom id in failed request:  0x0                                                                                                                                                                                                 
  Serial number of failed request:  168                                                                                                                                                                                             
  Current serial number in output stream:  168
```

Other functions that get or set NET_WM_STATE check for the atom being null. The fix is to add a simple check to _glfwPlatformWindowMaximized.

### DIFF
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -2450,6 +2450,14 @@ int _glfwPlatformWindowMaximized(_GLFWwindow* window)
     Atom* states;
     unsigned long i;
     GLFWbool maximized = GLFW_FALSE;
+
+    if (!_glfw.x11.NET_WM_STATE ||
+        !_glfw.x11.NET_WM_STATE_MAXIMIZED_VERT ||
+        !_glfw.x11.NET_WM_STATE_MAXIMIZED_HORZ)
+    {
+        return maximized;
+    }
+
     const unsigned long count =
         _glfwGetWindowPropertyX11(window->x11.handle,
                                   _glfw.x11.NET_WM_STATE,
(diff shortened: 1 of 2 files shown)


src/x11_window.c
```diff
@@ -2450,6 +2450,10 @@ int _glfwPlatformWindowMaximized(_GLFWwindow* window)
     Atom* states;
     unsigned long i;
     GLFWbool maximized = GLFW_FALSE;
+
+    if (!_glfw.x11.NET_WM_STATE)
+      return maximized;
```
linkmauve: You need to use four spaces for indentation here.

```suggestion
        return maximized;
```

src/x11_window.c
```diff
@@ -2450,6 +2450,10 @@ int _glfwPlatformWindowMaximized(_GLFWwindow* window)
     Atom* states;
     unsigned long i;
     GLFWbool maximized = GLFW_FALSE;
+
+    if (!_glfw.x11.NET_WM_STATE)
```
elmindreda: This should also check for `_NET_WM_STATE_MAXIMIZED_VERT` and `_NET_WM_STATE_MAXIMIZED_HORZ`.

```suggestion
    if (!_glfw.x11.NET_WM_STATE ||
        !_glfw.x11.NET_WM_STATE_MAXIMIZED_VERT ||
        !_glfw.x11.NET_WM_STATE_MAXIMIZED_HORZ)
```

src/x11_window.c
cyclopsian: Good catch, I've just fixed this.

src/x11_window.c
cyclopsian: This has been corrected.

elmindreda:
Thank you!

PR #1362, opened 2018-10-19
Manage correctly the key combination Ctrl + Tab under macOS

Alzathar:
From qutebrowser/qutebrowser#2142 and specifically this [comment](https://github.com/qutebrowser/qutebrowser/issues/2142#issuecomment-263254133):
> The combination CTRL + Tab is specially handled by OSX/Cocoa text input method.

At least two solutions exists ([here](https://bugreports.qt.io/browse/QTBUG-8596?focusedCommentId=321526&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-321526) and [there](https://codereview.qt-project.org/#/c/161214/1/src/plugins/platforms/cocoa/qnsview.mm)) to fix the problem. However, the first one use a private/undocumented API. The second is used by this PR.

**NOTE**: I am not the author of the proposed code.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -643,6 +643,36 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
     [super updateTrackingAreas];
 }
 
+- (BOOL)performKeyEquivalent:(NSEvent *)nsevent
+{
+    // At the moment the only reason we define this method
+    // is Ctrl-tab key event never reaching any view at all.
+    // After -performKeyEquivalent: returns NO on all responders,
+    // we receve Ctrl-tab in our QNSWindow's -sendEvent:,
+    // but somehow it never reaches QNSView's -keyDown:.
+    // Apparently, it's treated in a special (and undocumented)
+    // way by Cocoa. 'Illegal' but really nice, clear and safe solution
+    // would be to define _wantsKeyDownForEvent, but it's a
+    // private/undocumented API.
+    if ([[self window] firstResponder] == self)
+    {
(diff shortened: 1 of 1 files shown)


intrigus:
Do you have permission to use the code if you are not the author? 
And is it compatible with glfw's license?

Alzathar:
@intrigus I've sent an email to the original author to answer to your questions.

tombsar:
It looks as though the code in this PR comes from QT, and is protected by a LGPL license. We can not simply copy the code without at minimum keeping the origin, authorship, and license intact. In my opinion (not a lawyer) we cannot include this code at all without explicit written permission from QT (and/or the original author) granting us an unrestricted license that does not carry with it the same responsibilities of the LGPL.

Are the comments included in the patch relevant to GLFW? I notice terms like `performKeyEquivalent` and `QNSView` that are unfamiliar, and I suspect may be referring to QT-specific code.

Alzathar:
@tombsar `performKeyEquivalent:` is proposed by Apple in its [AppKit API](https://developer.apple.com/documentation/appkit/nsresponder/1524690-performkeyequivalent?language=objc). This is a way to [handle keyboard shortcut](https://developer.apple.com/documentation/appkit/nsview/1483664-performkeyequivalent?language=objc). You can replace `QNSView` by `NSView` (available in [AppKit API](https://developer.apple.com/documentation/appkit/nsview?language=objc)) as the former inherits from the latter. I will modify the comment if the original author gives the permission to use its code.

PR #1364, opened 2018-10-22
The cursor position callback is named differently example usage and example declaration

tnixeu:
In the documentation the example for `glfwSetCursorPosCallback` uses a different name for the callback function than the example function blow. In all other examples the same name is used.


elmindreda:
Cleared the CI error. It was some miscommunication between GitHub and AppVeyor and nothing wrong with your patch.

PR #1369, opened 2018-11-01
add text input test/example

yairchu:
This is a simple test for text input, which can be useful when developing or trouble-shooting it.


elmindreda:
Thank you. Did you have a look at the existing `events` test program?

yairchu:
> Thank you. Did you have a look at the existing `events` test program?

Oh I didn't notice that my example is redundant by that one, will close this PR and use that program, thanks!

PR #1371, opened 2018-11-02
Use HTTPS when possible

MrVallentin:


docs/context.dox
```diff
@@ -255,7 +255,7 @@ of OpenGL ES extensions is identical except for the name of the extension header
 
 The `glext.h` extension header is a continually updated file that defines the
 interfaces for all OpenGL extensions.  The latest version of this can always be
-found at the [OpenGL Registry](http://www.opengl.org/registry/).  There are also
+found at the [OpenGL Registry](https://www.opengl.org/registry/).  There are also
```
linkmauve: This page redirects to https://www.khronos.org/registry/OpenGL/index_gl.php but it still may be the correct one, it’s probably better to keep your version.

docs/compile.dox
```diff
@@ -71,7 +71,7 @@ cmake -DCMAKE_TOOLCHAIN_FILE=CMake/x86_64-w64-mingw32.cmake .
 @endcode
 
 For more details see the article
-[CMake Cross Compiling](http://www.paraview.org/Wiki/CMake_Cross_Compiling) on
+[CMake Cross Compiling](https://www.paraview.org/Wiki/CMake_Cross_Compiling) on
```
linkmauve: This page says it has moved to https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/CrossCompiling so you should replace it.

docs/compile.dox
```diff
@@ -10,12 +10,12 @@ build applications that use GLFW, see @ref build_guide.
...
 @section compile_cmake Using CMake
 
-GLFW uses [CMake](http://www.cmake.org/) to generate project files or makefiles
+GLFW uses [CMake](https://www.cmake.org/) to generate project files or makefiles
 for a particular development environment.  If you are on a Unix-like system such
 as Linux or FreeBSD or have a package system like Fink, MacPorts, Cygwin or
 Homebrew, you can install its CMake package.  If not, you can download
 installers for Windows and macOS from the
-[CMake website](http://www.cmake.org/).
+[CMake website](https://www.cmake.org/).
```
linkmauve: Please remove the www. part here and above, to avoid one redirection too many.

linkmauve:
That’s my impression too. :)

Issue #1368, opened 2018-10-31
Caps lock on Mac has no key callback


AdamXA:
OS and version: macOS Mojave 10.14 (and happened on another Mac not on Mojave)
Release or commit: 3.2.1 stable
Error messages: None
Events output: None

I can detect every key on my keyboard apart from the caps lock with:
```void key_callback(GLFWwindow* win, int key, int scancode, int action, int mods) {```

This is set from:
```glfwSetKeyCallback(win, key_callback);```

The key_callback function isn't even called when the caps lock key is pressed on the keyboard.
Any ideas?

GreenLightning:
It is possible to detect when caps lock is turned on/off, so you would get a GLFW_PRESS event when the user presses the caps lock key and turns caps lock on and you would get a GLFW_RELEASE event when the user presses the caps lock key and turns caps lock off. See this [Technical Q&A from Apple](https://developer.apple.com/library/archive/qa/qa1519/_index.html) for reference.

However, treating the caps lock key as a regular key and getting events when the physical key is released seems to be more difficult. Apparently it is not possible to get this information from Cocoa events ([Stack Overflow 1](https://stackoverflow.com/questions/12536356/how-to-detect-key-up-or-key-release-for-the-capslock-key-in-os-x?rq=1)) and one would have to fall back to IOHIDManager or similar (if you want to go this route there is more information here: [Stack Overflow 2](https://stackoverflow.com/questions/7190852/using-iohidmanager-to-get-modifier-key-events)).

AdamXA:
Can confirm fixed in 3.3, thanks 👍

PR #1373, opened 2018-11-04
Send key events for caps lock on macOS

OneSadCookie:


elmindreda:
This is related to #1368.

PR #1374, opened 2018-11-05
Adds a `glfwInitVulkan` function to provide a vkGetInstanceProcAddr.

pmuetschard:
This allows an application to handle locating and loading a vulkan loader, but still use GLFW to handle the native window support.


Dominaezzz:
`glfwInitHint` instead?

pmuetschard:
I would like to close this out... after so much time :)

The `glfwInitHint` function takes in an `int` value, however, we'd need it to pass in a function pointer, which won't fit into an `int`. Should I 1) expand the value type of that function to also fit a function pointer, 2) add a `glfwInitHintFp` function ala `glfwWindowHintString`, 3) keep the `glfwInitVulkan` function?

PR #1381, opened 2018-11-15
Win32 CreateWindow performance improvement: Array based choose pixel format

dougbinks:
This change appears to improve, but not fix, the severe delay in glfwCreateWindow observed:

http://discourse.glfw.org/t/glfw-has-long-delay-when-creating-a-window/1072

I do not know how wide-spread this issue is, and as the fix adds complexity to the codebase I'm unsure whether it's a good idea to add.

If we do wish to go ahead, I can make any further required changes to this proposed pull request.


dougbinks:
It should be simple enough to add/remove the extension attributes before calling the query, and a linear search for the extension to find it's index in an array this size will indeed be very fast and remove the need to maintain the enum.

I'm on vacation at the moment but will take a look at these changes before too long.

dougbinks:
I've updated the PR branch to remove the enum and only check the ARB/EXT attributes when the relevant extensions are available.

This is somewhat more complex code, but it does still perform much faster than the individual checks - on the only machine I have available at the moment the check is 2x faster at 1ms versus more than 2ms. The hintPos could be removed for relatively little performance loss and some gain in code simplicity.

I am still unsure that the performance gain is worth the additional code complexity.

elmindreda:
Here are some results from my current laptop.

| | pixel formats | master (seconds) | patch (seconds)
| --- | --- | --- | ---
| HD 630 | 103 | 0.003 | 0.001
| GTX 1050M | 670 | 0.132 | 0.020

I tried simplifying the patch little further, in the style of the context creation functions, with 29661f10f5863672802b18501b2bccf275d56b2a. Does that seem sane?

dougbinks:
That looks very sane - I considered macros to simplify the code but wasn't sure about the style in GLFW.

The perf increase for NVIDIA GTX 1050M seems very reasonable, and better than I was seeing on the test machines I used (AMD and Intel), and the absolute time gain is perceptible.

elmindreda:
This was merged as babafc13dbdf6b031659ef54259185f72263e1d3. Thank you!

PR #1383, opened 2018-11-19
Disable non-client painting on borderless windows

aloucks:
Fixes an issue where a small title bar and window caption buttons were being painted after restoring a minimized undecorated window.

OS: Windows 10
GLFW: master branch

Steps to reproduce:

Repeatedly clicking the window's icon _on the task bar_ will cause the window to alternate between minimized and restored. Prior to this change, restoring the window causes the following title bar and caption buttons to be rendered in the upper left hand corner of the window.

image

```c
#include "GLFW/glfw3.h"

void key_callback(GLFWwindow* window, int key, int scancode, int action, int mods) {
    if (key == GLFW_KEY_ESCAPE) {
        glfwSetWindowShouldClose(window, GLFW_TRUE);
    }
}

int main() {
    glfwInit();
    glfwWindowHint(GLFW_DECORATED, GLFW_FALSE);
    GLFWwindow* window = glfwCreateWindow(800, 600, "GLFW Window", NULL, NULL);
    glfwSetKeyCallback(window, key_callback);
    while (!glfwWindowShouldClose(window)) {
        glfwWaitEvents();
    }
    glfwTerminate();
}
```

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -1161,6 +1161,19 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
             DragFinish(drop);
             return 0;
         }
+
+        case WM_NCACTIVATE:
+        case WM_NCPAINT:
+        {
+            // HACK: Prevent title bar artifacts from appearing after restoring
+            //       a minimized borderless window
+            if (!window->decorated)
+            {
+                return TRUE;
+            }
+
+            break;
+        }
     }
 
     return DefWindowProcW(hWnd, uMsg, wParam, lParam);


aloucks:
This also fixes some rendering issues when enabling the DWM drop shadow effect on a borderless window (which was the primary motivation; I want to draw my own window title bar). 

In order to draw the drop shadow, the `DWMNCRENDERINGPOLICY` must be set to `DWMNCRP_ENABLED` when the window style doesn't indicate that a border should be drawn. Discarding both the `WM_NCPAINT` and `WM_NCACTIVATE` messages have been the only way that I've found that prevents all of the window chrome from ever being painted. 

Example: 

```c
HWND hwnd = glfwGetWin32Window(window);
MARGINS margins = { 0 };
margins.cyTopHeight = 1;
DwmExtendFrameIntoClientArea(hWnd, &margins);
DWMNCRENDERINGPOLICY ncrp = DWMNCRP_ENABLED;
DwmSetWindowAttribute(hWnd, DWMWA_NCRENDERING_POLICY, &ncrp, sizeof(ncrp));
```

With this PR applied, the window will be fully transparent and the edges will have the drop shadow effect. 

Without this PR applied, a border will be displayed with any variations of the `MARGINS` variable and will have some amount of border drawn on the edges or corners. The borders can be painted over, but they flicker when the window is resized (note that resizing is now handled by the app, and not by the window proc).

elmindreda:
Thank you! This has been merged as 9ac9d7b85a224bb932f4a74912ddf5625a4eb110.

Issue #1209, opened 2018-02-10
input not received when switching between multiple windows


bsekura:
GLFW 3.2.1 on OSX.

I created two windows. Clicking between windows, the first click doesn't generate input event. After that, clicking in the same window, events come normally. Clicking on another window repeats the pattern (first click is lost).

GreenLightning:
This appears to be the default behavior of MacOS. From the [Cocoa Event Handling Guide](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/EventOverview/HandlingMouseEvents/HandlingMouseEvents.html):

> By default, a mouse-down event in a window that isn’t the key window simply brings the window forward and makes it key; the event isn’t sent to the NSView object over which the mouse click occurs. The NSView can claim an initial mouse-down event, however, by overriding acceptsFirstMouse: to return YES. The argument of this method is the mouse-down event that occurred in the non-key window, which the view object can examine to determine whether it wants to receive the mouse event and potentially become first responder. You want the default behavior of this method in, for example, a control that affects the selected object in a window. However, in certain cases it’s appropriate to override this behavior, such as for controls that should receive mouseDown: messages even when the window is inactive. Examples of controls that support this click-through behavior are the title-bar buttons of a window.

bsekura:
Thank you for finding this, makes sense.
I need this behavior in my case and in the meantime I can make changes myself. 

For the future, perhaps I could request an enhancement to allow the user to configure this behavior.

PR #1386, opened 2018-11-24
Trigger mouse button event when the window is not yet focused on Cocoa

AndrewBelt:
Solves https://github.com/glfw/glfw/issues/1209

This *could* have some debate, but this is the behavior on X11 and Win32. On Cocoa, when a window is not focused, the first click currently does not trigger a mouse button callback. This fixes that.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -497,6 +497,11 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
     updateCursorImage(window);
 }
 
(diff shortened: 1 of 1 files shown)


Luflosi:
I would like debate this. Is this the right place or is the forum better for that?

AndrewBelt:
Probably create a new issue or PR. Maybe propose a `GLFW_COCOA_UNFOCUSED_CLICK` window hint or something.

Issue #1400, opened 2018-12-13
Add an option to disable the use of raw mouse input


nathan818fr:
Hi,

I'm currently working on a mod for the Minecraft game. One of the important changes I have made is the upgrade from lwjgl 2 to lwjgl 3 (which upgrade glwf). Everything is really great, there is even a visible performance gain!

Unfortunately, even after 2 months some players still complain about the sensitivity of the mouse they find "strange".
Apparently the use of raw mouse input made the difference (https://github.com/glfw/glfw/commit/918b4e81d264a12ae5ad3c798ab34d9a2c53d2e3 / #125).
I totally agree that this change is excellent and consistent! But after years of playing some players can't play without the windows cursor movement improvments.

Would it be possible to add an option to force the use of WM_MOUSEMOVE instead of WM_INPUT?
By default, WM_INPUT would be used but the developer will be able to modify this behavior at runtime.

If I propose a pull request does it have a chance to be accepted?

Note: Many game have an option to enable/disable the use of raw inputs (Rainbow6, CS:GO, ...).

PR #1401, opened 2018-12-14
Allow to toggle the use of raw input per window

nathan818fr:
Following my issue #1400.

Features:
- Enable or disable the use raw input for mouse movements (when cursor is disabled) per window
  It's enabled by default and work only if the machine support it
  ~~`int glfwGetWindowUseRawInput(GLFWwindow* handle)`~~
  ~~`void glfwSetWindowUseRawInput(GLFWwindow* handle, int value)`~~
  it use `glfwGetInputMode` / `glfwSetInputMode` with the new mode `GLFW_RAW_INPUT`
- Check weather the machine support raw inputs
  `int glfwRawInputSupported(void)`


nathan818fr:
I like the idea of `glfwSetInputMode` with a new `GLFW_RAW_INPUT` mode !

nathan818fr:
So I used `glfwSetInputMode` and added doc (I tried to document properly).

include/GLFW/glfw3.h
```diff
@@ -3875,9 +3877,14 @@ GLFWAPI int glfwGetInputMode(GLFWwindow* window, int mode);
  *  GLFW_MOD_CAPS_LOCK bit set when the event was generated with Caps Lock on,
  *  and the @ref GLFW_MOD_NUM_LOCK bit when Num Lock was on.
  *
+ *  If the mode is `GLFW_RAW_INPUT`, the value must be either `GLFW_TRUE` to
+ *  enable the use of raw input, or `GLFW_FALSE` to disable it.  If enabled and
```
intrigus: You have a double space between "disable it.__If"

nathan818fr: Yes, I saw that everywhere else there were two spaces after dots, so I did the same.

nathan818fr:
After several days of use under Windows (win 10, x86_64) and Linux (Debian 9, x86_64), I can confirm that it works well!

nathan818fr:
I test this feature since a week on a Minecraft launcher (about 15 000 users per day), no bug report, everything goes well!

elmindreda:
Looks good. f48146ebaad0c0c7181e86679b0ad8118431599d in the [raw-mouse-motion-control](https://github.com/glfw/glfw/compare/raw-mouse-motion-control) branch are my proposed changes.

TL;DR from the summary:
> This renames 'raw input' to 'raw mouse motion' as there are other kinds of raw input.  The update path is restructured to avoid reinitializing all of disabled cursor mode.  Modification of shared state is moved out into shared code.  Raw mouse motion is disabled by default for compatibility.

nathan818fr:
Great!

PR #1406, opened 2018-12-19
wayland native: support for next-frame rendering using presentation-time

ewtoombs:
I added two functions to the wayland native API:
glfwWaylandPresSupported() checks the compositor for presentation-time
support. glfwWaylandSwapPres() swaps the buffers, then blocks
until the last-rendered frame is displayed on the screen, processing
events for the whole time. It returns the frame's presentation time with
respect to the glfwGetTime() clock.

I did some testing with weston and a test app. The app has a black background. It renders a white cursor at the location of every mouse move event seen so far. The closer the hardware cursor is to the rendered cursor, the less delay there is.

Here is Mesa's synchronisation with SwapInterval set to 1.
mesa

Here is presentation-time (SwapInterval has to be set to 0 for this method to work.)
pres

Here's the same thing with a 10ms delay added after the swap to wait for as many new mouse events as possible before starting the render.
pres-10ms-delay
The cursors are synchronised! They said it couldn't be done. Thus I have achieved next-frame rendering. I had to set weston's repaint-window to 3ms to get this last result.

### DIFF
--- a/include/GLFW/glfw3native.h
+++ b/include/GLFW/glfw3native.h
@@ -413,6 +413,31 @@ GLFWAPI struct wl_output* glfwGetWaylandMonitor(GLFWmonitor* monitor);
  *  @ingroup native
  */
 GLFWAPI struct wl_surface* glfwGetWaylandWindow(GLFWwindow* window);
+
+/*! @brief Check to see if the Wayland server supports the presentation-time
+ * protocol.
+ *
+ *  @return GLFW_TRUE iff presentation-time is supported.
+ *
+ *  @since Added in version 3.3.
+ *
+ *  @ingroup native
+ */
+GLFWAPI int glfwWaylandPresSupported(void);
+
+/*! @brief Swap the buffers, then use the Wayland presentation-time protocol to
+ * wait for the new frame to be presented. GLFW window events are processed for
+ * the whole time this function waits for the frame to be presented.
+ *
+ *  @return The presentation time of the frame, with respect to the same time
+ *  measure glfwGetTime() uses, or 0.0 if the frame was not shown due to a
+ *  vblank miss.
+ *
+ *  @since Added in version 3.3.
+ *
+ *  @ingroup native
+ */
+GLFWAPI double glfwWaylandSwapPres(GLFWwindow* window);
 #endif
 
 #if defined(GLFW_EXPOSE_NATIVE_EGL)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -55,6 +55,10 @@ elseif (_GLFW_WAYLAND)
         PROTOCOL
         "${WAYLAND_PROTOCOLS_PKGDATADIR}/unstable/idle-inhibit/idle-inhibit-unstable-v1.xml"
         BASENAME idle-inhibit-unstable-v1)
+    ecm_add_wayland_client_protocol(glfw_SOURCES
+        PROTOCOL
+        "${WAYLAND_PROTOCOLS_PKGDATADIR}/stable/presentation-time/presentation-time.xml"
+        BASENAME presentation-time)
 elseif (_GLFW_OSMESA)
     set(glfw_HEADERS ${common_HEADERS} null_platform.h null_joystick.h
                      posix_time.h posix_thread.h osmesa_context.h)
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -780,12 +780,28 @@ static const struct xdg_wm_base_listener wmBaseListener = {
     wmBaseHandlePing
 };
 
+static void
+presSetClockId(void *data, struct wp_presentation *pres, uint32_t clockId)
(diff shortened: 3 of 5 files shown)


ewtoombs:
My test code is here: https://gitlab.freedesktop.org/ewtoombs/opengl_glfw_latency_test

amonakov:
Nice! Can you comment to what degree this is implementable on other platforms?

OpenGL does not expose frame queue depth, so on Windows and X11 this is not straightforward, but nevertheless may be possible. For example, Croteam's Serious Engine can utilize the "presentation" part of VDPAU to keep frame visibility delays under control, iiuc. Other avenues are `OML_sync_control` with Mesa drivers and `NV_delay_before_swap` on Nvidia, I think.

ewtoombs:
> Nice! Can you comment to what degree this is implementable on other platforms?

I wish I could, @amonakov. I am pretty new to this, so wayland is the only one I have really looked into in-depth.

> OpenGL does not expose frame queue depth, so on Windows and X11 this is not straightforward, but nevertheless may be possible. For example, Croteam's Serious Engine can utilize the "presentation" part of VDPAU to keep frame visibility delays under control, iiuc. Other avenues are `OML_sync_control` with Mesa drivers and `NV_delay_before_swap` on Nvidia, I think.

I see the `GLX_OML_sync_control` extension in my glxinfo! I can test that one. The others, I don't think I can. I am hoping somebody with the hardware can test them. There are also a whole bunch of other GLX extensions with the words swap or sync in them that are supported by my card:

```
server glx extensions:
    GLX_INTEL_swap_event, GLX_OML_swap_method, GLX_SGI_swap_control
client glx extensions:
    GLX_INTEL_swap_event, GLX_MESA_swap_control, GLX_OML_swap_method, GLX_OML_sync_control, GLX_SGI_swap_control, GLX_SGI_video_sync
```

vans163:
This works well on wayland, tested with weston, really good addition.

ninja-:
what's the status of this? :)

ewtoombs:
It was working a year ago, when I had written it. Now, I don't know. As a part of wayland native support, I thought it could have been added right after I had written it.

ninja-:
@ewtoombs I am currently taking care of wayland fixes so I might be interested in taking over this one as well :thinking: 

Do you think this should be an opt-in or always sync render time this way?
Does this improve smoothness of v-sync running games or there wouldn't be much difference in this case?

ewtoombs:
It has to be opt-in, because SwapInterval() must be set to 0 in order for it to work.

ewtoombs:
It is not a smoothness fix, but a latency fix, capable of reducing the mouse to screen latency by around a whole frame. Smoothness will be unaffected if all you do is adopt presentation-time. If you also add the 10ms rendering delay after swap, which you have to in order to get next-frame rendering, you lose the same number of milliseconds from your rendering budget, so that could actually make smoothness much worse, if the game requires the entire frame's rendering budget for worst cast renders.

ninja-:
> It is not a smoothness fix, but a latency fix, capable of reducing the mouse to screen latency by around a whole frame. Smoothness will be unaffected if all you do is adopt presentation-time. If you also add the 10ms rendering delay after swap, which you have to in order to get next-frame rendering, you lose the same number of milliseconds from your rendering budget, so that could actually make smoothness much worse, if the game requires the entire frame's rendering budget for worst cast renders.

Typical budget is 1000/60 = 16.66ms, so losing 10ms sounds like quite a lot, especially for games?
What apps do you think could benefit from this change and which not?
Input lag is of course important to games, but not at the cost od 2/3 rendering time.

ewtoombs:
Only AAA games use all 17ms of their rendering budget. The vast majority of regular games could render in 1 or 2 ms without a problem. Such games could work just fine with next frame scheduling. Emulators in particular benefit enormously. Super Mario World rendered next-frame is decidedly twitchy. It's awesome. But yes, this won't work for all games.

tinne26:
I recently came across the same problem and luckily found this discussion. There are a couple parts that puzzle me:
- OSes do this properly for cursors, but I assume they are not breaking vsync for that? What do they know that we don't?
- I came from stackoverflow 19102189, where you mention that SwapInterval(1) adds an extra 1 frame delay. Why isn't there a 0 version that doesn't break vsync, and why didn't they leave some other value to vsync off? Is that what OSes may be doing? I doubt something so obvious isn't allowed unless there's a good reason for it.
- What are first person shooters and similar games doing then? Leaving players to disable vsync, or simply running away from glfw? If so, where do they run to and what are other libraries doing?

I also noticed that creating sleeps such that the system runs just slightly below it's vsync FPS (e.g: 59.67 vs 60.00) makes the cursor much better synchronized too, even under SwapInterval(1). I guess this is effectively another way to break the synchronization, presumably forcing glfw to use the current frame it's given when swapping instead of being allowed to keep it until the next frame's swap.

PR #1407, opened 2018-12-22
Wayland: use SHM_ANON on FreeBSD (like memfd on Linux)

myfreeweb:
Reference: [shm_open(2)](https://www.freebsd.org/cgi/man.cgi?shm_open)  
Similar patches: [GTK](https://gitlab.gnome.org/GNOME/gtk/merge_requests/203), [Firefox](https://hg.mozilla.org/mozilla-central/rev/83bab8cf29bf)


linkmauve:
LGTM, and sorry for not noticing this patch earlier!

PR #1416, opened 2019-01-11
CMake: enable policy CMP0077.

0xb8:
This policy makes it easier for users to set library option()s when used
as CMake subproject.


elmindreda:
Apologies. This was fixed with 71e6ff386d9a3e0cc7c069be9b0ee1a9ac738647 before I reviewed this PR.

Issue #1337, opened 2018-09-29
Vsync doesn't work on macOS 10.14 (Mojave)


hajimehoshi:
I checked example `boing` (built as `boing.app`) and it didn't sync with display refresh rate even with `glfwSwapInterval(1)` and `glfwSwapBuffer`. I'm not sure whether it worked on 10.13, but I think so.

ghuser404:
It may depend on whether your GPU driver is forcing this.

hajimehoshi:
Hi, I'm using MacBook Pro 2018, and I think vsync is enforced.

hajimehoshi:
My graphics card info

```
Intel Iris Plus Graphics 655:

  Chipset Model:	Intel Iris Plus Graphics 655
  Type:	GPU
  Bus:	Built-In
  VRAM (Dynamic, Max):	1536 MB
  Vendor:	Intel
  Device ID:	0x3ea5
  Revision ID:	0x0001
  Metal:	Supported, feature set macOS GPUFamily2 v1
  Displays:
Color LCD:
  Display Type:	Built-In Retina LCD
  Resolution:	2560 x 1600 Retina
  Framebuffer Depth:	24-Bit Color (ARGB8888)
  Main Display:	Yes
  Mirror:	Off
  Online:	Yes
  Rotation:	Supported
  Automatically Adjust Brightness:	No
```

ghuser404:
I don't know much about macOS, but in Windows you can enforce vsync on/off, so that the GLFW setting won't do anything.

kivutar:
I can confirm this is a regression due to 10.14. On the previous version of OSX, vsync used to work.

mokafolio:
Can confirm. This is new on mojave and does only happen when using my dedicated GPU, not the integrated one.

gimblll:
Still broken with the 10.14.2 update. Anyone have any workarounds?

gimblll:
Answering myself: the workaround I got working now was compiling with XCode 9 instead of 10. Seems SDL has come to the same conclusion https://discourse.libsdl.org/t/sdl2-no-renderer-vsync-on-macos-mojave/25169/21

rcgordon:
Heads up, I just got a solution into SDL for this problem, if you want to adapt it for GLFW:

https://hg.libsdl.org/SDL/rev/73f3ca85ac0e

gerryhernandez:
Great work, and you saved my bacon. Thank you!

amarcu5:
[...]

PR #1417, opened 2019-01-14
Fix for VSync on 10.14 Mojave

amarcu5:
Addresses issue #1337

Fixes VSync on macOS 10.14 (Mojave) by using CVDisplayLink to synchronize frames rather than setting NSOpenGLContextParameterSwapInterval.

Based on work for SDL by @rcgordon: https://hg.libsdl.org/SDL/rev/73f3ca85ac0e

### DIFF
--- a/src/nsgl_context.h
+++ b/src/nsgl_context.h
@@ -27,13 +27,21 @@
 #define _GLFW_PLATFORM_CONTEXT_STATE            _GLFWcontextNSGL nsgl
 #define _GLFW_PLATFORM_LIBRARY_CONTEXT_STATE    _GLFWlibraryNSGL nsgl
 
+#import <CoreVideo/CoreVideo.h>
+
+#include <stdatomic.h>
+
 
 // NSGL-specific per-context data
 //
 typedef struct _GLFWcontextNSGL
 {
-    id           pixelFormat;
-    id	         object;
+    id                pixelFormat;
+    id	              object;
+    CVDisplayLinkRef  displayLink;
+    atomic_int        swapInterval;
+    int               swapIntervalsPassed;
+    id                swapIntervalCond;
 
 } _GLFWcontextNSGL;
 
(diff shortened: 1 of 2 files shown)


GreenLightning:
My initial testing suggests that this fixes #680 as well!

kovidgoyal:
This PR is not really correct. Various problems:

1) Race condition between CVLinkDisplayThread and main thread on window close. This can lead to crashes, such as https://github.com/kovidgoyal/kitty/issues/1393

2) This will create a new thread for every window, which is a waste of resources. CVDisplayLink is a simple timer, there should be one per monitor, not one per window.

3) Different monitors can have different refresh rates, CVDisplayLinkSetCurrentCGDisplayFromOpenGLContext simply sets the rate to the slowest connected monitor, which is sub-optimal https://developer.apple.com/documentation/corevideo/1457164-cvdisplaylinksetcurrentcgdisplay?language=objc

In my glfw fork I abandoned setInterval completely and instead used a render frame callback, so applications can choose to do rendering work as and when needed, reducinf CPU consumption. This also works on wayland which has a render frame protocol.

elmindreda:
@kovidgoyal See above for fixes of issues 1 and 3 and related bugs.

kovidgoyal:
As far as I can tell, the race still exists. Imagine the display link thread gets stopped in the middle of displayLinkCallback and the window gets destroyed  in the main thread. Then the display link thread is resumed, leading to a crash. This might be fixable by stopping the display link thread at the begininng of destroying the window, but that is assuming that stopping the thread actually stops it even if it is in the middle of a callback, and not after the callback completes.

elmindreda:
I was hoping display link was sane enough to join on its worker thread at stop or release but will look into whether this is actually done.

kovidgoyal:
Since the documentation does not talk about the threading semantics of display link at all, I would not be comfortable relying on it. In my code, I use a global lock in _glfw.ns to ensure there are no races between monitor removal/display link.

PR #1418, opened 2019-01-15
Fixed a little bug in CrossProduct function.

maobaoqi:
Hi, I read boing.c and find a little bug about cross product alogrithm. Although it doesn't matter for rendering result, I want to fix it. Thanks!^_^


elmindreda:
Thank you!

PR #991, opened 2017-04-08
Support for resize operation on borderless windows

felipefs:
This patch adds the functionality to resize a borderless window, by specifying the region relative to the resize operation:

- ```GLFW_WINDOW_LEFT```
- ```GLFW_WINDOW_TOP```
- ```GLFW_WINDOW_RIGHT```
- ```GLFW_WINDOW_BOTTOM```
- ```GLFW_WINDOW_TOPLEFT```
- ```GLFW_WINDOW_TOPRIGHT```
- ```GLFW_WINDOW_BOTTOMLEFT```
- ```GLFW_WINDOW_BOTTOMRIGHT```

This is related with the issue https://github.com/glfw/glfw/issues/990.

Issue #990, opened 2017-04-07
Support for resizing undecorated window


felipefs:
I'm opening this issue for myself, related to https://github.com/glfw/glfw/issues/923.

The same way it's possible to start a drag operation of an undecorated window, it's also possible to start a resize operation. This feature might have some particularities in some platforms, as Wayland is still in development and Cocoa doesn't allow some of these operations (or they are slightly different).

felipefs:
I made the X11 implementation on my fork: https://github.com/felipefs/glfw/tree/window-resize

Here's the internal X11 implementation:
```c
void _glfwPlatformResizeWindow(_GLFWwindow* window, int border)
{
    int winXpos, winYpos;
    double curXpos, curYpos;
    XClientMessageEvent xclient;
    memset(&xclient, 0, sizeof(XClientMessageEvent));
    XUngrabPointer(_glfw.x11.display, 0);
    XFlush(_glfw.x11.display);
    _glfwPlatformGetCursorPos(window, &curXpos, &curYpos);
    _glfwPlatformGetWindowPos(window, &winXpos, &winYpos);
    xclient.type = ClientMessage;
    xclient.window = window->x11.handle;
    xclient.message_type = XInternAtom(_glfw.x11.display, "_NET_WM_MOVERESIZE", False);
    xclient.format = 32;
    xclient.data.l[0] = winXpos + curXpos;
    xclient.data.l[1] = winYpos + curYpos;
    switch (border)
    {
        case GLFW_WINDOW_LEFT:
            xclient.data.l[2] = _NET_WM_MOVERESIZE_SIZE_LEFT;
            break;
        case GLFW_WINDOW_TOP:
            xclient.data.l[2] = _NET_WM_MOVERESIZE_SIZE_TOP;
            break;
        case GLFW_WINDOW_RIGHT:
            xclient.data.l[2] = _NET_WM_MOVERESIZE_SIZE_RIGHT;
            break;
        case GLFW_WINDOW_BOTTOM:
            xclient.data.l[2] = _NET_WM_MOVERESIZE_SIZE_BOTTOM;
            break;
        case GLFW_WINDOW_TOPLEFT:
            xclient.data.l[2] = _NET_WM_MOVERESIZE_SIZE_TOPLEFT;
            break;
        case GLFW_WINDOW_TOPRIGHT:
            xclient.data.l[2] = _NET_WM_MOVERESIZE_SIZE_TOPRIGHT;
            break;
[...]

Issue #923, opened 2016-12-21
Provide a way to spcify the window drag area on an undecorrated window


adambrown:
I am doing exactly what @Andos is doing here: #902. Creating windows with completely re-purposed title bar area. The current trend on Windows 10, and possibly some other platforms as well given that @Andos is doing it on OSX, seems to be that the title bar is dead (Firefox, Chrome, Edge, Photoshop, Visual Studio, and a multitude of others, too numerous to name or remember off the top of my head, seem to agree). This space can be better used by moving up your menu, toolbar, tabs or whatever you have right below the title bar in order to maximize usable space.

All of these examples given above provide some portion of the former title bar as a drag area for moving the window. It would be nice if GLFW provided a way to specify a dynamically resizable area, probably a rectangle, within an un-decorated window to use as a drag handle, to provide native window dragging functionality without a title bar.

In #902 @Andos gives an example of how this can be hacked around OSX with `performWindowDragWithEvent`. In windows this is done with `Window.DragMove`. Combined with the native resizing functionality in #902 this will allow fully functional undecorated windows like those used by Firefox, Chrome, Photoshop etc...

richtw1:
In my opinion, this is better achieved in Windows by catching the [WM_NCHITTEST](https://msdn.microsoft.com/en-us/library/windows/desktop/ms645618(v=vs.85).aspx) message. I would love a callback which can return some kind of 'hit code' based on the screen coordinates relative to the upper-left of the client area.  Not sure if X11/Cocoa have a similar kind of functionality.

As others have noted, the current vogue is to abandon the native title bar and/or window borders in favour of rendering them yourself, and a callback system like this would be far better than the alternative laggy method of calculating cursor pos deltas and moving the window accordingly in the event loop.

felipefs:
Instead of defining an area for dragging the window, it would be more flexible to just add a function to start the drag operation (example, ```glfwWindowDrag```).

X11 (and probably Wayland) has this functionality, too. Long ago I did this with X11, so I can implement for X11 if no one is working on this yet.

felipefs:
I started a implementation on X11 (Here's my commit on my fork: https://github.com/felipefs/glfw/commit/3d3a94b4be6fb96faa0c7e22b6d59c296677ec6d).

It adds the function ```glfwDragWindow(glfwWindow *window)```. Here are the most important internal part of the implementation:

```c
/* window.c */
GLFWAPI void glfwDragWindow(GLFWwindow* handle)
{
[...]

PR #1420, opened 2019-01-19
Support for drag and resize operations

RobbinMarcus:
This is a cleaned up version that merges both changes for the drag operation and resize operation. I implemented the requested changes by @jadahl and @linkmauve and solved conflicts generated by the old pull requests.

Merged these two pull requests together:
- Add support for drag operation on borderless window #987
- Support for resize operation on borderless windows #991

Affected issues:
- Provide a way to specify the window drag area on an undecorrated window #923
- Support for resizing undecorated window #990

Current state:
I tested these changes on windows 10 (win32). I changed the "window" example locally to have an undecorated window and added a buttons to test the drag and resize operations (glfwDragWindow or glfwResizeWindow).

Specifically for win32 this PR implements the `WM_NCHITTEST` and `WM_NCCALCSIZE` to get rid of the 7px white header above the window. These changes led me to add 2 functions to customize the caption area and the resize border size. Currently this information is available for every platform, but only used for win32. 

UndecoratedResize

Reviewers: please let me know if this needs to be split up into separate PRs. The drag/resize operations are mostly separate from the win32 specific undecorated window changes. 

Testing: This PR is only tested in windows 10 at the moment. The other platforms are in the same state as the previous PRs mentioned above, except for the changes I made to get them up to date to the latest master branch.


xteo:
@RobbinMarcus : Thanks for this MR
I am trying to setup our Application ( using GLFW) to be without OS Title bar so we can control the Look of the Titlebar fully like other moderm application 

I will try you MR this week but wanted to ask a question. in your note you mentioned you still have issue with Resize ? did I understand well ? 

** Undecorated glfwResizeWindow does not work, I cannot figure out why.**

Thanks for your contribution hope to get it working!

RobbinMarcus:
Hi @xteo. 
Yes that's correct, trying to resize the window does not work in this PR. 
I've tried to fiddle with some things but I couldn't get it to work. Good luck!

rokups:
This feature would be very helpful to Dear ImGui, so i was testing this functionality a bit. X11 implementation has an issue where drag operation makes mouse button stuck in pressed state even after drag operation is done and mouse button is released (which happens over window by the way). If anyone has any ideas how to work this around - please speak 🙏🏻

emoon:
It would be great to get this feature in. It's something I want myself as well

rokups:
I spent some time ironing it out. It works quite good on X11 now, but in the process i had to also fix inconsistent mouse button reporting. On windows however resizing seems to not work at all. I am not sure why, but maybe application has to support handling these resize messages..?

JasonTheKitten:
I believe this PR would be extremely useful.
Many times, one wishes to use an undecorated window. Perhaps they want their custom window decorations to have a different aesthetic. Or perhaps they want to add extra widgets to the decoration.
While it is possible (and easy) to manually write code to drag an undecorated window, it may not encompass all of the features of the default drag behavior of the OS's windowing system.
For example, on Windows, the default windowing system will allow you to maximize windows if you drag them to the top, or do a side-by-side view if you drag them to the sides.

This is why I would like this PR to be merged.

Cons-Cat:
I would certainly use this!

CybernetHacker14:
For the issue - "Undecorated glfwResizeWindow does not work, I cannot figure out why" - can this solution work - https://github.com/melak47/BorderlessWindow ? I have tested it out, and it seems to be exactly equivalent to what the resultant output should turn out to be.

RobbinMarcus:
> For the issue - "Undecorated glfwResizeWindow does not work, I cannot figure out why" - can this solution work - https://github.com/melak47/BorderlessWindow ? I have tested it out, and it seems to be exactly equivalent to what the resultant output should turn out to be.

I've added `WS_THICKFRAME` to the window style which seems to do the trick to enable window resizing while undecorated.

There's a 'beauty problem' where you will get the following result with this small white frame:
Ryne_589utCQFhf

It seems this is expected behavior though: https://stackoverflow.com/questions/39731497/create-window-without-titlebar-with-resizable-border-and-without-bogus-6px-whit
Removing the top bar requires additional work to fix and be compatible with all windows versions.

CybernetHacker14:
The problem now is that, with the `WS_THICKFRAME` option, dragging the window seems to not work, cause when the cursor is near the border, the resize prompt is thrown, instead of dragging. I am testing this on Windows, comparing it with the example repo which I have posted in my earlier comment.

CybernetHacker14:
- This is my specific functionality just for the windows - https://github.com/CybernetHacker14/glfw/tree/win32-borderless . I have 
implemented the `WM_NCHITTEST` and `WM_NCCALCSIZE` functionality.
- This is extremely platform specific, with code taken up from sections of - https://github.com/melak47/BorderlessWindow .

RobbinMarcus:
Thanks for the input @CybernetHacker14. I took up a part of the code to implement the win32 `WM_NCHITTEST` and `WM_NCCALCSIZE`. The resulting window looks pretty good! 

I added customizability options to set a rectangle area for the caption (so you can have some menu buttons where the window won't be dragged) and the border size where you can start the window resize. 

The defaults for these are as follows:
- Caption Area will be offset by the border to start at (0, 0) in the client area. It's size will be full-width and 16px in height. 
- The resize border size is 4px wide for all borders.

I have also added `WS_MAXIMIZEBOX` back in the style which allows the window to be maximized when you drag it to the top of the screen for example.

In the example below I set the caption area to be the top right side, so pressing in the top left of the window won't start the drag operation. This is in addition to the already implemented callable drag/resize functions which still work the same.

UndecoratedResize

CybernetHacker14:
Looks pretty good. Check whether you can double click to maximize and minimize the windows. Also, see how the window behaves when you drag-dock it to different sides of the screen.

CybernetHacker14:
> Thanks for the input @CybernetHacker14. I took up a part of the code to implement the win32 `WM_NCHITTEST` and `WM_NCCALCSIZE`. The resulting window looks pretty good!
> 
> I added customizability options to set a rectangle area for the caption (so you can have some menu buttons where the window won't be dragged) and the border size where you can start the window resize.
> 
> The defaults for these are as follows:
> 
> * Caption Area will be offset by the border to start at (0, 0) in the client area. It's size will be full-width and 16px in height.
> * The resize border size is 4px wide for all borders.
> 
> I have also added `WS_MAXIMIZEBOX` back in the style which allows the window to be maximized when you drag it to the top of the screen for example.
> 
> In the example below I set the caption area to be the top right side, so pressing in the top left of the window won't start the drag operation. This is in addition to the already implemented callable drag/resize functions which still work the same.
> 
> UndecoratedResize [ UndecoratedResize ](https://user-images.githubusercontent.com/4852449/192385172-073f6556-dcc0-4888-aa36-68c9866c1047.gif) [ ](https://user-images.githubusercontent.com/4852449/192385172-073f6556-dcc0-4888-aa36-68c9866c1047.gif)

Does this exclude the area where the "Test no dragging" button is present, or there is some callback type thing, where you are checking if any button is hovered or not.

RobbinMarcus:
> Looks pretty good. Check whether you can double click to maximize and minimize the windows. Also, see how the window behaves when you drag-dock it to different sides of the screen.

All the normal windows 10 functionality now works similar to the decorated version as far as I can tell. 

> Does this exclude the area where the "Test no dragging" button is present, or there is some callback type thing, where you are checking if any button is hovered or not.

I'm explicitly not checking any window contents as this will be very application specific. Without the caption area I implemented the window would be dragged if you pressed on the "Test no drag" button there. The `WM_NCHITTEST` overrides any other application behavior by returning that it hit the caption.

Scr3amer:
Hello !

I wanted to play with your branch.
When I compile with clang 14 I have a warning:

win32_window.c:1896:3: warning: variable 'wBorder' is used uninitialized whenever switch default is taken

Should we early return ?
Or set wBorder to HTBORDER ?
Or any other option ?



PS:
I was trying to do exactly that but just using GLFW API but no luck. (It worked for the bottom and right side but resizing and repositioning the window at the same time was buggy for left and top)

Scr3amer:
> I've added an early return for this, hope this works. If this default case hits, it will trigger the assert as this should not happen, but adding the return should make it compile for clang.

So we don't want to always call ReleaseCapture ? I am not very familiar with Win32 API (even though I read the doc of that function)

> I did also notice this, but I don't think there is a use case for continually swapping between resizable or not. I'll have to look into this though, probably have to shrink the window by the border size whenever we go back to undecorated.

I agree with you, it should be set only once. It was more for the polish than the functional aspect of it.

> Do you mean you were trying to replicate my example from above and couldn't resize or move? Did you use the new functions to increase the caption size and resize border:

Sorry I was not very clear in what I said. I meant initially I tried to implement the feature of this PR using only the GLFW API with glfwSetWindowSize followed by glfwSetWindowPos but it didn't work for left and top borders. So I went on Github and found this PR :). That's all.

Scr3amer:
Me again.

It works fine on my PC (Win10 + Clang14 UCRT MinGW)

I just noticed one last thing, when we resize ( or drag) it pauses the event loop meaning the content doesn't scale dynamically. It is paused  and then refreshed only when we release the mouse.

While when we resize Firefox, Chrome, VS Code etc.... The content is refreshed and updated while we resize / drag the window.

Scr3amer:
Hello hello !

Me again, just adding some things I noticed while I use it.

Overall It works like a charm on my PC (Win10 + Clang14 UCRT MinGW)

3 things I noticed though:

- the small detail of the size increase on toggle of the decoration I mentionned earlier
- why not setting GLFW_INVALID_ENUM for glfwResizeWindow if border is wrong ?
**-  when we drag the window to the top it maximizes but then dragging it down doesn't minimize it again. Feels like a missing feature, do you agree ?**

Cheers

AlfredXiaSF:
> Hello hello !
> 
> Me again, just adding some things I noticed while I use it.
> 
> Overall It works like a charm on my PC (Win10 + Clang14 UCRT MinGW)
> 
> 3 things I noticed though:
> 
> * the small detail of the size increase on toggle of the decoration I mentionned earlier
> * why not setting GLFW_INVALID_ENUM for glfwResizeWindow if border is wrong ?
> * **sometimes dragging the top border of a maximized window doesn't "register"** and the window is not minimized. Especially if the mouse cursor is at y = 0 but it can also happen within the border.
> 
> Cheers

Hello, @Scr3amer 
> sometimes dragging the top border of a maximized window doesn't "register" and the window is not minimized. Especially if the mouse cursor is at y = 0 but it can also happen within the border.
I try to make some changes in `HitTest`, seems fix this. The key is, if Maximized set `clientAreaTop` to zero.

```cpp
static int HitTest(_GLFWwindow* window, RECT windowRect, POINT cursor, POINT border)
{
    const int captionOffsetX = window->captionOffsetX != GLFW_DONT_CARE ? window->captionOffsetX : 0;
    const int captionOffsetY = window->captionOffsetY != GLFW_DONT_CARE ? window->captionOffsetY : 0;
    const int captionSizeX = window->captionSizeX != GLFW_DONT_CARE ? window->captionSizeX : windowRect.right - windowRect.left;
    const int captionSizeY = window->captionSizeY;

    const int clientAreaLeft = windowRect.left + border.x;
    const int clientAreaRight = windowRect.right - border.x;
    const int clientAreaTop = windowRect.top + (_glfw.platform.windowMaximized(window) ? 0 : border.y);
    const int clientAreaBottom = windowRect.bottom - border.y;

    const int cursorInCaption =
        cursor.x >= windowRect.left + captionOffsetX &&
        cursor.x <= windowRect.left + captionOffsetX + captionSizeX &&
        cursor.y >= windowRect.top + captionOffsetY &&
        cursor.y <= windowRect.top + captionOffsetY + captionSizeY;
```
---
 Case I only want to set CaptionArea height, so I set `offsetX` `offsetY` and `sizeX` to `GLFW_DONT_CARE`, but it not work, so I change `glfwSetWindowCaptionArea`
```cpp
GLFWAPI void glfwSetWindowCaptionArea(GLFWwindow* handle, int offsetX, int offsetY, int sizeX, int sizeY)
{
    _GLFWwindow* window = (_GLFWwindow*)handle;
    assert(window != NULL);

    _GLFW_REQUIRE_INIT();

    if (offsetX != GLFW_DONT_CARE && offsetX < 0)
        return;
    if (offsetY != GLFW_DONT_CARE && offsetY < 0)
        return;
    if (sizeX != GLFW_DONT_CARE && sizeX < 1)
        return;
    if (sizeY < 1)
        return;

    window->captionOffsetX = offsetX;
    window->captionOffsetY = offsetY;
    window->captionSizeX = sizeX;
    window->captionSizeY = sizeY;
}
```

sasichkamega:
why not merged?

nulldoubt:
When will this be merged? Anything holding us back? 🤔🤔

RobbinMarcus:
There has been no comment from reviewers on this pull request for 6 years now. 
From their perspective I could see maintenance would increase if they would support this option: if it doesn't work on any of the platforms they would have to fix it. 
It could be they have decided there is no place for this functionality in GLFW and/or the amount of people that would use it is too small (though I would appreciate a reply with that information so we can close it). 

I haven't taken the time to fix the conflicts for this PR with the latest release of GLFW, because I think it would be a waste of time if this issue would be in conflict again after a couple changes and then remain like this for another 6 years!

Give this message a thumbs up if you use or would use this functionality. I'll update this PR to resolve the conflicts with the latest if I get over 100 on this comment.

alexv-ds:
@RobbinMarcus ✊

constantitus:
I'm a bit confused.
Why does `resizeBorderSize` only affect the windows implementation ? I thought that since I have to manage borders myself on the other platforms, I'll just set it to zero to disable it, but I can't do that either ?
```c
GLFWAPI void glfwSetWindowResizeBorderSize(GLFWwindow* handle, int size)
{
...
    if (size < 1)
        return;
```
Same with the captionArea, it's only checked on windows and you can't fully disable it.
```c
GLFWAPI void glfwSetWindowCaptionArea(GLFWwindow* handle, int offsetX, int offsetY, int sizeX, int sizeY)
{
...
    if (offsetX < 0 || offsetY < 0 || sizeX < 1 || sizeY < 1)
        return;
```
Why ?

RobbinMarcus:
Back when I implemented this I only ran this on Windows, not taking into account any of the other platforms except the work done by previous PRs. 

I don't see much traction for this PR to be merged so I'm going to close it. Feel free to use any or all code from this PR if you want to split it up into separate parts or update it to the latest.

Issue #1231, opened 2018-03-09
Resizing windows on Windows 10 shows jumpy behaviour 


edwinRNDR:
When resizing a window on Windows 10 (likely other versions of Windows are affected too) the window shows a stretched version of the frame buffer until the refresh callback has completed. This is different from GLFW's behaviour on MacOS where the window contents are cropped. The stretching behaviour makes resizing look very jumpy, especially in cases where the window is used to show a user interface.

ghuser404:
You should do rendering and event processing on separate threads.

edwinRNDR:
@ghuser404 Can you explain how this would improve the resizing behaviour? As I see it the window's contents will always be shown in stretched fashion until a redraw has completed and buffers are swapped.

ghuser404:
Correct me if I'm wrong, but if you're doing it on one thread, then while resizing, you're not rendering new frames. If you have rendering on a separate thread, then you can render new frames while resize takes place. This way, instead of stretching, you can show left/right or top/bottom black bars.

edwinRNDR:
OK, I can see that work. Still I am wondering if the default stretching behaviour can be replaced/configured to crop/pad to address this without the complexity of extra threads.

tombsar:
To add to ghuser404's answer, Windows does some really questionable stuff with the window's message queue while the resize is going on, causing glfw's message pump to become blocked. It seems there are ways to get around this without extra threads (see https://www.gamedev.net/forums/topic/488074-win32-message-pump-and-opengl---rendering-pauses-while-draggingresizing/?do=findComment&comment=4189300), but it looks fairly messy to me.

I'm not aware of a way to change the stretch/crop behaviour. Any Windows devs know about this?

khaledkee:
I've tested it on windows 10 and windows 81. The stretching behavior occur only when running on Nvidia GPU. On AMD GPU, Intel HD, and Intel iris, black padding appears while resizing without stretching.

malytomas:
I would LOVE to see the window resizing problem being _fixed_ in glfw.
Glfw seems to me like the right place to do it.

Issue #561, opened 2015-07-19
Win32 Modal Loop


xanather:
Due to the Win32 modal message loop whenever the window is being dragged or resized the rendering/updating stops due to the thread being locked in the glfwPollEvents() function.

The Win32 implementation should call the WindowRefreshCallback whenever the window is in a modal loop. It is possible through the use of Win32 timers and dispatching constant WM_TIMER events (Windows supports a rate greater than 60 FPS with this).

I am unsure how glfw handles this on other platforms but I am willing to implement this on Windows at-least since I have it already done on in my own code for the Direct3D back-end and it works well.

I'll test to see what happens when dragging/resizing on other platforms soonish.

elmindreda:
If you want to keep rendering while the window is being moved, do it from another thread.

xanather:
Is that really the only option? I prefer not to have a separate thread to do this.

I will fork and add support for it since is quite straightforward, I'll link the changes here and let you decide since I think it fits well with the definition of WindowRefreshCallback (contents of the window become outdated when dragging).

elmindreda:
Any serious game or tool should already be multi-threaded nowadays and any toy program can probably do without rendering during window dragging.

shurcooL:
> Any serious game or tool should already be multi-threaded nowadays and any toy program can probably do without rendering during window dragging.

I agree. As far as I know, there's no better solution because the Win32 API is the way it is.

xanather:
https://github.com/xanather/glfw/commit/7b205ed2dea4ab78d9781377aece9c19f1cd217f Opinion? It's fine if you'd prefer not to have it added to the main repo. I'll just use it locally for now.

elmindreda:
I won't be merging this. Multi-threading is the robust solution and adds just a few lines in its simplest form.

shurcooL:
> Multi-threading is the robust solution and adds just a few lines in its simplest form.

Does one of the examples demonstrate this?

elmindreda:
@shurcooL https://github.com/glfw/glfw/blob/master/tests/threads.c

xanather:
[...]

Issue #185, opened 2013-12-04
Window moving and resizing blocks in event processing


Cooldude234:
When I re-size the window by use of the cursor or when I pan the window my code stops looping. This means rendering and swapping buffers stop.
I do not like this because my GUI library requires the screen to be constantly updated (so you can tell where buttons, text, containers, etc. are when you resize the window).
I would be very grateful if this was implemented.

PR #1426, opened 2019-02-03
Win32: Do the even processing during window resize/move

mmozeiko:
This change provides workaround for Windows modal event loop while window is being resized/moved.
User can continue using their standard "while" event loop in main thread to do updates in their application.

Change includes two parts:

* Move message dispatching in separate fiber. This allows to "break out" the modal event loop. This is done by using timer which fires after 1msec on window resize/move event. After timer fires, the control returns to main code. There is pretty much no performance overhead for switching fibers in this situation.

* Handle non-client left mouse click on title bar. Otherwise Windows does not return in DefWindowProc. Easy way to do this is to defer calling DefWindowProc on WM_NCLBUTTONDOWN message. Wait until user moves mouse (WMNC_MOUSEMOVE or WM_MOUSEMOVE) and then deliver original message. For example, exact same thing is done in [Google Chromium](https://chromium.googlesource.com/chromium/src/+/master/ui/views/win/hwnd_message_handler.cc#3187) source code.

This fixes issues like #1231, #561, #185.

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -809,10 +809,46 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
             return 0;
         }
 
+        case WM_NCLBUTTONDOWN:
+        {
+            if (wParam == HTCAPTION)
+            {
+                window->win32.ncMouseButton = uMsg;
+                window->win32.ncMousePos = lParam;
+                OutputDebugStringA("down started\n");
+                return 0;
+            }
+            break;
+        }
+
+        case WM_NCMOUSEMOVE:
+        {
+            if (window->win32.ncMouseButton)
+            {
+                if (GET_X_LPARAM(window->win32.ncMousePos) != GET_X_LPARAM(lParam) ||
+                    GET_Y_LPARAM(window->win32.ncMousePos) != GET_Y_LPARAM(lParam))
+                {
+                    DefWindowProcW(hWnd, window->win32.ncMouseButton, HTCAPTION, window->win32.ncMousePos);
+                    window->win32.ncMouseButton = 0;
+                }
+            }
+            break;
+        }
+
         case WM_MOUSEMOVE:
         {
             const int x = GET_X_LPARAM(lParam);
             const int y = GET_Y_LPARAM(lParam);
+
+            if (window->win32.ncMouseButton)
+            {
+                if (GET_X_LPARAM(window->win32.ncMousePos) != x ||
+                    GET_Y_LPARAM(window->win32.ncMousePos) != y)
+                {
+                    DefWindowProcW(hWnd, window->win32.ncMouseButton, HTCAPTION, window->win32.ncMousePos);
+                    window->win32.ncMouseButton = 0;
+                }
+            }
 
             // Disabled cursor motion input is provided by WM_INPUT
             if (window->cursorMode == GLFW_CURSOR_DISABLED)
@@ -918,6 +954,7 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
             if (window->cursorMode == GLFW_CURSOR_DISABLED)
                 enableCursor(window);
 
+            SetTimer(hWnd, 1, 1, NULL);
             break;
         }
 
@@ -929,6 +966,16 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
             if (window->cursorMode == GLFW_CURSOR_DISABLED)
                 disableCursor(window);
 
+            KillTimer(hWnd, 1);
+            break;
+        }
+
+        case WM_TIMER:
+        {
+            if (wParam == 1)
+            {
+                SwitchToFiber(_glfw.win32.mainFiber);
+            }
             break;
         }
 
@@ -1847,31 +1894,10 @@ void _glfwPlatformSetWindowOpacity(_GLFWwindow* window, float opacity)
 
 void _glfwPlatformPollEvents(void)
 {
-    MSG msg;
     HWND handle;
     _GLFWwindow* window;
 
-    while (PeekMessageW(&msg, NULL, 0, 0, PM_REMOVE))
-    {
-        if (msg.message == WM_QUIT)
-        {
-            // NOTE: While GLFW does not itself post WM_QUIT, other processes
-            //       may post it to this one, for example Task Manager
-            // HACK: Treat WM_QUIT as a close on all windows
-
-            window = _glfw.windowListHead;
-            while (window)
-            {
-                _glfwInputWindowCloseRequest(window);
-                window = window->next;
-            }
-        }
-        else
-        {
-            TranslateMessage(&msg);
-            DispatchMessageW(&msg);
-        }
-    }
+    SwitchToFiber(_glfw.win32.messageFiber);
 
     handle = GetActiveWindow();
     if (handle)
--- a/src/win32_init.c
+++ b/src/win32_init.c
@@ -331,7 +331,6 @@ static void createKeyTables(void)
 //
 static HWND createHelperWindow(void)
 {
-    MSG msg;
     HWND window = CreateWindowExW(WS_EX_OVERLAPPEDWINDOW,
                                   _GLFW_WNDCLASSNAME,
                                   L"GLFW message window",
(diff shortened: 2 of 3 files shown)


src/win32_window.c
```diff
@@ -809,11 +809,47 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg,
             return 0;
         }
 
+        case WM_NCLBUTTONDOWN:
+        {
+            if (wParam == HTCAPTION)
+            {
+                window->win32.ncMouseButton = uMsg;
+                window->win32.ncMousePos = lParam;
+                OutputDebugStringA("down started\n");
```
GreenLightning: This line should probably be removed.

GreenLightning:
Very nice pull request!

I tested it on Windows 10.

I am currently rendering from the window position callback and the window refresh callback to work around the modal event loop. This kind of works if the user keeps moving the mouse, which luckily is usually the case while moving / resizing. But my workaround fails when the user stops moving the mouse or just clicks on the title bar without dragging. Using this pull request it works perfectly and the code is simpler because I no longer need the callbacks.

One potential disadvantage is that this causes the application to block Windows from updating the window position and size. This is probably not noticeable in the example applications, because they update and render so quickly. To make the issue apparent, e. g. in the `particles` example, add a delay to the beginning of `draw_scene` like this:

```
	// Sleep for 50 milliseconds
	struct timespec time;
	clock_gettime(CLOCK_REALTIME, &time);
	time.tv_nsec += 50 * 1000000;
	thrd_sleep(&time, NULL);
```

This will reduce the framerate to slightly less than 20 FPS to simulate a really slow application. If you move / resize the window now it will look very choppy with this pull request, whereas before the rendering was interrupted by the modal event loop, but the window was moving very smoothly.

src/win32_window.c
mmozeiko: Thanks! Removed it.

elmindreda:
Will look into this after 3.3 is out.

nathan818fr:
Great pull request !
I have implemented it on the alpha version of one of my java program (using Java w/ LWJGL). 
Some users have reported a stack overflow error from the glfwPollEvents call (on Windows 10, amd64).

Source code: https://github.com/nathan818fr/glfw/tree/3f2cd09cc3017b9f6c89b3156d2106b776916cef
Note: The stack overflow error comes from the native method, not from the Java program. And when I revert the commit: no more bug.

mmozeiko:
Oh, it slipped my mind to consider using glfw in other languages. I could see garbage collected languages such as Java or C# having an issue here. Not sure how well they will work with fibers.

Probably this kind of functionality should be off by default and turned on with hint from user only.

Any chance to get full stack backtrace from the error? Or coredump/minidump?

nathan818fr:
I will send you more information as soon as possible (I can not reproduce the bug on my PC).
Note however that the pull request worked well for more than 10,000 users. But 3 users raised this problem.

mmozeiko:
Yeah, that sounds exactly like the result what I would expect from garbage collector. Most of the time it works fine. But in some very special case, when memory aligns in specific way and GC kicks in at particular time  - everything breaks in a weird way. The more I think about this, the more I'm convinced that fibers used in this way won't work well with managed languages who have garbage collectors. The stack will be completely different from normal execution, so it won't be able to collect all live references and will probably try to free objects too soon.

I guess I'll update pull request with extra window hint that enables this behavior.

RUSshy:
Thanks i needed this !!!

RUSshy:
I just tried with: https://github.com/BindBC/bindbc-glfw/

And it works fine, thanks for this !!

IamRyusei:
does this solves #1451 ?

mmozeiko:
I've added GLFW_WIN32_MESSAGES_IN_FIBER init hint to enable this functionality only on request from user.

prithvidiamond1:
I have a doubt... how can I make use of this to fix the skippy resize issue? I am a noob and I am using the python bindings which has a garbage collector, so has the issue with garbage collector languages been patched?

mmozeiko:
You add `glfwWindowHint(GLFW_WIN32_MESSAGES_IN_FIBER, GLFW_TRUE)` before call to glfwCreateWindow. There is no way to "patch" issues with GC languages. This is fundamentally incompatible with languages that use GC, unless they are fiber aware (which I don't think they are).

But I think Python should be fine - because it does not JIT the code. All GC work happens only on Python objects, not on native memory. Haven't tried it though, so it's up to you.

prithvidiamond1:
@mmozeiko Thanks! seems easy enough to do... However I don't think any of the python bindings have been updated with GLFW_WIN32_MESSAGES_IN_FIBRE... I overheard that this patch hasn't been merged with the official repo... So until that happens I guess I am stuck... Any idea when it will be merged?

mmozeiko:
As long as you build glfw dll yourself from this pull request, you can use existing Python bindings. 

GLFW_WIN32_MESSAGES_IN_FIBER is just a number, take it from glfw3.h header (value 0x00051003) and write it explicitly in Python code.

prithvidiamond1:
Ok I will look into it tomorrow (it is really late where I live...). However, I might need some help, as I am kinda of a noob at this kinda of stuff...

Xiliusha:
Thanks! Moving and sizing modal loop block my game. Some player using this feature to cheat. I hate it.
Could I embedded this solution in my game?

mmozeiko:
You should prevent cheating by logical checks, not by workarounds for win32 api functions. If delaying window messages is all the cheaters need to cheat, then they can simply write extra software that will hook your GLFW window procedure and produce artificial delays. It's pretty trivial to do that. And this branch or any other GLFW functionality won't prevent that.

joaodasilva:
The resize callbacks don't get dispatched if the resize mouse-down happens while the window isn't focused. E.g. click on another application, then click on the window border and start resizing: events won't get dispatched again.

This is caused by the window->win32.frameAction check and early return when handling WM_ENTERSIZEMOVE / WM_EXITSIZEMOVE; that check is meant to skip the enableCursor / disableCursor calls though, and shouldn't skip the timer setup code introduced by the patch.

tycho:
I just tested this out (manually rebased on current master branch). It works **very** well in the samples I tested -- better than other approaches I've seen in other libraries like this, and better than my own hacked-together solution as well.

The only complaint I have is that rendering still pauses if the user clicks and holds on any of the minimize/maximize/close window buttons. Know of any way to address that?

jeancolasp:
Hi everyone.

Thanks for the great work done with GLFW. Is there any update about this feature? It looks promising and seems like it could be fixing people's issues with moving a window through the title bar and resizing windows. Is it now available in GLFW's main branch, and will it be? Thank you.

aqilc:
> Will look into this after 3.3 is out.

Will this be looked at now? It has been a substantial amount of time and many people would benefit from this feature.

Kian738:
What happend to this PR? Many people would indeed benefit from this 😄

@GreenLightning is there anything that keeps this from being merged?

GreenLightning:
> What happend to this PR? Many people would indeed benefit from this 😄
> 
> @GreenLightning is there anything that keeps this from being merged?

I am not a maintainer, just an enthusiastic GLFW user that has the same problems with the win32 api as everyone else :D

Pinging @elmindreda for decisions...

Kian738:
> > What happend to this PR? Many people would indeed benefit from this 😄
> > @GreenLightning is there anything that keeps this from being merged?
> 
> I am not a maintainer, just an enthusiastic GLFW user that has the same problems with the win32 api as everyone else :D
> 
> Pinging @elmindreda for decisions...

Whoops! Sorry for the slip up 😅
But thanks for the ping - much appreciated!

james-bern:
Hi! :) Is this going to be merged? It's been over 6 years and seems kind of important. Thanks!

Kian738:
@dougbinks @elmindreda Is there a specific reason to why this has been stalling for so long? Afaik there will be no compatibility issues either.

dougbinks:
A significant amount of open source, like GLFW, are essentially hobby projects undertaken by a small number of people with limited time. If you need this functionality then I suggest you merge the PR into your own fork - I have personal forks with changes in many of the open source projects I use including GLFW.

On a personal note every time I see a comment like this it makes it harder for me to do any work on the project.

Kian738:
> A significant amount of open source, like GLFW, are essentially hobby projects undertaken by a small number of people with limited time. If you need this functionality then I suggest you merge the PR into your own fork - I have personal forks with changes in many of the open source projects I use including GLFW.
> 
> On a personal note every time I see a comment like this it makes it harder for me to do any work on the project.

That's indeed what I do, I was just annoyed at the time and acted a bit rude. I was annoyed that I had to merge in new commits into my fork haha

malytomas:
has anyone tested this with vulkan? i am trying to use fibers (not this PR, but my own attempt that does not require modifying glfw), and i get an error VK_ERROR_NATIVE_WINDOW_IN_USE_KHR.

PR #1428, opened 2019-02-07
Win32: Fix flaky GetAsyncKeyState bit for CAPITAL and NUMLOCK

Kangz:
This bug was reported on the Chromium bug tracker at https://crbug.com/922095 and seems to have been caused by a race between 7c87948eafa63cf8b6b3d020a26189df6a6d3527 that fixed the use of the first bit, and 0e8c4ea7ce653cc7346bfe9fd5e5ab3933961934 that added support for CAPITAL and NUM_LOCK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ information on what to include when reporting a bug.
                   (#1315,#1316)
 - [Win32] Bugfix: A title bar would be drawn over undecorated windows in some
                   circumstances (#1383)
+- [Win32] Bugfix: GetAsyncKeyState used flaky bit for CAPS and NUM LOCK (#1428)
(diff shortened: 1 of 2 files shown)


elmindreda:
The code seems correct as is. The MSB of the Win32 key state does indicate whether the physical key is down, but for a <kbd>* Lock</kbd> key GLFW uses its lock state for the corresponding modifier flag, not its pressed state. The lock state is indicated in the LSB of the Win32 key state.

Kangz:
My bad, the bug report in Chromium was against an old version of GLFW that uses `GetAsyncKeyState` while current master uses `GetKeyState` which doesn't have this issue.

Sorry for the noise, closing this.

elmindreda:
No worries!

PR #1431, opened 2019-02-11
DPI aware standard cursor icons

MaxRis:
Since [LoadCursor doesn't support dpi virtualization](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-loadcursora#dpi-virtualization), it leads to wrong scaling of user pointer cursor icon on multihead Windows setup in case if connected monitors have different dpi.

Changelist replaces `LoadCursor` with `LoadImage` to redirect correct scaling job to OS by setting `LR_DEFAULTSIZE`. `LR_SHARED` is required there since standard icons images are used and it assures that memory will be reused for similar cursor icons.

Normal "standard" cursor size after app started on primary monitor:

normal-cursor-size

Oversized cursor issue after external monitor with lower DPI pluged-in as primary display:

cursor-size-lower-dpi

### DIFF
--- a/src/win32_platform.h
+++ b/src/win32_platform.h
@@ -61,6 +61,9 @@
 // GLFW uses DirectInput8 interfaces
 #define DIRECTINPUT_VERSION 0x0800
 
+// Enable standard cursors images defines
+#define OEMRESOURCE
+
 #include <wctype.h>
 #include <windows.h>
 #include <dinput.h>
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -1999,24 +1999,25 @@ int _glfwPlatformCreateCursor(_GLFWcursor* cursor,
 
 int _glfwPlatformCreateStandardCursor(_GLFWcursor* cursor, int shape)
 {
-    LPCWSTR name = NULL;
+    int id = 0;
 
     if (shape == GLFW_ARROW_CURSOR)
-        name = IDC_ARROW;
+        id = OCR_NORMAL;
     else if (shape == GLFW_IBEAM_CURSOR)
-        name = IDC_IBEAM;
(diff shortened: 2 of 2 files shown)


elmindreda:
Looks good, thank you!

MaxRis:
@elmindreda thanks! have updated description with screenshots

elmindreda:
This patch loads a shared cursor, which is then destroyed in `_glfwPlatformDestroyCursor`. It should either make a copy of it or track whether the cursor is shared.

MaxRis:
@elmindreda makes sense!

MaxRis:
@elmindreda , If I'm going to use `CopyCursor` the issue with wrong size returns back.

It seems there is no way in Win API to check if cursor is shared or not.

Beside the fact that [doc](url) says "Don't use DestroyCursor" to destroy shared cursor, another more general [doc](https://docs.microsoft.com/en-us/windows/desktop/menurc/about-cursors#cursor-destruction) clearly states:
```
You can destroy the cursor handle and free the memory the cursor used by calling the DestroyCursor function. However, this function has no effect on a shared cursor. A shared cursor is valid as long as the module from which it was loaded remains in memory.
```
So calling of `DestroyCursor` doesn't return error or corrupts shared cursor data, it's just silently returns.
Does it make sense to left the change as is, or, maybe, add collection to gather standard cursors handles to avoid call of `DestroyCursor` if collection lookup succeeded?

elmindreda:
Oh, that's fine then. Thank you! I'll merge this.

elmindreda:
This has been merged as cf79f0625717db86dd86afae2c169d87ba1eb650. Thank you!

PR #1443, opened 2019-03-01
Remove wl_shell support

linkmauve:
This protocol is part of the core Wayland, but it is pretty badly
designed and is missing quite a few features, and is in the process of
being phased out in compositors.  Its support in GLFW requires
duplicating pretty much every single window management codepath.

This bumps the required compositor versions to the ones which have
implemented xdg-shell, approximately two years ago, which seems sensible
to me.

### DIFF
--- a/docs/compat.dox
+++ b/docs/compat.dox
@@ -104,11 +104,7 @@ has been configured in the compositor.
 GLFW uses the [xdg-shell
 protocol](https://cgit.freedesktop.org/wayland/wayland-protocols/tree/stable/xdg-shell/xdg-shell.xml)
 to provide better window management.  This protocol is part of
-wayland-protocols 1.12, and mandatory at build time.  If the running compositor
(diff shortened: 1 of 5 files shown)


elmindreda:
Seems good 👍

kovidgoyal:
FYI neither weston nor sway have current releases that support the stable version of this, so if you want to do it, you will need to add support for the unstable version as well (zxdg_shell_v6)

linkmauve:
Sway at least does, since its 1.0 release.

As for Weston, it is in RC1 stage now, and most users won’t use it as their main compositor anyway so we can easily ignore it.

kovidgoyal:
Yes but sway 1.0 was just released, which means lots of people wont have
it on their distros for a while.

linkmauve:
3.3 got released, there is no reason not to merge this anymore, the next version will most likely get released in a world where wl_shell is completely irrelevant. \o/

PR #1445, opened 2019-03-06
Add minimal wintab32 support

anael-seghezzi:
Hi, I would like to propose this minimal pen tablet support (wacom etc) using wintab32 for win32.
With support for x, y, z, pressure, pitch, yaw, roll, cursor (pen, eraser etc) and proximity.

@elmindreda, I have something similar for cocoa and x11 but I wanted to check first if you are ok with this.

I tested variants of these over the years by modifying glfw's code for in-house painting tools, mainly on mac and windows.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1510,6 +1510,57 @@ typedef void (* GLFWmonitorfun)(GLFWmonitor*,int);
  */
 typedef void (* GLFWjoystickfun)(int,int);
 
+/*! @brief The function signature for pen tablet data callbacks.
+ *
+ *  This is the function signature for pen tablet data callback functions.
+ *
+ *  @param[in] x pen position relative to the screen.
+ *  @param[in] y pen position relative to the screen.
+ *  @param[in] z pen position relative to the tablet.
+ *  @param[in] pen pressure from 0.0 to 1.0.
+ *  @param[in] pen pitch in degree.
+ *  @param[in] pen yaw in degree.
+ *  @param[in] pen roll in degree.
+ *
+ *  @sa @ref pen_tablet_data
+ *  @sa @ref glfwSetPenTabletDataCallback
+ *
+ *  @since Added in version 3.3.
+ *
+ *  @ingroup input
+ */
+typedef void (* GLFWpentabletdatafun)(double,double,double,double,double,double,double);
+
+/*! @brief The function signature for pen tablet cursor callbacks.
+ *
+ *  This is the function signature for pen tablet cursor callback functions.
+ *
+ *  @param[in] pen cursor identifier.
+ *
+ *  @sa @ref pen_tablet_cursor
+ *  @sa @ref glfwSetPenTabletCursorCallback
+ *
+ *  @since Added in version 3.3.
+ *
+ *  @ingroup input
+ */
+typedef void (* GLFWpentabletcursorfun)(unsigned int);
+
+/*! @brief The function signature for pen tablet proximity callbacks.
+ *
+ *  This is the function signature for pen tablet proximity callback functions.
+ *
+ *  @param[in] pen proximity state.
+ *
+ *  @sa @ref pen_tablet_proximity
+ *  @sa @ref glfwSetPenTabletProximityCallback
+ *
+ *  @since Added in version 3.3.
+ *
+ *  @ingroup input
+ */
+typedef void (* GLFWpentabletproximityfun)(int);
+
 /*! @brief Video mode type.
  *
  *  This describes a single video mode.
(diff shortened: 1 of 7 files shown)


elmindreda:
This looks very promising. Will review after 3.3 is out. Two issues remaining now.

anael-seghezzi:
If you are ok with the structure of it, I can do the same for X11. What do you think ? (The issues reported by Travis should be fixed now).

elmindreda:
I can't review it in detail right now but at minimum it looks like a very solid foundation. The cursor and proximity events and their parameters could perhaps do with a bit more documentation.

anael-seghezzi:
Ok, thanks, I'll detail the doc.

anael-seghezzi:
I added XInput2 basic support. Handling the "Coordinate Transformation Matrix" property was a bit tricky but this way we can get X and Y at full precision. I didn't find how to read Z coordinate though. I tested it on Linux Mint with a Graphire2 and an Intuos4 but it should be tested on other systems.

jigpu:
Saw your tweet @anael-seghezzi and thought I'd let you know that the Z coordinate isn't reported by the X driver. The only way to get it on Linux is through the Wayland tablet protocol or directly through the kernel. Also looks like you're properly using the min/max data of the valuators, so it should be 8K-safe :+1:

anael-seghezzi:
Another thing I don't know how to handle correctly on x11 linux is to catch the tablet being unplugged en route and disable it. Right now it causes a crash.

elmindreda:
I have an Intuos on loan now and have tested this PR on Windows, X11 and macOS. It worked well on all platforms. Will try to help out with the X11 disconnect issue.

_Definitely_ want this in 3.4. Thank you!

jigpu:
@anael-seghezzi I'm not sure if you're already handling the `HierarchyChanged` events that are sent when devices are enabled/disabled/added/removed, but that might be the key if you aren't. See http://who-t.blogspot.com/2009/06/xi2-recipies-part-2.html for more detail on event and the following trimmed-down trace when unplugging a tablet from my system:

~~~
$ xinput test-xi2 --root
⎡ Virtual core pointer                    	id=2	[master pointer  (3)]
[...]
⎜   ↳ Wacom Intuos Pro M Pen stylus           	id=9	[slave  pointer  (2)]
⎜   ↳ Wacom Intuos Pro M Pad pad              	id=10	[slave  pointer  (2)]
⎜   ↳ Wacom Intuos Pro M Finger touch         	id=11	[slave  pointer  (2)]
[...]
⎜   ↳ Wacom Intuos Pro M Pen eraser           	id=22	[slave  pointer  (2)]
⎜   ↳ Wacom Intuos Pro M Pen cursor           	id=23	[slave  pointer  (2)]
[...]

EVENT type 12 (PropertyEvent)
     property: 166 'Device Enabled'
     changed: modified
EVENT type 11 (HierarchyChanged)
    Changes happened:        [device disabled]
[...]
    device 10 [slave pointer (2)] is disabled
    changes:        [device disabled]
[... Lots of "EVENT type 12 (PropertyEvent)" as device props are deleted ...]
EVENT type 11 (HierarchyChanged)
    Changes happened:    [slave removed]    
[...]
    device 10 [<undefined> (0)] is disabled
    changes:    [slave removed]    
[... Repeat the sequence of "HierarchyChanged" and "PropertyEvent" events for all the other disconnected devices ...]
~~~

anael-seghezzi:
Thank you jigpu I'll have a look at it.

hanatos:
heya, just wanted to say thanks for the pull request and that i'm successfully using it. to resolve build/merge problems, i had to add this commit:
https://github.com/hanatos/glfw/commit/709d88f5d87d12c7b2b58e601526a35fd00508a3

Cazadorro:
What is going on with this branch, are we just waiting for review?

Cazadorro:
@hanatos Hey there's something very screwy going on with your repo.  Like I think the git history is corrupted or something, if I look at your repo on github, I see case WT_PROXIMITY: in the correct place, but if I clone your repo, `case WT_PROXIMITY:` appears out of order and causes `\hanatos_glfw\src\win32_window.c(550): error C2046: illegal case`

hanatos:
@Cazadorro ah nice catch. i don't normally have access to macintosh or windows computers, so i just didn't notice. i mean i can try and fix this syntactically (it's a mixture of cascaded `if` and `switch` statements, probably should be switch all the way) but i won't be able to thoroughly test on windows computers.

hanatos:
but yes, it would be great to have this merged! let me know if i can help. i could offer rebasing to current master, testing the linux version and i might figure out a way to test windows too.

Cazadorro:
@hanatos You can use github CI for free to test these compilation issues on windows with out using a windows computer yourself, the fixes appear to have gotten the vulkan example to work in glfw.

hanatos:
fair point. pushed to the master branch in my repo now, so it will run the checks. it's also rebased to current upstream/master which was surprisingly simple. i tested drawing on linux and can confirm this still works.

Cazadorro:
@anael-seghezzi There needs to be the ability to set a user pointer for use in the callbacks.  Every single other GLFW API callback function has this capability.  Because there's apparently no object assigned to this (unlike monitor or joystick, or no associated window ala with `GLFWWindow`), I'm not sure how that would work, maybe just have `void* user_ptr` be pass in as the last parameter to the function.  

The reason `user_ptr` is so important is that it allows wrappers to use arbitrary `std::function` objects, which may capture state, by delegating calls to the `user_ptr` instead of the actual function pointer with a wrapper function.  This also allows *lambdas which capture variables* to be passed in as callbacks.

Cazadorro:
@anael-seghezzi @hanatos Confirms this *does not* work on windows.  My huion tablet can correctly be used as a mouse cursor, but no tablet callbacks are triggered when running tests.  I've never had issues with this in any other program.  It's possible that wintab32 is just not the correct API to use on windows period, as even Wacom recommends against using it, instead opting for Windows Ink.

Cazadorro:
I'm investigating what needs to happen to get Windows Ink API to work.  there may be many touch devices, and I think a GLFWmonitor like API may be needed for this.  This repo seems to have a pretty minimalistic easy example of how to use the API on windows: https://github.com/glynskyi/stylus_api/blob/main/windows/stylus_api_plugin.cpp

Cazadorro:
I got windows ink to work with GLFW (on the outside), but I found in the process that the wintab version of GLFW eats my inputs or does something else horribly wrong, windows ink will only work for my device if I *don't* use wintab.  This means that you will actually *break peoples code* with the current implementation.    I also tested Krita and realized that I had been using wintab32 the entire time with out problems, and can *also* use windows ink version entirely with out problems, which seems to indicate to me there's something weird going on with this library.  I have a feeling it has to do with initialization, in the fact that we suspiciously aren't passing any info on exactly *what* device we want to use, but it's hard to find API documentation for wintab to figure out if this is actually the case.

anael-seghezzi:
I have been using my original pull request fork on production for some time, tested on windows 7 and windows 10, on many wacom products. But when testing on XP-PEN Artist 22 Pro, I ran in some issues, you can see the type of fixes here https://github.com/anael-seghezzi/glfw/commit/8533f58c91f94421f6b7a6204ff9073959159695 and here https://github.com/anael-seghezzi/glfw/commit/213a2965ddc29dc31843524f5baab13926d5b5f3. It's possible that the Huion tablet trigger some specific issue similarly to the XP-PEN.

Issue #1452, opened 2019-03-19
GLFW makes unnecessary and expensive calls to DwmIsCompositionEnabled


kliffy542:
In wgl_context.c there are two methods which call the Windows API DwmIsCompositionEnabled. On Windows 8+ these calls are unnecessary because dwm composition is always enabled and the method is guaranteed to return true. See the Chromium patch referenced here: https://bugs.chromium.org/p/chromium/issues/detail?id=942165

PR #1453, opened 2019-03-19
Bypass DwmIsCompositionEnabled checks on Win8+.

kliffy542:
DWM composition is always enabled on Win8+, so there's no need to call DwmIsCompositionEnabled on those platforms. See also https://bugs.chromium.org/p/chromium/issues/detail?id=942165.

### DIFF
--- a/src/wgl_context.c
+++ b/src/wgl_context.c
@@ -321,10 +321,15 @@ static void swapBuffersWGL(_GLFWwindow* window)
     {
         if (IsWindowsVistaOrGreater())
         {
-            BOOL enabled;
+            BOOL enabled = FALSE;
+          
+            // DWM Composition is always enabled on Win8+
+            if (IsWindows8OrGreater())
+              enabled = TRUE;
 
             // HACK: Use DwmFlush when desktop composition is enabled
-            if (SUCCEEDED(DwmIsCompositionEnabled(&enabled)) && enabled)
+            if (enabled ||
+                (SUCCEEDED(DwmIsCompositionEnabled(&enabled)) && enabled))
             {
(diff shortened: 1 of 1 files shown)


kliffy542:
This is in reference to issue #1452 but I'm not sure how to link them.

src/wgl_context.c
```diff
@@ -321,10 +321,15 @@ static void swapBuffersWGL(_GLFWwindow* window)
     {
         if (IsWindowsVistaOrGreater())
         {
-            BOOL enabled;
+            BOOL enabled = FALSE;
+          
+            // DWM Composition is always enabled on Win8+
+            if (IsWindows8OrGreater())
```
ghuser404: Instead of doing this check, you can just initialize `enabled` with this call
```
BOOL enabled = IsWindows8OrGreater();
```

ghuser404: An even better solution would be to change the initial condition
```
if (IsWindowsVistaOrGreater() && !IsWindows8OrGreater())
```

kliffy542: Good call on the initialization, that's much better. I don't think I can do your second suggestion though because these blocks are intended to run when dwm composition is enabled, so they definitely should execute on Win8+.

ghuser404: My bad, sorry. I got confused, thought we're skipping this part.

elmindreda:
I'm looking to remove or replace the DWM hack but if we do keep it I will definitely merge this fix. Thanks!

elmindreda:
This has been merged as 87458617e0d6094630a255d043c55e085f70c07f.

PR #1455, opened 2019-03-23
Alternative syntax for function access

Johnmper:
Does not had overhead.
Can be used in junction with the normal `<GLFW/glfw3.h>` in older projects or as a standalone `<GLFW/glfw3cxx.h>`
Both of the following are valid:
```
#include<GLFW/glfw3cxx.h>
glfw::Init();
glfw::Terminate();
```
or integrating with older projects as long as `glfw3cxx.h` is included after `glfw3.h`
```
#include<GLFW/glfw3.h>
#include<GLFW/glfw3cxx.h>
glfwInit();
glfw::Terminate();
```

For simplicity sake, the creation of the function names for `glfw3cxx.h` is partially automated, by directly extracting from the glfw3.h using the following script:
```
grep "glfw[^(]*([^)]*);" glfw3.h | grep -oh "glfw[^(]*" | sed "s/glfw//" |  while read FUNC; do echo "constexpr auto $FUNC = glfw$FUNC;"; done
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+- Added c++ namespace syntax as alternative for accessing glfw functions,
+  `glfwInit()` becomes `glfw::Init()`.
 - Added `glfwGetError` function for querying the last error code and its
   description (#970)
 - Added `glfwUpdateGamepadMappings` function for importing gamepad mappings in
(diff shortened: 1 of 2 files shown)


Johnmper:
Ups, forget to protect the glfw3cxx header from multiple inclusions.

gracicot:
I'm not sure it's a good idea to make glfw functions private symbols with a `namespace {`

Johnmper:
Maybe not, i got it to work though, had to do LOTS of modifications. All of the tests and examples work just fine with the alternative syntax. 
But maybe it isn't really useful.

PR #1460, opened 2019-04-04
Add an 'n' in window.dox

r00ster91:
A tiny docs fix.


elmindreda:
Thank you for this 'n'. It may be tiny but that doesn't make it less welcome.

PR #1468, opened 2019-04-17
Fix issue with Windows multi-threaded applications not being able to correctly get the GLFW_FOCUSED attribute from outside the main thread

lggmonclar:
Multi-threaded Windows applications that attempt to get the `GLFW_FOCUSED` attribute from one of its worker threads will always be returned 0. 

This issue is being caused by the `GetActiveWindow()` calls in `win32_window.c:1791` and `win32_window.c:1921`, which returns a `nullptr` if it is called by a worker thread, even if `glfwMakeContextCurrent()` is correctly set.

Replacing these calls with `GetForegroundWindow()` fixes this issue, because it always returns the foreground window's handle, even from worker threads.

References:
* **GetActiveWindow**: https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-getactivewindow
* **GetForegroundWindow**: https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-getforegroundwindow
* **Relevant Microsoft devblogpost**: https://devblogs.microsoft.com/oldnewthing/?p=20643

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
-User-visible changes since the last release.
+* [Win32] Bugfix: Worker threads are now able to correctly detect if the window has focus by getting the `GLFW_FOCUSED` attribute.
 
 
 ## Contact
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -1788,7 +1788,7 @@ void _glfwPlatformSetWindowMonitor(_GLFWwindow* window,
 
 int _glfwPlatformWindowFocused(_GLFWwindow* window)
 {
-    return window->win32.handle == GetActiveWindow();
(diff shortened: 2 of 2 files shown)


jlodos:
FWIW we have encountered this issue. It also affects glfwSetCursorPos which internally uses _glfwPlatformWindowFocused. So glfwSetCursorPos won't work when used from a thread different from the window creation thread. I can also confirm the proposed solution fixes the problem.
We are using ImGui with GLFW, ImGui fails in a multithreaded application because it internally uses both glfwGetWindowAttrib with GLFW_FOCUSED and glfwSetCursorPos.

JeremyBorys:
Can anyone think of anything that might cause any issues with this? Is there any reason this hasn't been put in?

PR #1490, opened 2019-05-27
Win32: Invoke cursor enter callback before cursor position callback

TheMrMilchmann:
Currently, on windows, when the cursor enters a window the cursor position callback is invoked once before the cursor enter callback is invoked. This behavior is confusing, does not seem to have any purpose. Additionally (from what I can tell by looking at the code) this is not the case for other platforms.

This PR provides a simple fix by moving some logic in the related Win32 event handling code in order to guarantee that the cursor enter callback is invoked before the cursor position callback (when the cursor enters the window.) After some brief testing I could not find any regressions caused by this change.

Note: The PR is currently targeting the `master` branch and thus the `3.4` release since I was not sure whether this could go into a patch update. Other than that I followed the contribution guidelines and enabled PR edits from maintainers, but if there is anything else for me to do, please let me know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ information on what to include when reporting a bug.
  - Disabled tests and examples by default when built as a CMake subdirectory
  - Bugfix: The CMake config-file package used an absolute path and was not
    relocatable (#1470)
+ - [Win32] Bugfix: Invoke cursor enter callback before cursor position callback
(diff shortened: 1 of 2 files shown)


elmindreda:
This was merged as 4f0b8b0ddaada126cd257f2e5b06450e932ad5f6. Thank you!

PR #1491, opened 2019-05-28
Remove _glfwInputWindowMonitor

tombsar:
`_glfwInputWindowMonitor` did nothing more than assign a member of the `window` struct, so I think the code actually becomes more readable by removing the function and doing the assignment explicitly.

Also cleaned up a pointer non-NULL check in win32_window.c that was causing a false-positive warning in MSVC 2019.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1213,7 +1213,7 @@ void _glfwPlatformSetWindowMonitor(_GLFWwindow* window,
     if (window->monitor)
         releaseMonitor(window);
 
-    _glfwInputWindowMonitor(window, monitor);
+    window->monitor = monitor;
 
     // HACK: Allow the state cached in Cocoa to catch up to reality
     // TODO: Solve this in a less terrible way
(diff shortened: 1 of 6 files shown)


elmindreda:
`_glfwInputWindowMonitor` is a consequence of platform-specific code not being allowed to modify the shared portion of structs like library, window, monitor, context and cursor. There is also `_glfwInputMonitorWindow` for the same reason.

elmindreda:
I'll gladly grab 5950382616df13c547a592635e447a7b43b5f296 right away. I can cherry-pick it into my queue, or if you prefer to have it merged signed, please [wrap the description at 72 characters](https://github.com/agis/git-style-guide).

tombsar:
D'oh! I was misled by nearby lines that modify the platform-specific parts of the same structs. Thanks for the explanation.

Feel free to cherry-pick the one commit. Sorry about the formatting of the commit message; that's what happens when I use GitHub's built-in editor rather than something proper.

elmindreda:
Merged, thank you!

> Sorry about the formatting of the commit message; that's what happens when I use GitHub's built-in editor rather than something proper.

No worries! It's incredibly silly that the GitHub editor doesn't help one to format messages as 50/72 automatically.

Issue #1499, opened 2019-06-13
GLFW_MAXIMIZED  hint not working on Windows in version 3.3


dougbinks:
Reported on our forums in post [Maximising Window differences from 3.2 to 3.3 ](http://discourse.glfw.org/t/maximising-window-differences-from-3-2-to-3-3/1338).

I have confirmed the issue and have a fix in testing which I'll create a PR for shortly, however this turns out to be a slightly complex and subtle area.

**Details**

When the window hint GLFW_MAXIMIZED is set to GLFW_TRUE and a window is created it appears with maximized style but not with a maximized size.

It should instead appear maximized as well as in the maximized style.

**Notes**

Related to issue #780.

dougbinks:
The simplest solution I can see is to alter _glfwPlatformShowWindow() to:

```
void _glfwPlatformShowWindow(_GLFWwindow* window)
{
	if (window->win32.maximized)
		ShowWindow(window->win32.handle, SW_MAXIMIZE);
	else
		ShowWindow(window->win32.handle, SW_SHOWNA);
}
```

Currently this means any maximized window would then be focused, as SW_MAXIMIZE will activate the window and there is no maximize without activate option that I can see. This makes sense as a newly created maximized window would cover all other windows.

elmindreda:
This should be fixed now with 711b9694a133ef338482ae75ada724e3a9915d91. Thank you for reporting this!

PR #1503, opened 2019-06-14
Win32 maximize for issue #1499

dougbinks:


dougbinks:
Note that I haven't altered the changelog as that has been removed from the readme.

Luflosi:
This patch uses tabs and I think spaces are the standard in GLFW.

dougbinks:
Spaces are indeed standard, I've switched them.

elmindreda:
This sentence adds cross-links with #1499.

elmindreda:
I won't merge this fix but it did lead me to the cause of #1499 and a bonus bug. Thank you!

At first I thought the fix shouldn't work as the `win32.maximized` boolean is only used for disambiguation inside the `WM_SIZE` handler and isn't set during creation (the bonus bug). #1499 is caused by the content scale adjustment call to `SetWindowPos` during window creation, which both undoes the maximization from creating the window with `WS_MAXIMIZE` and causes the boolean to be updated early enough.

711b9694a133ef338482ae75ada724e3a9915d91 switches to `SetWindowPlacement` for adjusting the normal window rect without breaking initial maximization.

dougbinks:
Excellent fix, good find!

PR #1514, opened 2019-06-20
[WIP] Add vivante backend

YarikTH:
This adds interface to graphics application to create i.MX Framebuffer to initialize graphics context without any window subsystems. Also provides input using /dev/input devices like linux_joystick.c does.

Modifies in GLFW library to add vivante backend.

This includes modifying a few existing files and creating five new files which are
specific to the backend that we are adding. For eg. in our case these are
vivante_platform.h, vivante_init.c, vivante_window.c, evdev.c, evdev.h .

Files "evdev.c", "evdev.h" can be used for other embended linux platforms.

Files "egl_context.h" and "internal.h" has minor changes to include the header files.
Also there are changes in CMake code files to include the newly created files.

Funtion glfwCreatePositionedWindow added and some platform code was adjusted to use it. It was done because window creation api in vivante doesn't support window position change nor window resize, both position and size set on creation time.

Example "gears-gles2.c" added because all examples use opengl 1.0 api, while my target board doesn't support OpenGL api, but supports GLES 2.0 api. This example is optional and can be removed.

Links to the api description:
https://www.nxp.com/docs/en/user-guide/i.MX_Graphics_User's_Guide_Linux.pdf
Chapter 4 i.MX Framebuffer API

P.S. SDL2 already supports this backed since 2.0.5


README.md
```diff
@@ -1,9 +1,5 @@
 # GLFW
 
-[![Build status](https://travis-ci.org/glfw/glfw.svg?branch=master)](https://travis-ci.org/glfw/glfw)
-[![Build status](https://ci.appveyor.com/api/projects/status/0kf0ct9831i5l6sp/branch/master?svg=true)](https://ci.appveyor.com/project/elmindreda/glfw)
-[![Coverity Scan](https://scan.coverity.com/projects/4884/badge.svg)](https://scan.coverity.com/projects/glfw-glfw)
-
```
YarikTH: I have no idea why it changes. Need to revert

include/GLFW/glfw3.h
```diff
@@ -2399,6 +2399,162 @@ GLFWAPI void glfwWindowHint(int hint, int value);
...
+ *  @thread_safety This function must only be called from the main thread.
+ *
+ *  @sa @ref window_creation
+ *  @sa @ref glfwDestroyWindow
+ *
+ *  @since Added in version 3.3_viv.
+ *
+ *  @ingroup window
+ */
+GLFWAPI GLFWwindow* glfwCreatePositionedWindow(int xpos, int ypos, int width, int height, const char* title, GLFWmonitor* monitor, GLFWwindow* share);
```
YarikTH: Added to allow to create window in given position on vivante backend. But should be implemented for all other platforms. I implemented it for win, x11 and vivante platforms. Tested only on linux and vivante. Don't forget to update version in "since" tag.

src/internal.h
```diff
@@ -255,6 +257,8 @@ struct _GLFWwndconfig
 {
     int           width;
     int           height;
+    int           xpos;
+    int           ypos;
```
YarikTH: added to support positioned windows creation

src/win32_window.c
```diff
@@ -1222,8 +1222,13 @@ static int createNativeWindow(_GLFWwindow* window,
...
     {
-        xpos = CW_USEDEFAULT;
-        ypos = CW_USEDEFAULT;
+        xpos = wndconfig->xpos;
+        ypos = wndconfig->ypos;
+
+        if (xpos == GLFW_DONT_CARE)
+            xpos = CW_USEDEFAULT;
+        if (ypos == GLFW_DONT_CARE)
+            ypos = CW_USEDEFAULT;
```
YarikTH: win32 positioned windows creation

src/x11_window.c
```diff
@@ -787,6 +789,17 @@ static GLFWbool createNativeWindow(_GLFWwindow* window,
...
 
+    // It's useless to set position in XCreateWindow because it is overwrote
+    // anyway. So just call glfwSetWindowPos
+    if (xpos != GLFW_DONT_CARE || ypos != GLFW_DONT_CARE)
+    {
+        if (xpos == GLFW_DONT_CARE)
+            xpos = 0;
+        if (ypos == GLFW_DONT_CARE)
+            ypos = 0;
+        glfwSetWindowPos(window, xpos, ypos);
```
YarikTH: x11 positioned windows creation

YarikTH:
Maybe I should add separate pull request for glfwCreatePositionedWindow. Because it is severe change and doesn't directly connected with vivante backend

linkmauve:
Hello, and thanks for your work!

Are you aware of the [Etnaviv driver](https://gitlab.freedesktop.org/mesa/mesa/commit/c9e8b49b885242d84ba031dacef5aa4a5ac1e5b6) though?  It is a driver for Vivante GPUs which doesn’t require the use of proprietary APIs, which otherwise make it very hard for us to test.  Etnaviv provides standard Linux APIs instead, such as DRM, Wayland, X11, etc., and lets you use the existing backends without any modification.

Just like the Nvidia proprietary backend, I really don’t think the costs of an additional backend for very specific hardware is justified, especially since a free driver exists.

Issue #1520, opened 2019-06-22
glfwPostEmptyEvent does not unblock glfwWaitEvents on Wayland


eXeC64:
I played with [this test case](https://github.com/glfw/glfw/blob/master/tests/empty.c) to check if I was doing something wrong in my own code, but it also reproduces the issue.

On X11, glfw behaves correctly. On Wayland, `glfwPostEmptyEvent` from a second thread fails to wake up `glfwWaitEvents` and `glfwWaitEventsTimeout`.

OlivierSohn:
I think I fixed that in #1285 which is not merged yet

elmindreda:
This should be finally fixed now with a32cbf6d4febde6343167f19529b8bde8966670d, which was partly based on #1285 though the specific fix for Wayland was different. Thank you for the bug report and your patience!

PR #1521, opened 2019-06-22
wl: Fix PostEmptyEvent not waking WaitEvents

eXeC64:
Use an eventfd as a means to signal waitEvents to wake up as
wl_display_sync does not always achieve this.

This fixes GitHub Issue #1520

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -35,6 +35,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/eventfd.h>
 #include <sys/mman.h>
 #include <sys/timerfd.h>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


eXeC64:
Is there anything else needed on my side for this? I'm quite keen to get this fixed, as it causes a deadlock in my application.

warisb:
I ran into the same problem in Ubuntu 18.04 (using GNOME Wayland session). My application was hanging on `glfwWaitEvents()`. I pulled from your fork and it fixed my problem. I also second merging this PR if others don't see any issues.

eXeC64:
@elmindreda Would you be able to spare some time to review this? I'm finding that even non-multithreaded glfw applications randomly hang on Wayland without it.

elmindreda:
@eXeC64 Will do. Cannot promise when.

OlivierSohn:
@eXeC64 in #1285, we fixed a race condition, are you sure this is not fixing the error you are attempting to fix here?

PR #1525, opened 2019-06-29
win32_window : Respect floating.

p-groarke:
From my other closed PR. Fullscreen should not be topmost unless floating is set.
This is important for borderless fullscreen windows, which are expected to behave like normal windows.

If this is too drastic of a change. A new window option should be added to control this behavior, as it makes true borderless fullscreen impossible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,4 @@ tests/title
 tests/triangle-vulkan
 tests/windows
 
+/[Bb]uild*/
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -1741,8 +1741,13 @@ void _glfwPlatformSetWindowMonitor(_GLFWwindow* window,
 
         acquireMonitor(window);
 
+        HWND z_mode = HWND_TOP;
+        if (window->floating) {
+                z_mode = HWND_TOPMOST;
+        }
+
         GetMonitorInfo(window->monitor->win32.handle, &mi);
-        SetWindowPos(window->win32.handle, HWND_TOPMOST,
+        SetWindowPos(window->win32.handle, z_mode,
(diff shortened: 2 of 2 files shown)


elmindreda:
I believe you can achieve this behavior with a non-floating, non-fullscreen borderless window that covers the desired monitor.

p-groarke:
Hi @elmindreda, that's what I've been trying to do :) But when creating the borderless fullscreen window, it goes through this code-path. I'm using the instructions in the window guide, though I might be missing something.

Basically,
```
		if (GLFWmonitor* monitor = get_monitor(_position)) {
			if (const GLFWvidmode* mode = glfwGetVideoMode(monitor)) {
				glfwSetWindowMonitor(_window_ptr.get(), monitor, 0, 0,
						mode->width, mode->height, mode->refreshRate);

				on_resize(mode->width, mode->height);
			}
		}
```

elmindreda:
I mean leave it as a non-fullscreen window (i.e. no monitor set for the window) which will avoid the GLFW fullscreen path.
```c
if (GLFWmonitor* monitor = get_monitor(_position))
{
    int x, y;
    glfwGetMonitorPos(monitor, &x, &y);
    const GLFWvidmode* mode = glfwGetVideoMode(monitor);
    glfwSetWindowAttrib(GLFW_DECORATED, 0);
    glfwSetWindowMonitor(window, NULL, x, y, mode->width, mode->height, 0);
    on_resize(mode->width, mode->height);
}
```
The call to `glfwSetWindowMonitor` isn't doing anything fullscreen-related here, it just sets the size and position of the window.

p-groarke:
I think that will work, thx.

Issue #1201, opened 2018-01-30
Keyboard layout changed callback


ghuser404:
Imagine a game shows a floating tutorial hint where it says "Press W to go forward". I would like to be notified when user switches their keyboard layout to another language, to show W equivalent in that layout. Otherwise, I would have to query the key name over and over.

elmindreda:
I was hoping no one would bring this up until 4.0 when events should be cheaper (in terms of API surface) to add, but you're right. This is needed.

NeatSketch:
@ghuser404 
There are keyboards with two layouts printed on the keys, Latin and Cyrillic. The two alphabets share some common letters, but they usually don't share the same keys. For example, the Cyrillic letter 'A' on a Bulgarian keyboard is on the same place as the English 'D'. Keyboard hotkeys and shortcuts are usually shown in English and the user usually expects that. Imagine showing a hint 'Press A'. The user will probably think the app/game means the English 'A'. But if your app means 'A' in currently selected keyboard layout, the user will try to press the wrong key. I actually saw someone in this situation, and the user was upset because "the controls won't work".

ghuser404:
I am well aware of such keyboards. But think French, for example. Their 'A' is not where English 'A' is.

NeatSketch:
@ghuser404
I know. How is it different from my example?

ghuser404:
Because French keyboard doesn't have English letters printed on the keys, like Bulgarian one does. So if you have a French keyboard with French layout set in OS (because why wouldn't you) and a game tells you to press 'A', you will be pressing 'Q' (because that's what it is in US layout) and be more frustrated.

NeatSketch:
French use AZERTY for both English and French, so if you request the user to press English 'A' they will press the key that both looks to them as 'A' and recognized by the OS as 'A'.

ghuser404:
To ask user with French layout to press 'A', you have to call `glfwGetKeyName(GLFW_KEY_Q, 0);`.

NeatSketch:
Yes, but that has nothing to do with the user's currently selected input language. What matters is the physical keyboard layout. To determine it you should detect which layout is selected for English and display the English letter that corresponds to the needed key. That way you don't have the problem that I explained and also don't have the problem that you explained.

ghuser404:
Please provide code.
AFAIK, there is no such API available.

ghuser404:
Actually, even if there was such API, imagine following scenario:
[...]

PR #1528, opened 2019-07-02
Query and keep track of Xkb group index

### LINKED ISSUE #1462: glfwGetKeyName reports incorrect key strings
I'm having quite some issues with `glfwGetKeyName`.
My keyboard layout is QWERTZ, the country Germany and the language German.
```c
#include <GLFW/glfw3.h>
#include <stdlib.h>
#include <stdio.h>

void key_callback(GLFWwindow* window, int key, int scancode, int action, int mods)
{
    printf("Key: %i, Scancode: %i\n", key, scancode);
    printf("glfwGetKeyName: %s\n\n", glfwGetKeyName(key, scancode));
}

int main()
{
    GLFWwindow* window;

    /* Initialize the library */
    if (!glfwInit())
        return -1;

    /* Create a windowed mode window */
    window = glfwCreateWindow(640, 480, "Hello World", NULL, NULL);
    if (!window)
    {
        glfwTerminate();
        return -1;
    }

    /* Register a callback */
    glfwSetKeyCallback(window, key_callback);

    /* Loop until the user closes the window */
    while (!glfwWindowShouldClose(window))
    {
        /* Poll for and process events */
        glfwPollEvents();
    }

    glfwTerminate();
    return 0;
}
```

Here's a list of keyboard keys on my keyboard (no caps lock) and their output of the above program, all of the outputs are incorrect:

* <kbd>-</kbd>
  ```
  Key: 47, Scancode: 61
  glfwGetKeyName: /
  ```

* <kbd>ß</kbd>
  ```
  Key: 45, Scancode: 20
  glfwGetKeyName: -
  ```

* <kbd>`</kbd>
  ```
  Key: 61, Scancode: 21
  glfwGetKeyName: ▒
  ```

* <kbd>ä</kbd>
  ```
  Key: 39, Scancode: 48
  glfwGetKeyName: ▒
  ```

* <kbd>ö</kbd>
  ```
  Key: 59, Scancode: 47
  glfwGetKeyName: ▒
  ```

* <kbd>ü</kbd>
  ```
  Key: 91, Scancode: 34
  glfwGetKeyName: ▒
  ```

All other key outputs are correct or are `null`.

The terminal I'm using (MobaXterm) is able to display all of these characters when I type them in the terminal.

Operating system: Windows 10, 64-bit
GLFW version string: 3.2.1 X11 GLX EGL clock_gettime /dev/js Xf86vm shared
I'm using the Windows Subsystem for Linux.

Can anyone else reproduce this?

db47h:
Fixes #1462 

For users with multiple keyboard layouts configured, `glfwGetKeyName` works fine only with the primary layout (i.e. on Ubuntu 18.04, the first layout in the list). Switching layouts results in changing the group index used in various Xkb API calls. With this commit, the current group index is fetched when initializing keyboard input and an event handler is set up to keep track of any change to this group index.

As a result the scancode -> keyname mapping may change while the program is running (needs to be documented). This also partially addresses #1201: GLFW now receives the event, a complete fix would need to push it to the application.

Note that this code deals only with the default keyboard (see https://github.com/glfw/glfw/issues/1462#issuecomment-501696245).

### DIFF
--- a/src/x11_init.c
+++ b/src/x11_init.c
@@ -54,7 +54,7 @@ static int translateKeyCode(int scancode)
         // Note: This way we always force "NumLock = ON", which is intentional
         // since the returned key code should correspond to a physical
         // location.
-        keySym = XkbKeycodeToKeysym(_glfw.x11.display, scancode, 0, 1);
+        keySym = XkbKeycodeToKeysym(_glfw.x11.display, scancode, _glfw.x11.xkb.group, 1);
         switch (keySym)
         {
             case XK_KP_0:           return GLFW_KEY_KP_0;
@@ -76,7 +76,7 @@ static int translateKeyCode(int scancode)
 
         // Now try primary keysym for function keys (non-printable keys)
         // These should not depend on the current keyboard layout
-        keySym = XkbKeycodeToKeysym(_glfw.x11.display, scancode, 0, 0);
+        keySym = XkbKeycodeToKeysym(_glfw.x11.display, scancode, _glfw.x11.xkb.group, 0);
     }
     else
     {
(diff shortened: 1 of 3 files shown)


elmindreda:
Tested on MATE on Cygwin/X. Works perfectly.

PR #1530, opened 2019-07-08
CMake: remove -DGLFW_DLL on non-Windows targets

avemilia:
When using GLFW with CMake and installed GLFW binaries, `-DGLFW_DLL` is
passed on Linux, which should not happen.


elmindreda:
Good point, thank you!

PR #1540, opened 2019-07-24
Removed overloading of INSTALL_NAME_DIR with ${CMAKE_INSTALL_LIBDIR} …

realzhtw:
Hi. I don't quite understand why this was done in the first place, but it prevents users from specifying CMAKE_INSTALL_NAME_DIR. In particular it breaks linking against libglfw when (for exmaple) in tracy (https://bitbucket.org/wolfpld/tracy/src) when built using pkgsrc framework that relies on being able to pass CMAKE_INSTAL_NAME_DIR (see https://github.com/jsonn/pkgsrc/blob/trunk/mk/configure/cmake.mk, line 46).

Removing those lines make everything work as expected.


linkmauve:
Hi, thanks for your PR, but this work change is a duplicate of #1504.

PR #1541, opened 2019-07-25

Luflosi:
The `size` member in the `GLFWgammaramp` struct is of type `unsigned int`, so the `for` loop iterating over it should also use the type `unsigned int`.


Luflosi:
I rebased the commit onto the latest master.

Luflosi:
I added the "Closes #1541." text to the commit message.

Luflosi:
I rebased the commit onto the latest master.

Luflosi:
I rebased again.

elmindreda:
Thank you!

PR #1546, opened 2019-08-07
Meson support

wompus1:
added support for meson build system


elmindreda:
Meson is a nice system and this is a good adaption but I will not merge this. Switching meta build system is very disruptive and maintaining two sets of build files is not a good idea.

Issue #1555, opened 2019-08-31
_glfwPlatformGetVideoModes filters out video modes with the same screen area


httpdigest:
There is a bug in how GLFW detects duplicate reported video modes in [_glfwPlatformGetVideoModes](https://github.com/glfw/glfw/blob/master/src/win32_monitor.c#L418), respectively _glfwCompareVideoModes and [compareVideoModes](https://github.com/glfw/glfw/blob/51bb76c7c3de934b2dd3affcdc297a32e4817835/src/monitor.c#L56-L57).
When the OS reports both the resolutions 1920x1440 _and_ 2560x1080, then the latter will be [filtered out](https://github.com/glfw/glfw/blob/master/src/win32_monitor.c#L424) by GLFW since both resolutions have the same screen area (width x height) and are therefore considered equal, which is incorrect.

PR #1556, opened 2019-08-31
Fix comparison of video modes (#1555)

httpdigest:
This fixes the video mode comparator function which is also used to filter-out assumed duplicate video modes (#1555) when returning 0.

### DIFF
--- a/src/monitor.c
+++ b/src/monitor.c
@@ -56,6 +56,10 @@ static int compareVideoModes(const void* fp, const void* sp)
     if (farea != sarea)
         return farea - sarea;
 
+    // Then sort on width
+    if (fm->width != sm->width)
+        return fm->width - sm->width;
+
     // Lastly sort on refresh rate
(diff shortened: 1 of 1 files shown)


tombsar:
Alternative is to leave the screen area check in place, and add your x and y checks afterwards. That way we fix the bug (not being able to distinguish "same area but differently shaped" modes) while keeping more of the current behaviour (preferring "larger" modes first, rather than "wider").

httpdigest:
@tombsar agreed. I changed the commit accordingly.

tombsar:
Thinking about this a bit more, I think the height comparison is now redundant. If the areas are equal and the widths are equal, then the two modes must have the same dimensions.

httpdigest:
> Thinking about this a bit more, I think the height comparison is now redundant. If the areas are equal and the widths are equal, then the two modes must have the same dimensions.
True. Made changes accordingly.

Issue #1236, opened 2018-03-20
Ignoring mouse events / transparent for inputs / disable hit testing


ocornut:
Hello,

Using a long subject line because I don't know what is the most appropriate cross-platform terminology and people might search for this under different names

This is feature request for a window hint, let's tentatively call it `GLFW_MOUSE_PASSTHRU` (exact naming to be determined). I am using it as part of the same work I mentioned in #1166 (GIF at the end of this post).

(Note, this is unrelated to transparency from a visual point of view, already discussed many times here: https://github.com/glfw/glfw/issues?utf8=%E2%9C%93&q=transparent)

----

**Windows**
The WndProc handler would do:
```
if (msg == WM_NCHITTEST)
    return HTTRANSPARENT;
```

**Mac**
I think we can use the 'ignoresMouseEvents' property of 'NSWindow' (untested)
https://developer.apple.com/documentation/appkit/nswindow/1419354-ignoresmouseevents?language=objc
> A Boolean value that indicates whether the window is transparent to mouse events.

**Wayland**
Possibly 'wl_surface::set_input_region' (untested)
https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_surface
> Input events happening outside of this region will try the next surface in the server surface stack. The compositor ignores the parts of the input region that fall outside of the surface. [...] The initial value for an input region is infinite. That means the whole surface will accept input. Setting the pending input region has copy semantics, and the wl_region object can be destroyed immediately. A NULL wl_region causes the input region to be set to infinite. 

**X11**
I'm not sure how X11 works. For reference, Qt uses `xcb_xfixes_create_region()`+`xcb_xfixes_set_window_shape_region_checked`:
https://git.merproject.org/mer-core/qtbase/blob/master/qtbase/src/plugins/platforms/xcb/qxcbwindow.cpp#L1151

----

Context (gif): I am dragging a glfw window (that is right under my mouse cursor) and need to detect the window under so I can drop into them. Similarly to #1166 I can provide a rough (but usable) approximation on client-side by using the desktop size/position of all my windows, but of course this would bypass the window manager and ignore the proper z-order of windows (including windows from foreign application).

![viewport_20180318](https://user-images.githubusercontent.com/8225057/37652084-98d6b8a0-2c3a-11e8-8ee1-0d2d86815ad0.gif)

----

[...]

PR #1568, opened 2019-09-30
Implement window transparency to mouse events (#1236)

rokups:
This is an implementation for window transparency for mouse events requested in #1236. There are some things to discuss before merging.

`GLFW_MOUSE_PASSTHRU` define and `mousePassthru` variable. I used terminology suggested in #1236. Naming is hard so if anyone has a better idea - speak up. Maybe we shuld do `PASSTHRU` -> `PASSTHROUGH` at least.

Is `GLFW_MOUSE_PASSTHRU` defined in appropriate place and is it's value proper?

I tried to add documentation comments but i did so without fully understanding what i am doing. In case anything is missing/incorrect please let me know.

`libXext.so.6` - hope that ".6" is pretty much same across all distributions. I may be wrong.

CYGWIN - untested. Only reason why it would break is `libXext-6.so` being named something else i think.

Not sure if i got all `_glfw.x11.xshape` right.

Wayland while i think is implemented correctly - does not work. I tested with gears example, setting `glfwWindowHint(GLFW_MOUSE_PASSTHRU, GLFW_TRUE);` before window creation. It may be that kwin compositor does not have required bits for this to work. If anyone runs Gnome wayland - i would greatly appreciate if you could test this PR. Otherwise i may get around to setting up a VM for a test, but i would love to avoid it if possible.

X11/Windows/MacOS - everything works as expected.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -832,6 +832,13 @@ extern "C" {
  */
 #define GLFW_FOCUS_ON_SHOW          0x0002000C
 
+/*! @brief Forward mouse input to window behind.
+ *
+ *  Mouse input forwarding[window hint](@ref GLFW_MOUSE_PASSTHRU_hint) or
+ *  [window attribute](@ref GLFW_MOUSE_PASSTHRU_attrib).
+ */
+#define GLFW_MOUSE_PASSTHRU         0x0002000D
+
 /*! @brief Framebuffer bit depth hint.
  *
  *  Framebuffer bit depth [hint](@ref GLFW_RED_BITS).
@@ -3499,6 +3506,7 @@ GLFWAPI int glfwGetWindowAttrib(GLFWwindow* window, int attrib);
  *  [GLFW_FLOATING](@ref GLFW_FLOATING_attrib),
  *  [GLFW_AUTO_ICONIFY](@ref GLFW_AUTO_ICONIFY_attrib) and
  *  [GLFW_FOCUS_ON_SHOW](@ref GLFW_FOCUS_ON_SHOW_attrib).
+ *  [GLFW_MOUSE_PASSTHRU](@ref GLFW_MOUSE_PASSTHRU_attrib)
  *
  *  Some of these attributes are ignored for full screen windows.  The new
  *  value will take effect if the window is later made windowed.
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1356,6 +1356,14 @@ void _glfwPlatformSetWindowFloating(_GLFWwindow* window, GLFWbool enabled)
     } // autoreleasepool
 }
 
+void _glfwPlatformSetWindowMousePassthru(_GLFWwindow* window, GLFWbool enabled)
+{
+    window->mousePassthru = enabled;
+    @autoreleasepool {
+    [window->ns.object setIgnoresMouseEvents:enabled];
+    }
+}
+
 float _glfwPlatformGetWindowOpacity(_GLFWwindow* window)
 {
     @autoreleasepool {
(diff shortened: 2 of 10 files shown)


elmindreda:
Thank you for this PR! My proposed changes are now in the `mouse-passthrough` branch of this repo. I didn't want to modify your commit without asking so I added a `PASSTHRU` to `PASSTHROUGH` rename pass for now.

> Is `GLFW_MOUSE_PASSTHRU` defined in appropriate place and is it's value proper?

Yup, looks good! I modified the logic for the internal flag a little, to keep modifications of shared state inside shared code.

> `libXext.so.6` - hope that ".6" is pretty much same across all distributions. I may be wrong.
> CYGWIN - untested.
> Not sure if i got all `_glfw.x11.xshape` right.

The X11 path looks good and worked fine on X11 on both Linux and Cygwin. It needed a pre-fix because GLFW now loads all of Xlib dynamically, and there was an issue with not removing the custom input shape, but that was all.

I had to replace the Win32 implementation, though. While the `WM_NCHITTEST` method does what Dear Imgui docking needs it didn't match the passthrough behavior of other platforms and pass the input through. Instead the combination of `WS_EX_TRANSPARENT` and `WS_EX_LAYERED` is required, and for hidden windows also a call to `SetLayeredWindowAttributes`.

> Wayland while i think is implemented correctly - does not work.

It seems to work perfectly in Weston. I've tried opaque windows and the gears example, setting it at creation or after.

The macOS path also just worked.

elmindreda:
I tested the `mouse-passthrough` branch with the current `docking` branch of Dear Imgui today and it appears to work as expected, i.e. the same as the current Win32 workaround in the demo code.

rokups:
Thank you for looking into my PR.

> I didn't want to modify your commit without asking

Please do whatever you think is necessary :+1: 

> libXext.so.6 - hope that ".6" is pretty much same across all distributions. I may be wrong.

Google search sends mixed signals. Maybe it should be `libXext.so` instead?

elmindreda:
> Google search sends mixed signals. Maybe it should be `libXext.so` instead?

That form of the name is typically only available if the development package is installed and may point to a library with a different ABI.

ocornut:
Thanks @rokups and @elmindreda for looking into this!

> I had to replace the Win32 implementation, though. While the WM_NCHITTEST method does what Dear Imgui docking needs it didn't match the passthrough behavior of other platforms and pass the input through. Instead the combination of WS_EX_TRANSPARENT and WS_EX_LAYERED is required, and for hidden windows also a call to SetLayeredWindowAttributes.

One thought was coming to mind but then I realized it **wouldn't meaningfully affect our use case**. For the sake of completeneness I'm writing it down:

I haven't performed the tests yet, but I wonder if this might be altering performances of the swap/composition in any way. I guess we could do a quick test of running unthrottled with both settings and see if it makes any diffferences (worth testing with both variations of `DXGI_SWAP_EFFECT_DISCARD` and Win10 compositor friendly `DXGI_SWAP_EFFECT_FLIP_DISCARD` settings of the SwapChain). I will run those tests if I have time and report the info.

As Dear ImGui multi-viewports only temporarily enable this flags during dragging of a window the difference would likely not matter.

ocornut:
Results: Well, it turns out that just copying `_glfwPlatformSetWindowMousePassthrough()` into my Win32 main.cpp did the job.
I tested this in a DX11 setup with both `DXGI_SWAP_EFFECT_DISCARD` and `DXGI_SWAP_EFFECT_FLIP_DISCARD` with both Integrated Intel HD chip and Integrated Nvidia on same laptop and couldn't measure any meaningful difference, so you can nuke my thought above. At this point I can't pretend I understand how the `WS_EX_TRANSPARENT` and `WS_EX_LAYERED` flags behave, but I assume with the current settings they don't have any incidence on per-pixel compositing.

elmindreda:
Thank you for looking into it and posting the results!

PR #1496, opened 2019-06-07
Replace some tabs with spaces

Luflosi:
I found some tabs where there should be spaces for consistency.
This is my first contribution to GLFW, please let me know if I did something wrong.

PR #1569, opened 2019-10-01
Use the correct type for a struct member

Luflosi:
The `name` member in the `_GLFWmonitorWayland` struct is used in two places. It is assigned the value from a variable of type `uint32_t` and is compared to another variable of type `uint32_t`, so `name` should also have the same type.

### DIFF
--- a/src/wl_platform.h
+++ b/src/wl_platform.h
@@ -330,7 +330,7 @@ typedef struct _GLFWlibraryWayland
 typedef struct _GLFWmonitorWayland
 {
     struct wl_output*           output;
-    int                         name;
+    uint32_t                    name;
     int                         currentMode;
 
     int                         x;


Luflosi:
I added "Closes #1569." and rebased onto the latest master.

linkmauve:
Thanks, this PR is correct, I tested it just in case and LGTM!

I wonder why compilers don’t output any warning in this case though.

Luflosi:
> Thanks, this PR is correct, I tested it just in case and LGTM!

Awesome, thanks.

> I wonder why compilers don’t output any warning in this case though.

I think the warnings for this sort of thing need to be enabled manually.

> Note that you don’t need the “Closes #NNNN” in pull requests, they get closed automatically once we press the merge button.

I am aware of that. However in my first PR @elmindreda recommended to add this: https://github.com/glfw/glfw/pull/1496#issuecomment-502457468.

elmindreda:
Yup, it helps when working with the Git history. Thank you for adding it!

PR #1570, opened 2019-10-03
OS X EGL context api support with ANGLE

some00:
ANGLE'S EGL implementation expects the NativeWindowType to be a CALayer*.
https://github.com/google/angle/blob/master/src/libANGLE/renderer/gl/cgl/DisplayCGL.mm: DisplayCGL::isValidNativeWindow

Current GLFW implementation uses NSView*.
Since I couldn't find any alternative EGL implementations, I assumed that ANGLE's CALayer* choice is correct, and changed GLFW without checking the vendor.

To test I used the following code: https://gist.github.com/some00/1a5c9962e1efced46c26624e0b90d584


elmindreda:
Thank you for the PR! I wanted to solve this issue the way this PR does and always use a layer but it caused NSGL vsync to be disabled on older macOS versions.

This issue was instead solved with 91eebe922de06a3ed4ba6622686c436c7941aa63 in #1680.

PR #1572, opened 2019-10-06

Luflosi:
The `monitorCount` member in the `_GLFWlibrary` struct is of type `int`, so the `for` loop iterating over it should also use the type `int`.

Closes #1572.


Luflosi:
Rebased again.

elmindreda:
Merged as 7749aac50dd404b53ac5c1b35b7935b660ecdad7. Thank you!

PR #1574, opened 2019-10-07
add object library support for parent projects

svenevs:
- Transitive dependency management via `target_link_libraries` with
  object libraries introduced in CMake 3.12.
- Update a few minor elements related to switching to new CMake.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,16 @@ matrix:
           env:
               - BUILD_SHARED_LIBS=OFF
               - CFLAGS=-Werror
+
+install:
+    # Travis CI macOS image ships with CMake 3.11.3, need 3.12 or later.
+    # Avoid `brew`, takes too long.
+    - if [ "$TRAVIS_OS_NAME" = "osx" ]; then
+          virtualenv venv
+          source venv/bin/activate
+          pip install cmake
+      fi
+
 script:
     - if grep -Inr '\s$' src include docs tests examples CMake *.md .gitattributes .gitignore; then
           echo Trailing whitespace found, aborting;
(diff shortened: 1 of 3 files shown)


svenevs:
I'm so sorry!!! I meant to open a PR up against my fork's `master`, was kind of still in a testing phase here ... didn't realize opening PR tab on my fork would bring me here :/

1. Are you willing to entertain having an object library?  There's no way for parent projects to combine everything together into a superlib if they want, unless an object library in the child project is used (limitation of CMake, for many years now).  This PR aims to enable this, while leaving everything else the same (unless people see there is an object library underneath, they'll have no idea / no impact on them).
2. What are your thoughts about updating to CMake 3.12 minimum version (noting that debian stable is currently 3.13, ubuntu (I believe) now has rolling CMake releases, and users can also `pip install cmake`)?  Object libraries become much better in 3.12.

Note that if the answer to (1) is *yes*, but (2) is *no*, I can implement definition propagation using genex.  It's just not very pretty...

To avoid noise, please advise: open issue to have discussion?  Close this while I test on my own fork?  Keep it and add review?

elmindreda:
Thank you for the PR and for your patience.

CMake 3.12 is still a bit new to set as the minimum version but GLFW now uses the [ellipsis syntax](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) in `cmake_minimum_required` to offer somewhat better support for newer versions. If there are specific flags or policies from newer versions, I'll gladly add conditional support for them. That said, the current minimum of 3.1 is definitely too old and will be updated.

The addition of the optional CMake variable `GLFW_LIBRARY_TYPE` by d83849792b0cf3e40c402f299c867ac6e1fd0e20 will allow one to build GLFW as an object library by turning the `glfw` target itself into one. This only requires CMake 3.12 of those who choose to build GLFW that way.

Issue #1560, opened 2019-09-04
Centos error: ‘for’ loop initial declarations are only allowed in C99 mode 


byungheon-jeong:
When building, I get the following error:

`/home/b1jeong/glfw/src/x11_init.c: In function ‘hasUsableInputMethodStyle’:
/home/b1jeong/glfw/src/x11_init.c:340:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (unsigned int i = 0;  i < styles->count_styles;  i++)
     ^
/home/b1jeong/glfw/src/x11_init.c:340:5: note: use option -std=c99 or -std=gnu99 to compile your code
/home/b1jeong/glfw/src/x11_init.c: In function ‘getSupportedAtom’:
/home/b1jeong/glfw/src/x11_init.c:361:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (unsigned int i = 0;  i < atomCount;  i++)
     ^
make[2]: *** [src/CMakeFiles/glfw.dir/x11_init.c.o] Error 1
make[1]: *** [src/CMakeFiles/glfw.dir/all] Error 2
make: *** [all] Error 2
`

My gcc version:
`(base) [b1jeong@k8s-u200-01 build]$ gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
`
My g++ version:
` g++ --version
g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR 
`

tombsar:
Could you run `make VERBOSE=1` to see what flags are being passed to the compiler? Or `make -n` if that doesn't work.

byungheon-jeong:
`(base) [b1jeong@k8s-u200-01 build]$ make VERBOSE=1
/usr/local/bin/cmake -S/home/b1jeong/glfw -B/home/b1jeong/glfw/build --check-build-system CMakeFiles/Makefile.cmake 0
[...]

PR #1576, opened 2019-10-09
Added a CMake line to select C99 in the implementation

gracicot:
This tells CMake to select the C99 standard in a compiler/platform agnostic way.

It should fix issue #1560

Reopened from gracicot#1

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -94,12 +94,21 @@ endif()
 
 add_library(glfw ${glfw_SOURCES} ${glfw_HEADERS})
 set_target_properties(glfw PROPERTIES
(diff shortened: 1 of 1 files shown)


elmindreda:
Thank you for all the workaround stuff so this can be merged before bumping the minimum CMake version!

tombsar:
I think this fix only covered the main glfw library, not the tests. I'm still getting errors about C99 code in tests/glfwinfo.c.

gracicot:
@tombsar I'll open a new pull request for that. It should be similar  and as simple as this one. I'll try to check if there is no other targets to change too.

elmindreda:
Thank you!

PR #1590, opened 2019-10-30
Added WIP types for the joystick callbacks.

beoran:
This PR is a work in progress for @601. Do not merge, may not even compile yet. I already opened the PR simply to get early feedback on it.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1613,6 +1613,105 @@ typedef void (* GLFWmonitorfun)(GLFWmonitor*,int);
  */
 typedef void (* GLFWjoystickfun)(int,int);
 
+/*! @brief The function pointer type for joystick button callbacks.
+ *
+ *  This is the function pointer type for joystick button callbacks.  A joystick
+ *  button callback function has the following signature:
+ *  @code
+ *  void function_name(int jid, int button, int action)
+ *  @endcode
+ *
+ *  @param[in] jid The joystick that had a button pressed or released.
+ *  @param[in] button The [joystick button](@ref buttons) that was pressed or released.
+ *  @param[in] action `GLFW_PRESS` or `GLFW_RELEASE`.  Future
+ *  releases may add more actions.
+ *
+ *  @sa @ref input_joystick_button
+ *  @sa @ref glfwSetJoystickButonCallback
+ *
+ *  @since Added in version 3.4.
+ *  @ingroup input
+ */
+typedef void (* GLFWjoystickbuttonfun)(int,int,int);
+
+/*! @brief The function pointer type for joystick axis movement callbacks.
+ *
+ *  This is the function pointer type for joystick axis movement callbacks.  A joystick
+ *  axis movement callback function has the following signature:
+ *  @code
+ *  void function_name(int jid, int axis, float position)
+ *  @endcode
+ *
+ *  @param[in] jid The joystick that had an axis moved.
+ *  @param[in] axis The [joystick axis](@ref gamepad axes) that was moved.
+ *  @param[in] position A value between -1.0 and 1.0 that indicates the position of the axis.
+ *
+ *  @sa @ref input_gamepad_axis
+ *  @sa @ref glfwSetJoystickAxisCallback
+ *
+ *  @since Added in version 3.4.
+ *  @ingroup input
+ */
+typedef void (* GLFWjoystickaxisfun)(int,int,float);
+
+/*! @brief The function pointer type for joystick hat movement callbacks.
+ *
+ *  This is the function pointer type for joystick hat movement callbacks.  A joystick
+ *  hat movement callback function has the following signature:
+ *  @code
+ *  void function_name(int jid, int hat, int position)
+ *  @endcode
+ *
+ *  @param[in] jid The joystick that had an axis moved.
+ *  @param[in] hat The [joystick hat](@ref joystick hats) that was moved.
+ *  @param[in] position A value that indicates the position of the hat.
+ *  The position parameter is one of the following values:
+ *
+ *  Name                  | Value
+ *  ----                  | -----
+ *  `GLFW_HAT_CENTERED`   | 0
+ *  `GLFW_HAT_UP`         | 1
+ *  `GLFW_HAT_RIGHT`      | 2
+ *  `GLFW_HAT_DOWN`       | 4
+ *  `GLFW_HAT_LEFT`       | 8
+ *  `GLFW_HAT_RIGHT_UP`   | `GLFW_HAT_RIGHT` \| `GLFW_HAT_UP`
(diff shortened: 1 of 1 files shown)


beoran:
Now, the callbacks have been implemented, which was easier than I expected, since most code to do so was already in place. The documentation is there too, though it still needs some small fixes, I don't know Doxygen that well. Now, next is to write tests and examples.

beoran:
Unfortunately it doesn't work because in GLFW the joystick is actively polled. I will probably have to add something to poll joysticks as well.

beoran:
It now works fine on Linux, tested with tests/event.  Please, could someone volunteer to test on other platforms as well?

Blergenstein:
Looking at the events test on Windows, the joystick button callback gets called constantly, even when the joystick isn't being touched.  I used the joystick test to ensure that the joystick isn't doing anything, (the axes always stated 0 for their value).

beoran:
Thanks for testing this. It would be great if you could also debug this to find the cause, otherwise this will have to wait until I can get a Windows system myself.

Blergenstein:
It looks like the axes updater only fires when there's a change.  I rewrapped it and the wrapped the others to do the same but in a potentially more optimized way.  Not sure if the `_glfwInputGamepad` functions should be run or not if there weren't any changes, so that may be broken with the below changes.  I don't use/know git or github at all, so I don't know how to "properly" send you these changes.

```
// Notifies shared code of the new value of a joystick axis
//
void _glfwInputJoystickAxis(_GLFWjoystick* js, int axis, int value)
{
  //When the function needing this is called, it will create its own variable, so is this needed?
  //const int jid = (int) (js - _glfw.joysticks);

  if (js->axes[axis] != value)
    {
      js->axes[axis] = value;
      if(_glfw.callbacks.joystick_axis)
        _glfw.callbacks.joystick_axis((int) (js - _glfw.joysticks), axis, value);
      _glfwInputGamepad(js);
    };
}

// Notifies shared code of the new value of a joystick button
//
void _glfwInputJoystickButton(_GLFWjoystick* js, int button, char value)
{
  //When the function needing this is called, it will create its own variable, so is this needed?
  //const int jid = (int) (js - _glfw.joysticks);

  if(js->buttons[button] != value)
    {
      js->buttons[button] = value;
      if(_glfw.callbacks.joystick_button)
        _glfw.callbacks.joystick_button((int) (js - _glfw.joysticks), button, value);
      _glfwInputGamepad(js);
    };
}

// Notifies shared code of the new value of a joystick hat
//
void _glfwInputJoystickHat(_GLFWjoystick* js, int hat, char value)
{
  //When the function needing this is called, it will create its own variable, so is this needed?
  //const int jid = (int) (js - _glfw.joysticks);
  const int base = js->buttonCount + hat * 4;

  js->buttons[base + 0] = (value & 0x01) ? GLFW_PRESS : GLFW_RELEASE;
  js->buttons[base + 1] = (value & 0x02) ? GLFW_PRESS : GLFW_RELEASE;
  js->buttons[base + 2] = (value & 0x04) ? GLFW_PRESS : GLFW_RELEASE;
  js->buttons[base + 3] = (value & 0x08) ? GLFW_PRESS : GLFW_RELEASE;

  if(js->hats[hat] != value)
    {
      js->hats[hat] = value;
      if(_glfw.callbacks.joystick_hat)
        _glfw.callbacks.joystick_hat((int) (js - _glfw.joysticks), hat, value);
      _glfwInputGamepad(js);
    };
}

```

Blergenstein:
My changes fixed the issue with the events test, but I don't get any events in my program.  I'm still looking into it.

beoran:
Thanks for your efforts, I will add your modifications to my PR once everything works.

Blergenstein:
Okay, scratch that last comment from me.  I made a really bad whoopsie.  I'm slowly removing SDL2 from my program and forgot to add `glfwPollEvents();` to my video loop.  The callbacks looks like they're running really well right now.

beoran:
OK, then I'll integrate your changes as fast as my "copious" free time allows me to.

beoran:
@Blergenstein, I have finally integrated your changes, could you please test again and see if it works for you as well?

@elmindreda, sorry to ping you again, but could you please review this PR, or assign someone else to this PR? This is a useful feature that we would like to use downstream in Ebiten.

Blergenstein:
Sorry for the delay.  It looks good; although the is a slight catch that I noticed with the Events test.  You have to use `glfwPollEvents();` to be able to use the joystick callbacks.  If you don't, the input from the joystick seems very pokey, and doesn't always seem to catch all of the inputs.

Other than that, it seems to work well.

beoran:
This seems to be platform dependent, on Linux the glfwPollEvents is not
strictly necessary. But for portability I will update the examples and
document that glfwPollEvents is needed to get joystick events when I get
time to do this.
Now for someone to test this on OSX.

Blergenstein:
Good luck finding a Mac person, and thanks again for your work on this.

cmedinasoriano:
I was able to test this branch on MacOS but the joystick callbacks were ONLY sent while any of the other events were active like moving the mouse / clicking the buttons or while keyboard key was pressed.  If I just use the joystick alone, no event is sent with the exception of the connect and disconnect event.

beoran:
@cmedinasoriano  Are you using glfwPollEvents() correctly (once per frame in your main loop)? If not, that may explain what you are seeing.

cmedinasoriano:
@beoran Sorry I was using the events test provided as is.  It was calling the `glfwWaitEvents` instead of the `glfwPollEvents` and I assumed one was replacing the other.  I can confirm that after using `glfwPollEvents` every event callback is working properly.

beoran:
Interesting, that does mean that glfwWaitEvents still has errors on OSX, and probably on Windows as well, because it should call  glfwPollEvents internally. For the time being glfwPollEventsis a workaround, but I will have to fix this issue as well.

wormyrocks:
It seems like under Linux this patch calls glfwPollEvents twice, both in window.c and x11_window.c.
Is this intentional? I assume the correct behavior is for it to only be called in x11_window.c.
For what it's worth, I've had no problems with this patch on my own branch since I initially merged your commit back from October.

beoran:
Yes, this is intentional, IIRC it did not always work correctly. Perhaps it can be improved as well, though.

beoran:
I updated the PR. Any news on this? This PR has been open for 2 years.

Enthuin:
Year 3 has come and gone.  I updated this branch with the latest changes and resolved the conflicts.  @beoran take a look at your fork and maybe merge the PR changes or some of them if you don't agree with all changes.  Would've posted my own PR for this repo but this issue is already on the 3.4 milestone so this seemed a better way to contribute

beoran:
@Enthuin thanks, I will take a look at resolving the conflicts as soon as I can get to it.

PeteHaughie:
It is now June 2024. Is there any movement on this PR?

PR #1593, opened 2019-11-04
Set C standard at 99 for all test and example targets

gracicot:
This pull request set the C standard for all remaining targets that wasn't covered by #1576 

I used the CMake style that was present in the file, which was to use directory based command for the test's CMakeLists file.


elmindreda:
Looks good! The examples directory also needs this.

elmindreda:
Thank you!

PR #1605, opened 2019-12-08
CMakeLists.txt: add hint when libxi-dev is missing

pthom:
This is just a helper message when "X11_Xi_INCLUDE_PATH" is not found, because googling does not help very much to spot the missing package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,7 +227,7 @@ if (_GLFW_X11)
 
     # Check for XInput (modern HID input)
     if (NOT X11_Xi_INCLUDE_PATH)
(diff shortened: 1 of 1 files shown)


linkmauve:
This only applies to Debian though, ArchLinux calls this package libxi, Fedora libXi-devel, and other distributions probably use very different names too.

pthom:
May be the message could be "Hint: install libxi devel package" (this would have been sufficient for me)

elmindreda:
Ooh, good idea! Will add for all the X errors.

elmindreda:
Package name hints added with a43f54567773c38ee9dda4ab6b6a65fd58a67080. Thank you!

Wish we could have a per-distro package name table but don't have time to compile and maintain one.

PR #1607, opened 2019-12-08
Expose damage tracking in the API

linkmauve:
This adds two new functions to the public API, `glfwGetBufferAge()` and `glfwSwapBuffersWithDamage()`, mirroring the EGL extensions EGL_EXT_buffer_age and EGL_KHR_swap_buffers_with_damage, as well as a new `GLFWrect` struct used by the latter.

These two functions can help efficiency tremendously, especially due to a lower memory bandwidth usage, especially on embedded GPUs:
- `glfwGetBufferAge()` lets the application know when the current buffer was last written to, so that it can avoid a clear and redraw of areas which haven’t changed.
- `glfwSwapBuffersWithDamage()` lets the compositor know which areas it should redraw, so that it can avoid compositing areas which haven’t changed, if it itself tracks the buffer age of its buffers.

Here is a video of [PyTouhou](https://pytouhou.linkmauve.fr) [patched](https://linkmauve.fr/files/pytouhou-damage-tracking.patch) to use this new API: [touhou-damage](https://linkmauve.fr/files/touhou-damage.webm)

I also modified `tests/msaa.c` to use damage to swap buffers, so that you can test this feature.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -5656,6 +5656,7 @@ GLFWAPI void glfwSwapBuffers(GLFWwindow* window);
  *  @sa @ref buffer_swap
  *  @sa @ref glfwSwapBuffers
  *  @sa @ref glfwSwapInterval
+ *  @sa @ref glfwGetBufferAge
  *
  *  @since Added in version 3.4.
  *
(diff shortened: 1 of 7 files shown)


neurotok:
Hello linkmauve. Thank you for your contribution, Im super interested in this feature (performance implications for the immediate mode GUIs).  I manage to compile this branch and tested msaa example on two maschines (integrated intel gpu -mesa and nvidia gtx 770 - nvidia drivers) on two different linux distros. They both have necessary extensions but I don't see any differences with thoes spinning white rectangles. I see msaa working but no visual implications of this functions running. I even try to visual myself 8 areas that you compose.

areas

I assumed (x, y, w, h) and now Im even more confused. what i'm doing wrong ? :(

PR #786, opened 2016-06-15
Add EGLDevice backend to GLFW

kchandra22134:
This adds interface to graphics application to create EGLDisplay
directly over GPU device as enabled by EGLDevice extensions
	
Modifies in GLFW library to add egldevice-drm backend.
	
This includes modifying a few existing files and creating four new files which are
specific to the backend that we are adding. For eg. in our case these are
egldevice_platform.h, egldevice_init.c, egldevice_window.c, egldevice_monitor.c .
	
In the file "egl_context.c" change selects the EGL_SURFACE_TYPE config attribute value to
EGL_STREAM_BIT_KHR instead of EGL_WINDOW_BIT  and specify the window surface to the surface which we created using EGLStreams.
	
Other files "egl_context.h" and "internal.h" has minor changes to include the header files.
Also there are changes in CMake code files to include the newly created files.

Links to the EGL extensions used:

1. EGL_EXT_device_drm
https://www.khronos.org/registry/egl/extensions/EXT/EGL_EXT_device_drm.txt

2. EGL_KHR_stream
https://www.khronos.org/registry/egl/extensions/KHR/EGL_KHR_stream.txt

3. EGL_EXT_stream_consumer_egloutput
 https://www.khronos.org/registry/egl/extensions/EXT/EGL_EXT_stream_consumer_egloutput.txt

4. EGL_KHR_stream_producer_eglsurface
https://www.khronos.org/registry/egl/extensions/KHR/EGL_KHR_stream_producer_eglsurface.txt

PR #1608, opened 2019-12-09
EGL headless mode

cshesse:
Inspired by https://github.com/glfw/glfw/pull/786 this is a PR to support OSMesa-style headless rendering with EGL.

If you want to run an OpenGL program on a linux server you can run an X server with either hardware or software rendering.  Configuring software rendering isn't too bad but hardware rendering [can be a bit annoying](https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/tesla-product-literature/remote-viz-tesla-gpus.pdf) and in either case you need to manage an X server process.

To run without an X server, you can use osmesa or EGL.  OSMesa headless can be used by compiling GLFW with `-DGLFW_USE_OSMESA=ON` , but only supports software rendering.  For hardware rendering, [EGL can be used](https://github.com/NVIDIA-developer-blog/code-samples/blob/master/posts/egl_OpenGl_without_Xserver/tinyegl.cc), but the existing code doesn't seem to quite support this setup.

This PR is a proof of concept for this mode (it's just hacked into `egl_context.(h|c)`) but if there is interest I could clean it up.

### DIFF
--- a/src/egl_context.c
+++ b/src/egl_context.c
@@ -98,33 +98,6 @@ static GLFWbool chooseEGLConfig(const _GLFWctxconfig* ctxconfig,
                                 const _GLFWfbconfig* desired,
                                 EGLConfig* result)
 {
-
-#if defined(_GLFW_EGLHEADLESS)
-{
-    EGLint attribs[40];
-    int index = 0;
-        setAttrib(EGL_SURFACE_TYPE, EGL_PBUFFER_BIT);
-        setAttrib(EGL_BLUE_SIZE, 8);
-        setAttrib(EGL_GREEN_SIZE, 8);
-        setAttrib(EGL_RED_SIZE, 8);
-        setAttrib(EGL_DEPTH_SIZE, 8);
-        if (ctxconfig->client == GLFW_OPENGL_ES_API)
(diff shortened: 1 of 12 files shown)


src/egl_context.c
```diff
@@ -656,8 +732,12 @@ GLFWbool _glfwCreateContextEGL(_GLFWwindow* window,
             _GLFW_OPENGL_LIBRARY,
 #elif defined(_GLFW_WIN32)
 #elif defined(_GLFW_COCOA)
+#else
+#if defined(_GLFW_EGLHEADLESS)
+            "libOpenGL.so.0",
```
cshesse: It's not clear why this is necessary on my system.  It could be glvnd or else my system is misconfigured somehow.

vchuravy:
Having this would be fantastic! Especially for scientific applications rendering on a server.

kivutar:
This would also be useful for automated testing

aramirezreyes:
Oh, this would be nice :(

xunshengliuyin (approved):

xiaohuihuige:
太好了，爱你

PR #1621, opened 2020-01-12
Update to GLFW 12/01/2020

JonLangfordUK:


intrigus:
Huh?
What is this aboutq?

elmindreda:
I won't merge this, even in a fixed form. Maintaining more than one build system is a constant hassle and currently CMake is the better choice.

PR #1631, opened 2020-01-22
Add vcpkg installation instructions

JackBoosY:
glfw is available as a port in vcpkg, a C++ library manager that simplifies installation for glfw and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build glfw, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/glfw3/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,17 @@ available for all supported compilers.
 See the [compilation guide](http://www.glfw.org/docs/latest/compile.html) for
 more information about how to compile GLFW yourself.
 
+# Package Managers
+
+You can download and install glfw using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install glfw
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,6 +58,17 @@ available for all supported compilers.
...
 
+## Package Managers
+
+You can download and install glfw using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install glfw
```
aloucks:
```
$ ./vcpkg install glfw
Error: Cannot find definition for package `glfw`.
```

However, `glfw3` works
```
$ ./vcpkg install glfw3
The following packages will be built and installed:
    glfw3[core]:x64-windows-static
...
```

JackBoosY: Sorry, fixed.

PR #1642, opened 2020-02-08
check available Wayland protocols

christianrauch:
Check the available protocol definitions when generating bindings and set `HAS_PROTO_<name>` for a compile-time check.


elmindreda:
Thank you for the PR!

I don't think this is the right approach for GLFW to use. If we end up using protocols so new that distributions don't yet have them, I think it would be better to bundle the protocol definition until the packages have caught up.

PR #1644, opened 2020-02-09
removed unnecessary initialization of values with zero

iamAdrianIusca:
xpos and ypos are assigned with a value inside _glfwPlatformGetMonitorPos


elmindreda:
Thank you for the PR but I will not merge this.

The shared function sets those because the API guarantees that out parameters will have defined values even if an error occurs, including `GLFW_NOT_INITIALIZED`. It is also a service to platform functions so they don't all have to duplicate this work.

iamAdrianIusca:
Ok, thanks for your time.

PR #1673, opened 2020-04-15
Correct sign in mat4x4_rotate_Y

gre-42:
Signs were incorrect in mat4x4_rotate_Y.


tombsar:
Good spot! I agree; the Y rotation is inconsistent with X and Z, and the signs should be swapped. (see https://en.wikipedia.org/wiki/Rotation_matrix#Basic_rotations).

I believe this code is taken from the repository here: https://github.com/datenwolf/linmath.h/blob/master/linmath.h#L248 so it would be great if you could submit a PR there too.

gre-42:
I just submitted a PR in the source repository also.

tombsar:
NB: This change was accepted over at the source repository (https://github.com/datenwolf/linmath.h/pull/39)

PR #1678, opened 2020-04-18
Gamepad rumble effects

MDZhB:
This adds a function to control rumble effects on xbox-like gamepads, partially addressing #57. It also adds two sliders to the joystick test app, which allow the user to set each rumble motor's intensity. Some notes:

1. There is currently no MacOS support. I intend to implement this once I get my hands on an Apple machine. Linux and Windows work fine.
2. There is no fallback for DirectInput devices. Although all of the devices available to me support vibration effects, `IDirectInputDevice8_EnumEffects` produced no effect types; consequently, I couldn't test my work. If anybody can recommend a cheap pad with vibration effects that are available through DirectInput, I'd be happy to buy one and add some code to cover older devices.
3. The sole change to the API is a new method, `int glfwSetJoystickRumble(int jid, float slowMotorIntensity, float fastMotorIntensity)`, which returns `GLFW_FALSE` if it was unable to make the gamepad rumble for any reason. @elmindreda's comment on #57 regarding a more complex haptic feedback API got no response, so I assume this is sufficient in the maintainer's/community's view. 

Right now I'm looking for some feedback on my code, since I'm not really a C person. I'll add all the missing documentation (changelog, guide, etc.) once my work is good enough to merge.

### DIFF
--- a/src/linux_joystick.h
+++ b/src/linux_joystick.h
@@ -43,6 +43,7 @@ typedef struct _GLFWjoystickLinux
     int                     absMap[ABS_CNT];
     struct input_absinfo    absInfo[ABS_CNT];
     int                     hats[4][2];
+    struct ff_effect        *rumble;
 } _GLFWjoystickLinux;
 
 // Linux-specific joystick API data
--- a/deps/mingw/xinput.h
+++ b/deps/mingw/xinput.h
@@ -182,8 +182,8 @@ typedef struct _XINPUT_STATE {
  */
 
 typedef struct _XINPUT_VIBRATION {
-    WORD wLeftMotorIntensity;
-    WORD wRightMotorIntensity;
+    WORD wLeftMotorSpeed;
+    WORD wRightMotorSpeed;
 } XINPUT_VIBRATION, *PXINPUT_VIBRATION;
 
 /*
--- a/deps/mingw/xinput.h
+++ b/deps/mingw/xinput.h
@@ -182,8 +182,8 @@ typedef struct _XINPUT_STATE {
  */
 
 typedef struct _XINPUT_VIBRATION {
-    WORD wLeftMotorSpeed;
-    WORD wRightMotorSpeed;
+    WORD wLeftMotorIntensity;
+    WORD wRightMotorIntensity;
 } XINPUT_VIBRATION, *PXINPUT_VIBRATION;
 
 /*
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -5348,16 +5348,18 @@ GLFWAPI const char* glfwGetGamepadName(int jid);
  */
 GLFWAPI int glfwGetGamepadState(int jid, GLFWgamepadstate* state);
 
-/*! @brief Sets the specified gamepad's vibration effect intensity.
+/*! @brief Sets the intensity of an Xbox-like gamepad's rumble effect.
  *  
- *  This function sends vibration data to the specified Xbox-like gamepad. 
+ *  This function sends vibration data to gamepads that implement haptic
+ *  feedback effects using two vibration motors: a low-frequency motor, and
+ *  a high-frequency motor.
  *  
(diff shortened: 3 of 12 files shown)


JohnnytheRocket:
I tried this out with a PS4 controller, and whereas all the expected GLFW function remained I only got GLFW_FALSE  returned by glfwSetJoystickRumble, and no rumble.

I am running Linux and the controller was correctly identified with the expected button mapping. What type of controllers have successfully been tested?

MDZhB:
> I tried this out with a PS4 controller, and whereas all the expected GLFW function remained I only got GLFW_FALSE returned by glfwSetJoystickRumble, and no rumble.
> 
> I am running Linux and the controller was correctly identified with the expected button mapping. What type of controllers have successfully been tested?
> 
> **EDIT**
> It turns out that the force feedback-drivers were not activated in my kernel. After activating them the rumble works as expected. However, I think it would be useful to be able to set the duration of the rumble in the function call, or at least in a setup stage.

This code makes use of the XInput API on Windows, which does not support rumble duration. I'd like to stay away from adding machinery for this. My reasons are twofold:

1. it feels like a slippery slope toward cumbersome, DirectInput style effects (with parameters for duration, envelopes, etc.);
2. the application can achieve the same functionality by updating the rumble intensity in every frame where it is needed.

The current approach exposes gamepad rumble in a way that is simple and still allows for the implementation of "fire and forget" effects in the application code, if that is the programmer's preference.

JohnnytheRocket:
> The current approach exposes gamepad rumble in a way that is simple and still allows for the implementation of "fire and forget" effects in the application code, if that is the programmer's preference.

After some more experience working with it, I agree with your assessment. Good work with this, I at least appreciate it a lot!

stohrendorf:
What are the blockers to get this implemented/merged? This PR was opened quite some time ago, resolving an issue opened nearly a decade ago, and I haven't seen any comment indicating this can't be merged. The changes seem small enough to not consume too much time to review, and they are an optional feature posing no impact on existing functionality, so people can just ignore this feature if (for some reason) it migh cause problems.

wheremyfoodat:
@stohrendorf I'm guessing the lack of MacOS support is the blocker, but yes it would be incredibly useful to have it merged.

HughPH:
OpenGL is 'deprecated' in MacOS since 10.14, so I wouldn't wait on MacOS support for any changes to GLFW.

sauron65:
@HughPH Vulkan/MoltenVK exist. People also use GLFW for non-cross-platform APIs like D3D and Metal

asivitz:
The right approach on MacOS is probably to first move gamepad handling over to the newer Game Controller framework (https://developer.apple.com/documentation/gamecontroller?language=objc)

src/linux_joystick.h
```diff
@@ -43,6 +43,7 @@ typedef struct _GLFWjoystickLinux
     int                     absMap[ABS_CNT];
     struct input_absinfo    absInfo[ABS_CNT];
     int                     hats[4][2];
+    struct ff_effect        *rumble;
```
sir-pinecone, line 46: Had a look over the code and I don't see why this needs to be allocated with malloc. A pointer to the joystick is passed to `_glfwPlatformSetJoystickRumble` so the call to `ioctl` can just take the address of `linjs->rumble` and it removes some allocation code. Or am I missing something?

sir-pinecone:
Just did a quick test on my Windows 10 PC. Using the joystick test program, I was able to activate the rumble on an Xbox 360 wired controller and an Xbox One wireless connected with a USB cable. If I recall, Playstation controller support on Windows is a bit of a mess and I don't think you can use rumble without some additional driver support. Nonetheless I gave them a try. My Dualshock 3 is recognized as a PS3 controller but there's no input data and writing rumble data does nothing. The Dualshock 4 is recognized as `Wireless Controller`; it has input data, but rumble doesn't work.

LeandroSQ:
Is this archived?

kakesinfo:
Hello, I was wondering if this PR is still in the work ? Shall we recreate one ? GLFW must have this feature !

PR #1680, opened 2020-04-19
Use CALayer instead of NSView for EGLNativeWindowType

### LINKED ISSUE #1169: Failing to use ANGLE EGL
for osx, i compiled google angle.
	glfwWindowHint(GLFW_CONTEXT_CREATION_API, GLFW_EGL_CONTEXT_API);
	glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 2);
	glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 0);
	glfwWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_ES_API); 

i have this error: GLFW Error: Cocoa: EGL not available

Kangz:
The only two EGL implementations on macOS are Swiftshader and ANGLE.
While Swiftshader supports both `NSView` and `CALayer` as
`EGLNativeWindowType`, ANGLE supports only `CALayer`. Furthermore
Swiftshader's OpenGL ES frontend is deprecated in favor of using ANGLE
on top of Swiftshader's Vulkan frontend.

This means that on macOS `EGLNativeWindowType` should be a `CALayer` for
compatibility with ANGLE.

Fixes #1169

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -906,6 +906,11 @@ int _glfwPlatformCreateWindow(_GLFWwindow* window,
         }
         else if (ctxconfig->source == GLFW_EGL_CONTEXT_API)
         {
+            // EGL implementation on macOS use CALayer* EGLNativeWindowType so we
+            // need to get the layer for EGL window surface creation.
+            [window->ns.view setWantsLayer:YES];
+            window->ns.layer = [window->ns.view layer];
+
             if (!_glfwInitEGL())
                 return GLFW_FALSE;
             if (!_glfwCreateContextEGL(window, ctxconfig, fbconfig))
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ information on what to include when reporting a bug.
  - [Cocoa] Bugfix: Undecorated windows could not be iconified on recent macOS
  - [Cocoa] Bugfix: Touching event queue from secondary thread before main thread
    would abort (#1649)
+ - [Cocoa] Use `CALayer` instead of `NSView` for `EGLNativeWindowType` (#1169)
  - [X11] Bugfix: The CMake files did not check for the XInput headers (#1480)
  - [X11] Bugfix: Key names were not updated when the keyboard layout changed
    (#1462,#1528)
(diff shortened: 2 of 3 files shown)


elmindreda:
I'm a little concerned that our content can now be in all three states: non-layer-backed by default on pre-Mojave, layer-backed on Mojave and later or when using EGL, or layer-hosting when using Vulkan.

I tested making layer-backed the default state on pre-Mojave systems, but that disabled vsync for NSGL contexts for some reason I have yet to discover.

Kangz:
If I understand correctly the concern is that when using EGL on pre-Mojave, we could have the same VSync issue that the non-EGL path had?

elmindreda:
> If I understand correctly the concern is that when using EGL on pre-Mojave, we could have the same VSync issue that the non-EGL path had?

That seems to be unavoidable. I'm just vaguely worried, especially after the silent NSGL breakage in Mojave and what is probably more of the kind in the future.

I'll gladly merge this in the meantime.

src/cocoa_window.m
```diff
@@ -854,6 +855,17 @@ static GLFWbool createNativeWindow(_GLFWwindow* window,
...
+    // to get the layer for EGL window surface creation.
+    if (ctxconfig->source == GLFW_EGL_CONTEXT_API)
+    {
+        [window->ns.view setWantsLayer:YES];
+        window->ns.layer = [window->ns.view layer];
+
+        if (window->ns.retina)
+            [window->ns.layer setContentsScale:[window->ns.object backingScaleFactor]];
+    }
+
```
elmindreda, line 868: I would prefer this all be in `_glfwPlatformCreateWindow` after window creation, where we have other context creation API specific code.

src/cocoa_window.m
```diff
@@ -854,6 +855,17 @@ static GLFWbool createNativeWindow(_GLFWwindow* window,
...
 
+    // EGL implementation on macOS use CALayer* EGLNativeWindowType so we need
+    // to get the layer for EGL window surface creation.
+    if (ctxconfig->source == GLFW_EGL_CONTEXT_API)
+    {
+        [window->ns.view setWantsLayer:YES];
+        window->ns.layer = [window->ns.view layer];
+
+        if (window->ns.retina)
+            [window->ns.layer setContentsScale:[window->ns.object backingScaleFactor]];
```
elmindreda, line 866: Did the layer not pick up the contents scale from the view without this?

Kangz: It does, removed.

src/cocoa_window.m
Kangz: Done

Kangz:
Addressed all comments and the PR is much simpler now, PTAL again!

PR #1691, opened 2020-05-08
Don’t let the driver guess the platform

linkmauve:
EGL_EXT_platform_base, and later EGL 1.5, introduced two new functions
which take a platform enum instead of guessing based on the shape of the
passed pointer data, which is incredibly error-prone.

This commit makes use of them when they’re exposed by the driver, and
falls back to the old insecure API when not.


elmindreda:
Awesome, will review!

PR #1693, opened 2020-05-09
proper decorations with title and window buttons

### LINKED ISSUE #1639: Proper window decorations via libdecor
I am working on `libdecoration`, a project originated by Jonas Ådahl. I implemented a ["proper" title bar](https://gitlab.gnome.org/jadahl/libdecoration/merge_requests/2), which is still under review for merging this upstream.

I was wondering if there is interest in using this library for GLFW to implement Wayland client-side window decorations with title, min/max/close buttons and shadows to resizing windows:
libdecoration_cairo
and replace the older `viewporter`-based decorations that are missing buttons and shadows.

christianrauch:
This adds [`libdecor`](https://gitlab.gnome.org/jadahl/libdecor) for proper client-side window decoration support:
glfw_decorations

(cc @jadahl)

~~This PR currently checks out a development branch that contains some PRs that haven't been merged upstream yet.~~

For now, I am mainly interested to get some feedback on the decoration requirements in GLFW and see if `libdecor` fits your needs. Once the remaining PRs have been merged upstream, I would appreciate if GLFW could gain proper decoration support by merging this.

I also included some minor changes to remove glitches during resizing.

Fixes #1639 .

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -56,6 +56,7 @@ static _GLFWwindow* findWindowFromDecorationSurface(struct wl_surface* surface,
     _GLFWwindow* window = _glfw.windowListHead;
     if (!which)
         which = &focus;
+#ifndef WITH_DECORATION
     while (window)
     {
         if (surface == window->wl.decorations.top.surface)
@@ -80,6 +81,7 @@ static _GLFWwindow* findWindowFromDecorationSurface(struct wl_surface* surface,
         }
         window = window->next;
     }
+#endif
     return window;
 }
 
@@ -103,7 +105,10 @@ static void pointerHandleEnter(void* data,
             return;
     }
 
+#ifndef WITH_DECORATION
     window->wl.decorations.focus = focus;
+#endif
+
     _glfw.wl.serial = serial;
     _glfw.wl.pointerFocus = window;
 
@@ -194,6 +199,7 @@ static void pointerHandleMotion(void* data,
     x = wl_fixed_to_double(sx);
     y = wl_fixed_to_double(sy);
 
+#ifndef WITH_DECORATION
     switch (window->wl.decorations.focus)
     {
         case mainWindow:
@@ -231,6 +237,7 @@ static void pointerHandleMotion(void* data,
         default:
             assert(0);
     }
+#endif
     if (_glfw.wl.cursorPreviousName != cursorName)
         setCursor(window, cursorName);
 }
@@ -248,6 +255,7 @@ static void pointerHandleButton(void* data,
 
     if (!window)
         return;
+#ifndef WITH_DECORATION
     if (button == BTN_LEFT)
     {
         switch (window->wl.decorations.focus)
@@ -306,6 +314,7 @@ static void pointerHandleButton(void* data,
     // Don’t pass the button to the user if it was related to a decoration.
     if (window->wl.decorations.focus != mainWindow)
         return;
+#endif
 
     _glfw.wl.serial = serial;
 
@@ -840,11 +849,13 @@ static void registryHandleGlobal(void* data,
                              &zxdg_decoration_manager_v1_interface,
                              1);
     }
+#ifndef WITH_DECORATION
     else if (strcmp(interface, "wp_viewporter") == 0)
     {
         _glfw.wl.viewporter =
             wl_registry_bind(registry, name, &wp_viewporter_interface, 1);
     }
+#endif
     else if (strcmp(interface, "zwp_relative_pointer_manager_v1") == 0)
     {
         _glfw.wl.relativePointerManager =
@@ -1249,8 +1260,10 @@ void _glfwPlatformTerminate(void)
         wl_compositor_destroy(_glfw.wl.compositor);
     if (_glfw.wl.shm)
         wl_shm_destroy(_glfw.wl.shm);
+#ifndef WITH_DECORATION
     if (_glfw.wl.viewporter)
         wp_viewporter_destroy(_glfw.wl.viewporter);
+#endif
     if (_glfw.wl.decorationManager)
         zxdg_decoration_manager_v1_destroy(_glfw.wl.decorationManager);
     if (_glfw.wl.wmBase)
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -44,11 +44,14 @@
 #include <wayland-client.h>
 
 
+const char *proxy_tag = "glfw-proxy";
+
 static inline int min(int n1, int n2)
 {
     return n1 < n2 ? n1 : n2;
 }
 
+#ifndef WITH_DECORATION
 static _GLFWwindow* findWindowFromDecorationSurface(struct wl_surface* surface,
                                                     int* which)
 {
@@ -56,7 +59,6 @@ static _GLFWwindow* findWindowFromDecorationSurface(struct wl_surface* surface,
     _GLFWwindow* window = _glfw.windowListHead;
     if (!which)
         which = &focus;
-#ifndef WITH_DECORATION
     while (window)
     {
         if (surface == window->wl.decorations.top.surface)
@@ -81,9 +83,22 @@ static _GLFWwindow* findWindowFromDecorationSurface(struct wl_surface* surface,
         }
         window = window->next;
     }
-#endif
     return window;
 }
+#endif
+
+#ifdef WITH_DECORATION
+void decoration_error(struct libdecor *context,
+                      enum libdecor_error error,
+                      const char *message)
+{
+    _glfwInputError(GLFW_PLATFORM_ERROR, "Wayland: Caught error (%d): %s\n", error, message);
+}
+
+static struct libdecor_interface decoration_interface = {
+    decoration_error,
+};
+#endif
 
 static void pointerHandleEnter(void* data,
                                struct wl_pointer* pointer,
@@ -96,8 +111,16 @@ static void pointerHandleEnter(void* data,
     if (!surface)
         return;
 
-    int focus = 0;
+    if (wl_proxy_get_tag((struct wl_proxy *) surface) != &proxy_tag)
+        return;
+
     _GLFWwindow* window = wl_surface_get_user_data(surface);
+
+#ifdef WITH_DECORATION
+    if (surface != window->wl.surface)
+        return;
+#else
+    int focus = 0;
     if (!window)
     {
         window = findWindowFromDecorationSurface(surface, &focus);
@@ -105,7 +128,6 @@ static void pointerHandleEnter(void* data,
             return;
     }
 
-#ifndef WITH_DECORATION
     window->wl.decorations.focus = focus;
 #endif
 
@@ -196,10 +218,16 @@ static void pointerHandleMotion(void* data,
 
     if (window->cursorMode == GLFW_CURSOR_DISABLED)
         return;
+
     x = wl_fixed_to_double(sx);
     y = wl_fixed_to_double(sy);
 
-#ifndef WITH_DECORATION
+#ifdef WITH_DECORATION
+    window->wl.cursorPosX = x;
+    window->wl.cursorPosY = y;
+    _glfwInputCursorPos(window, x, y);
+    _glfw.wl.cursorPreviousName = NULL;
+#else
     switch (window->wl.decorations.focus)
     {
         case mainWindow:
@@ -251,11 +279,11 @@ static void pointerHandleButton(void* data,
 {
     _GLFWwindow* window = _glfw.wl.pointerFocus;
     int glfwButton;
-    uint32_t edges = XDG_TOPLEVEL_RESIZE_EDGE_NONE;
 
     if (!window)
         return;
 #ifndef WITH_DECORATION
+    uint32_t edges = XDG_TOPLEVEL_RESIZE_EDGE_NONE;
     if (button == BTN_LEFT)
     {
         switch (window->wl.decorations.focus)
@@ -477,13 +505,21 @@ static void keyboardHandleEnter(void* data,
     if (!surface)
         return;
 
+    if (wl_proxy_get_tag((struct wl_proxy *) surface) != &proxy_tag)
+        return;
+
     _GLFWwindow* window = wl_surface_get_user_data(surface);
+#ifdef WITH_DECORATION
+    if (surface != window->wl.surface)
+        return;
+#else
     if (!window)
     {
         window = findWindowFromDecorationSurface(surface, NULL);
         if (!window)
             return;
     }
+#endif
 
     _glfw.wl.serial = serial;
     _glfw.wl.keyboardFocus = window;
@@ -778,6 +814,7 @@ static const struct wl_data_device_listener dataDeviceListener = {
     dataDeviceHandleSelection,
 };
 
+#ifndef WITH_DECORATION
 static void wmBaseHandlePing(void* data,
                              struct xdg_wm_base* wmBase,
                              uint32_t serial)
@@ -788,6 +825,7 @@ static void wmBaseHandlePing(void* data,
 static const struct xdg_wm_base_listener wmBaseListener = {
     wmBaseHandlePing
 };
+#endif
 
 static void registryHandleGlobal(void* data,
                                  struct wl_registry* registry,
@@ -836,6 +874,7 @@ static void registryHandleGlobal(void* data,
                                  &wl_data_device_manager_interface, 1);
         }
     }
+#ifndef WITH_DECORATION
     else if (strcmp(interface, "xdg_wm_base") == 0)
     {
         _glfw.wl.wmBase =
@@ -849,7 +888,6 @@ static void registryHandleGlobal(void* data,
                              &zxdg_decoration_manager_v1_interface,
                              1);
     }
-#ifndef WITH_DECORATION
     else if (strcmp(interface, "wp_viewporter") == 0)
     {
         _glfw.wl.viewporter =
@@ -1163,12 +1201,14 @@ int _glfwPlatformInit(void)
     if (_glfw.wl.seatVersion >= 4)
         _glfw.wl.timerfd = timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC);
 
+#ifndef WITH_DECORATION
     if (!_glfw.wl.wmBase)
     {
         _glfwInputError(GLFW_PLATFORM_ERROR,
                         "Wayland: Failed to find xdg-shell in your compositor");
         return GLFW_FALSE;
     }
+#endif
 
     if (_glfw.wl.pointer && _glfw.wl.shm)
     {
@@ -1213,6 +1253,10 @@ int _glfwPlatformInit(void)
         }
         _glfw.wl.clipboardSize = 4096;
     }
+
+#ifdef WITH_DECORATION
+    _glfw.wl.csd_context = libdecor_new(_glfw.wl.display, &decoration_interface);
+#endif
 
     return GLFW_TRUE;
 }
@@ -1260,14 +1304,16 @@ void _glfwPlatformTerminate(void)
         wl_compositor_destroy(_glfw.wl.compositor);
     if (_glfw.wl.shm)
         wl_shm_destroy(_glfw.wl.shm);
-#ifndef WITH_DECORATION
+#ifdef WITH_DECORATION
+    libdecor_unref(_glfw.wl.csd_context);
+#else
     if (_glfw.wl.viewporter)
         wp_viewporter_destroy(_glfw.wl.viewporter);
-#endif
     if (_glfw.wl.decorationManager)
         zxdg_decoration_manager_v1_destroy(_glfw.wl.decorationManager);
     if (_glfw.wl.wmBase)
         xdg_wm_base_destroy(_glfw.wl.wmBase);
+#endif
     if (_glfw.wl.dataSource)
         wl_data_source_destroy(_glfw.wl.dataSource);
     if (_glfw.wl.dataDevice)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,12 @@ jobs:
               run: |
                   sudo apt update
                   sudo apt install wayland-protocols libwayland-dev libxkbcommon-dev
+                  sudo apt install meson libpango1.0-dev
+                  git clone --depth 1 https://gitlab.gnome.org/jadahl/libdecor.git --branch 0.1.0
+                  cd libdecor
+                  meson build --buildtype release -Ddemo=false -Ddbus=disabled
+                  ninja -C build
+                  sudo meson install -C build
 
             - name: Configure static library
               run: cmake -S . -B build-static -D GLFW_USE_WAYLAND=ON
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,25 @@ if (GLFW_BUILD_DOCS)
     find_package(Doxygen)
 endif()
 
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} \
+        -fsanitize=address \
+        -fsanitize=bool \
+        -fsanitize=bounds \
+        -fsanitize=enum \
+        -fsanitize=float-cast-overflow \
+        -fsanitize=float-divide-by-zero \
+        -fsanitize=nonnull-attribute \
+        -fsanitize=returns-nonnull-attribute \
+        -fsanitize=signed-integer-overflow \
+        -fsanitize=undefined \
+        -fsanitize=vla-bound \
+        -fno-sanitize=alignment \
+        -fsanitize=leak \
+        -fsanitize=object-size \
+    ")
+endif()
+
 #--------------------------------------------------------------------
 # Set compiler specific flags
 #--------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,8 @@ cmake_dependent_option(GLFW_USE_WAYLAND "Use Wayland for window creation" OFF
                        "UNIX;NOT APPLE" OFF)
 cmake_dependent_option(USE_MSVC_RUNTIME_LIBRARY_DLL "Use MSVC runtime library DLL" ON
                        "MSVC" OFF)
+cmake_dependent_option(GLFW_USE_LIBDECOR "use libdecor for client-side window decorations" ON
+                       "GLFW_USE_WAYLAND" OFF)
 
 if (BUILD_SHARED_LIBS AND UNIX)
     # On Unix-like systems, shared libraries can use the soname system.
@@ -224,6 +226,14 @@ if (_GLFW_WAYLAND)
 
     list(APPEND glfw_INCLUDE_DIRS "${Wayland_INCLUDE_DIRS}")
     list(APPEND glfw_LIBRARIES "${Wayland_LINK_LIBRARIES}")
+
+    if (GLFW_USE_LIBDECOR)
+        pkg_check_modules(libdecor REQUIRED libdecor-0)
+        list(APPEND glfw_PKG_DEPS "libdecor-0")
+        list(APPEND glfw_INCLUDE_DIRS "${libdecor_INCLUDE_DIRS}")
+        list(APPEND glfw_LIBRARIES "${libdecor_LINK_LIBRARIES}")
+        add_definitions(-DWITH_DECORATION)
+    endif()
 
     include(CheckIncludeFiles)
     include(CheckFunctionExists)
--- a/src/window.c
+++ b/src/window.c
@@ -279,7 +279,7 @@ void glfwDefaultWindowHints(void)
     _glfw.hints.framebuffer.redBits      = 8;
     _glfw.hints.framebuffer.greenBits    = 8;
     _glfw.hints.framebuffer.blueBits     = 8;
-    _glfw.hints.framebuffer.alphaBits    = 8;
+    _glfw.hints.framebuffer.alphaBits    = 0;
     _glfw.hints.framebuffer.depthBits    = 24;
     _glfw.hints.framebuffer.stencilBits  = 8;
     _glfw.hints.framebuffer.doublebuffer = GLFW_TRUE;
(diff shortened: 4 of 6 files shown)


christianrauch:
> Why the switch to focal?

`libdecoration` requires newer versions of some libraries and meson. Also, `xenial` is only supported for another year.

christianrauch:
@elmindreda @linkmauve Is the client-side decoration support in this PR something you would be interested in? With this PR the Wayland support gets in much better shape as it allows proper handling of windows and removes remaining visual glitches during resizing.

The window and decoration visibility toggling support has been deactivated for now. They are always visible and cannot be hidden. Is run-time window or decoration hiding something that is used in practice?

christianrauch:
@elmindreda @linkmauve Any interest in proper Wayland client-side window decorations for GLFW?

ninja-:
hey, nice work here :) I tried your branch of glfw with Minecraft.
It seems to have fixed the weird window size glitch that broke resizing as well, and the fullscreen switch crash :)
But looks like mouse input is currently broken :/ I can't get click anything within game

ninja-:
also I noticed the cursor doesn't go back to normal after resize

ninja-:
I just finished debugging the cursor issue while you pushed the fix lol! I got the same fix, but what a pain it was to find it...

ninja-:
Cursor position seems to be incorrent though, maybe we need to offset the decoration size?

ninja-:
The more I resize the window the more broken the cursor position is

ninja-:
well. as I try left top corner (0,0) and down right corner (width, height) it seems to match debugging pointer events, at least matches what game thinks the current width or height is. maybe the problem is with passing width/height on resize?

ninja-:
so, while I got no closer to finding a solution, I just spotted another bug: sometimes while game locks the pointer, the cursor may not disappear from screen

src/wl_init.c
```diff
@@ -94,8 +109,16 @@ static void pointerHandleEnter(void* data,
     if (!surface)
         return;
 
-    int focus = 0;
+    if (strcmp(*wl_proxy_get_tag((struct wl_proxy *) surface), proxy_tag)!=0)
```
ninja-, line 112: this should be null checked first

ninja-:
Ok I FINALLY fixed the cursor being out of sync!
Needs call to `        _glfwInputWindowSize(window, scaledWidth, scaledHeight);` in resizeWindow!

src/wl_init.c
christianrauch: Actually it is supposed to work like in the example of the commit https://cgit.freedesktop.org/wayland/wayland/commit/?id=493ab79bd2cdf8f9f12fb9e5522200dc668b85e0, i.e. just by comparing the addresses.

christianrauch: I changed the linking of `proxy_tag`. This now works by comparing pointers and we do not have to check for NULL manually.

christianrauch:
> Ok I FINALLY fixed the cursor being out of sync!
> Needs call to ` _glfwInputWindowSize(window, scaledWidth, scaledHeight);` in resizeWindow!

Cool. Can you rebase (I force-pushed to enable the `proxy_tag` pointer comparison) and send a pull request to my branch?

ninja-:
It's a one liner but I can if you wish ;)
any ideas on the pointer locking problem?

christianrauch:
> It's a one liner but I can if you wish ;)
> any ideas on the pointer locking problem?

Actually, I do not see what the problem is with the cursor position offset. The `cursor` example seems to report the correct positions.

Can you provide minimal source examples that replicate these two bugs and state what the expected and actual outcome is? Ideally, you could demonstrate these bugs with GLFW's built-in examples or tests.

ninja-:
> Can you provide minimal source examples that replicate these two bugs and state what the expected and actual outcome is? Ideally, you could demonstrate these bugs with GLFW's built-in examples or tests.

You can see _glfwInputWindowSize being called in code for older decorations but nowhere in the new code...
Cursor positions are correct, but minecraft uses window size for calculating offsets what is where relative to the cursor.
So that's fixed by calling it during resizeWindow.

Regarding cursor locking problem, I am not sure on the exact repro, spamming cursor lock/unlock should trigger it.
Cursor just stays on the screen when it's locked. Most of the time.

christianrauch:
> You can see _glfwInputWindowSize being called in code for older decorations but nowhere in the new code...

I saw this, but I did not get how this relates to the cursor position. The `resizeWindow` does not contain any references to window sizes, only to the old viewport decoration sizes. If, then this should be added in `frame_configure`, since `_glfwInputWindowSize` was previously called in `xdgToplevelHandleConfigure`.

> Cursor positions are correct, but minecraft uses window size for calculating offsets what is where relative to the cursor.
> So that's fixed by calling it during resizeWindow.

That sounds a bit weird. Why is it not requesting the cursor position via the API?

Since I cannot reproduce the issue, it would be better if you can submit a fix and make sure that the application behaves as expected.

> 
> Regarding cursor locking problem, I am not sure on the exact repro, spamming cursor lock/unlock should trigger it.
> Cursor just stays on the screen when it's locked. Most of the time.

Can you describe what the problem is? Did this used to work with the old viewport decorations but not the new libdecoration?

ninja-:
> I saw this, but I did not get how this relates to the cursor position. The resizeWindow does not contain any references to window sizes, only to the old viewport decoration sizes. If, then this should be added in frame_configure, since _glfwInputWindowSize was previously called in xdgToplevelHandleConfigure.

Call it wherever you prefer, just make sure it's updated during resize...

> That sounds a bit weird. Why is it not requesting the cursor position via the API?

As I said...Minecraft does math with both the cursor position [which is correct] and with window size [which is incorrect without fix] to know whether you are pointing your cursor at a button.

> Can you describe what the problem is? Did this used to work with the old viewport decorations but not the new libdecoration?

I am not 100% sure as it was unplayable with master glfw...I'll try to check if that helps.
Cursor is locked correctly, relative positions are passed correctly, just the cursor stays in place on the screen.
Gnome Shell, Ubuntu 20.04

ninja-:
btw. I tried to take a screenshot but game is not visible on screenshot, that's how it looks....
interesting whether this is a glfw bug or mutter bug...
Zrzut ekranu z 2020-06-28 20-31-56

christianrauch:
> Call it wherever you prefer, just make sure it's updated during resize...

Well, it's not about the function names... it's about the correct place to call it to have the same sequence of requests sent to the compositor.

Anyhow, I added
```C++
_glfwInputWindowSize(window, width, height);
_glfwPlatformSetWindowSize(window, width, height);
_glfwInputWindowDamage(window);
```
to the `frame_configure`. Those seem to be the important bits from `xdgToplevelHandleConfigure`. I don't know if this solves the problem or if there is more that has to be ported from `xdgToplevelHandleConfigure` to `frame_configure` since I cannot reproduce the issue. Let me know if the current state solves the issue.

> to know whether you are pointing your cursor at a button.

Well, this makes sense. The button position is, of course, irrelevant to the mouse cursor position. But both have to agree to detect if the cursor and button position match.

> Cursor is locked correctly, relative positions are passed correctly, just the cursor stays in place on the screen.

I assume for now that this is unrelated to the decoration code. But since we have no comparison, we won't know before digging into the code.

ninja-:
> Well, it's not about the function names... it's about the correct place to call it to have the same sequence of requests sent to the compositor.

I said `wherever` not `whatever` :D

> to the frame_configure. Those seem to be the important bits from xdgToplevelHandleConfigure. I don't know if this solves the problem or if there is more that has to be ported from xdgToplevelHandleConfigure to frame_configure since I cannot reproduce the issue. Let me know if the current state solves the issue.

I'll try to check your clean branch later.

> I assume for now that this is unrelated to the decoration code. But since we have no comparison, we won't know before digging into the code.

It may be or not, but it would be good to fix it together ^^

ninja-:
tried clean. window size fix is working, cursor locking is not - cursor is visible :/ let's work together on a fix :thinking:

ninja-:
Well according to docs and other examples, the api call to hide the cursor is fine - null surface.
I also checked it's not being "raced" with another call that would override the cursor.
I have no idea why it's not working.
I also tried with weston started in gnome shell, but it's even buggier there and pointer locking doesn't seem to work. I might try with clean weston to see the results, but it probably wouldn't hide the cursor anyway.

christianrauch:
> tried clean. window size fix is working, cursor locking is not - cursor is visible :/ let's work together on a fix thinking

If you can provide a minimal example that reproduces this, I might look into this.

However, I haven't received much feedback on the decoration work yet. I don't want to spend too much time on this if it isn't going to be merged upstream.

ninja-:
> However, I haven't received much feedback on the decoration work yet. I don't want to spend too much time on this if it isn't going to be merged upstream.

Current implementation on master is unusable anyway so I would merge this to move forward :)
From what I see, at least the "right click menu" on window frame is standarized and is handled by mutter. even with libdecoration it works.
It would be good to move forward with this in future and render 1:1 gnome style window frame, but so far I have no idea what would need to be done.

ninja-:
> If you can provide a minimal example that reproduces this, I might look into this.
If you would be interested in running Minecraft with some tooling that overrides libraries, I might help you set it up - it might require a game license though to run it...
Otherwise I could look into writing a _minimal_ reproducer later this week, but it could be hard.

christianrauch:
> If you would be interested in running Minecraft with some tooling that overrides libraries, I might help you set it up - it might require a game license though to run it...
Otherwise I could look into writing a minimal reproducer later this week, but it could be hard.

I don't have a license and I am not really interested in getting into minecraft :-) A minimal example would definitely help to constrain the problem.

> It would be good to move forward with this in future and render 1:1 gnome style window frame, but so far I have no idea what would need to be done.

If all you want is the GTK decoration style, it would probably be better to just use GTK.
Using GTK decorations with libdecoration requires a change in the API https://gitlab.gnome.org/jadahl/libdecoration/-/merge_requests/8 which might put off other projects that just want decorations without handing over the dispatching to libdecoration.
Otherwise, there is no way to get GTK deocrations. I asked for separating the GTK decorations into a dedicated library but this is not going to happen as the code is too heavily depending on GTK and its render engine.

ninja-:
I tried with pure weston and the cursor stays there as well....I am really dedicated to fixing this, hell of a challenge ;)
Also on weston when I maximize the window there are artifacts all over the game screen....
Not sure if this is some glfw bug that results in improper buffer sizes or something...
Sometimes on weston it would crash with xdg_wm_base@24: error 4: xdg_surface buffer does not match the configured state
- sounds like something to do with new decorations, wouldn't you think?

ninja-:
ok I fixed it....needs to destroy old cursor surface as well.
I have no idea why that's needed...some other libraries do this as well...

```
    wl_pointer_set_cursor(_glfw.wl.pointer, _glfw.wl.serial, NULL, 0, 0);
    wl_surface_attach(_glfw.wl.cursorSurface, NULL, 0, 0);
    wl_surface_commit(_glfw.wl.cursorSurface);
```

christianrauch:
> ok I fixed it....needs to destroy old cursor surface as well.

Can you send a PR to my branch or directly to GLFW?

>  Sometimes on weston it would crash with xdg_wm_base@24: error 4: xdg_surface buffer does not match the configured state

I will look into the other issue later. Would be good to have something reproducible for me to test.

ninja-:
repro:
Run pure weston session, run gears example and maximize. Artifacts all over the screen.

ninja-:
I will push a branch later with my fixes. Right now my next priority to fix is the `opaque` mode not fully working...gnome thinks it's transparent and breaks screenshots...

ninja-:
I was also able to finally fix the screenshot problem issue...what a debugging hell it was...
It's choosing wrong EGL config. It should choose one without alpha bits and then it works. I'll try to find a proper place to add the fix...

christianrauch:
> Run pure weston session, run gears example and maximize. Artifacts all over the screen.

I cannot reproduce this, either nested or on a VT. I see decoration artefacts when resizing, but those only appear on Weston.

ninja-:
@christianrauch why did you remove surface commit from setSurfaceOpaque? I can't see any difference in resize performance, but it looked like a bug to me

ninja-:
did you change something specific yesterday? hard to know when you force push every time..
anyway feel free to review #1725

christianrauch:
> why did you remove surface commit from setSurfaceOpaque? I can't see any difference in resize performance, but it looked like a bug to me

It is not required and introduced resize glitches. If you want an opaque surface, request an XRGB buffer or an EGL configuration without alpha channel.

ninja-:
> It is not required and introduced resize glitches. If you want an opaque surface, request an XRGB buffer or an EGL configuration without alpha channel.

What compositor are you testing with related to these glitches?
It's a hint for the compositor so it doesn't need to render other windows underneath.
If you look at my branch, that's what I did as well - changed EGL init to use XRGB when transparency is not requested.
But I really doubt mutter would just look at it being XRGB buffer and treat it the same way as opaque "hint" and properly optimize...

christianrauch:
> What compositor are you testing with related to these glitches?

Mutter

> It's a hint for the compositor so it doesn't need to render other windows underneath.
> If you look at my branch, that's what I did as well - changed EGL init to use XRGB when transparency is not requested.
> But I really doubt mutter would just look at it being XRGB buffer and treat it the same way as opaque "hint" and properly optimize...

Semantically, it has the same meaning. It just more efficient as you have fewer function calls and the buffer "automatically" has the correct opaque size. Overwise you have to make sure that the opaque region does not overlap with other surfaces which will create glitches, e.g. with the decoration shadows.

ninja-:
@christianrauch if only we could prove that mutter treats XRGB buffers and opaque hint as equals and it won't hit performance...because I've never seen any app or framework skip calls to opaque...

christianrauch:
> why did you remove surface commit from setSurfaceOpaque? I can't see any difference in resize performance, but it looked like a bug to me

I just saw that I gave an unrelated answer. Yes, I removed the `wl_surface_commit` because it caused glitches when the opaque region is committed to early. It has to be synchronized with the other calls. You should only commit once, once you have the complete state prepared. Premature commits will create glitches.

Unrelated to this, I think the `setOpaqueRegion` is redundant if you have an XRGB buffer or EGL configuration without an alpha channel. `wl_surface_set_opaque_region` only makes sense if subareas of your transparent surface can be considered opaque.

jadahl:
FWIW, GLFW setting an opaque region without explicitly asking for a XRGB buffer first is incorrect, unless explicitly told by the application what part of the buffer is opaque.

ninja-:
@jadahl right, that's what I adressed in https://github.com/glfw/glfw/pull/1725/commits/2227df83f032ee7868c39fc983749a3fcdc22c7b I think. Maybe we could skip setOpaque then completly

rmader:
FTR, AFAIK that XRGB handling is somewhat broken in EGL, i.e. on Wayland you'll still get an ARGB buffer and on X11 a XRGB one. So you'll likely have to set the opaque region anyway - you can check on Mutter 3.38.1 by running `alt`+`f2` -> `lg` -> `Meta.add_debug_paint_flag(Meta.DebugPaintFlag.OPAQUE_REGION)`  (it has https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1463 which fixes the detection for XRGB buffers)

See:
https://github.com/mpv-player/mpv/issues/8125
https://gitlab.freedesktop.org/mesa/mesa/-/issues/149
https://bugs.freedesktop.org/show_bug.cgi?id=67676#c1

rmader:
Oh and just in case you don't know: when resizing, the subsurfaces should be set to sync (`wl_subsurface_set_sync`), so they can be commited while waiting for the parent surface to commit. If that's feasible here somehow.

jwinarske:
I would prefer decorations to be a build option, and only pull in build dependencies if used.

christianrauch:
@jwinarske Do you mean "decorations" as a general feature to be a build-option or specifically the library that implements decorations? "decorations" as feature have never been a build option, neither on X11 nor on Wayland before. "libdecoration" is a build option (`GLFW_USE_LIBDECORATION`) that will only pull dependencies if activated.

jwinarske:
@christianrauch I'm just concerned about minimizing build deps.  It might also be better to patch the lib with your own cmake file than pull in meson.  Wayland is a bit of a serious army knife, and one size doesn't fit all; hence the concern.  I do like the idea of enabling/disabling things based on presence of protocol, as you have in your other PR.

christianrauch:
Overall, I think GLFW needs working client-side decorations on Wayland. The current implementation of client-side decorations in GLFW is unusable.

> I'm just concerned about minimizing build deps.

That makes sense, of course. Most of `libdecoration`'s library and protocol dependencies overlap with GLFW. `cairo` is the only required dependency for drawing the decorations and rendering fonts. Even if the drawing would be replaced by standard OpenGL functions, you probably need a library for text rendering. There is an optional dependency on dbus to fetch cursor settings, which I highly recommend to have the cursor match the desktop settings.

> Wayland is a bit of a serious army knife, and one size doesn't fit all; hence the concern.

I am not entirely sure what you mean by his. The core protocol does not have much functionality. The protocol extensions, such as `xdg-shell` and `xdg-decoration`, have a very specific scope.

> I do like the idea of enabling/disabling things based on presence of protocol, as you have in your other PR.

`libdecoration` uses the same Wayland protocols than GLFW. Using `xdg-decoration` allows to disable client-side decorations when server-side decorations are supported. However, you still need the functionality of drawing decorations when this is not supported.

DragonSWDev:
@christianrauch Your GLFW fork with libdecoration support causes massive VRAM leak and crashes GNOME Shell for me. My code is working because vanilla GLFW (X11 and Wayland) is working without any problems. I've compiled your version and forced loading it with "LD_PRELOAD". When application window shows up then whole desktop becomes unresposive. I can only move cursor but I can't click anything or use keyboard. After a while whole GNOME Shell crashes and I'm back on the login screen.

When I tried to figure out what happend in log I saw this line:
`gnome-shell[2894]: amdgpu: Not enough memory for command submission.`

I tried to run application again with radeontop running on second window. Right before interface hang I can see VRAM usage up to about 95%. Then it again crashed. I was able to run it once without crashing but desktop started to lag which ended after I closed application.

I don't know what is the cause of this problem (some bug in libdecoration, GLFW, GNOME Shell or GPU driver) and how to find it. If you need more informations please let me know and I'll try to provide them. About hardware and operating system I'm using Gigabyte Radeon RX570 on Fedora 33 with GNOME 3.38.1, kernel 5.8.17 and Mesa 20.2.1.

christianrauch:
@DragonSWDev Is this with any of the standard GLFW examples? Does this also happen with the libdecoration "egl" example?

It would be useful to have an example with reproducible steps to figure out if this is related to libdecoration or other changes I made to GLFW.

DragonSWDev:
@christianrauch No, it happened with my own OpenGL code. For the clarity I'll try with GLFW examples and I'll check libdecoration examples.

DragonSWDev:
@christianrauch Ok, I made tests with examples. Before I'll describe GLFW examples I want to let you know that both libdecoration examples are working without any issues (libdecor-demo and libdecor-egl). As for GLFW examples - some crashes with segmentation fault and log I'll provide below and some are running without GNOME crash but their window is transparent and I can't move it or click buttons on decoration. Here is the list of examples behaviour:
boing - crash
gears - crash
heightmap - unresponsive
offscreen - opens and closes instantly even on vanilla GLFW
particles - unresponsive
sharing - crash
simple - unresponsive
splitview - unresponsive
wave - unresponsive

Crashing examples aside from segmentation fault error are giving this output:
```
amdgpu: Failed to allocate a buffer:
amdgpu:    size      : 512586494132224 bytes
amdgpu:    alignment : 131072 bytes
amdgpu:    domains   : 4
amdgpu:    flags   : 6
amdgpu: Failed to allocate a buffer:
amdgpu:    size      : 512586494132224 bytes
amdgpu:    alignment : 131072 bytes
amdgpu:    domains   : 4
amdgpu:    flags   : 6
```

Bytes values are the same during every execute. They only differ between examples (this particular log is from boing example).

christianrauch:
None of these GLFW examples crash or become unresponsive on my setup (Ubuntu 20.04, Intel `iris` driver). I see glitches (on master and with this PR) with transparent surfaces in examples `heightmap`, `splitview` and `wave`.

Do you only see the VRAM leaking with this PR and with `GLFW_USE_LIBDECORATION` set `ON` or `OFF`? Since I do not see these leaks with the Intel driver, I assume that this is related to the AMD driver. I would suggest opening an issue for this in the mesa repo.

DragonSWDev:
@christianrauch I made another build and now all examples are working (with libdecoration support) without any problems (excluding glitches you also noticed). It's no more crashing or becoming unresponsive. I agree it looks like driver bug (application shouldn't be able to crash driver) but before I'll report it to Mesa repo I would like to have more informations about possible cause of this bug. Can you tell me what changes you did with your fork? It looks like those changes fixed problem at least for me. My own applications also started to work with this build.

christianrauch:
@DragonSWDev I fixed an issue that was reported by the AddressSanitizer. I added the AddressSanitizer in this PR to check for any memory issues. In debug builds, this should be reported issues on the stdout and quit the process, rather than crashing the application without any message.
I forgot to check if the payload provided in the callbacks is coming from the application/GLFW surface and not the libdecoration subsurfaces. This is fixed now, but I also don't see how this should leak VRAM.

If you want to undo this change to test if it causes your leaks, remove
```C++
if (wl_proxy_get_tag((struct wl_proxy *) output) != &proxy_tag)
    return;
```
from the `surfaceHandleEnter` and `surfaceHandleLeave` in `wl_window.c`.

rmader:
FTR., I spoke with other Gnome devs about the fact that clients apparently can starve Mutter concerning vram, making it crash. It looks like there are plans to allow vram resource management via cgroups, making sure Mutter always gets what it needs to keep functioning.

Clients should never be able to crash a session.

DragonSWDev:
@christianrauch @rmader pointed to the Mutter as cause. I blamed driver but actually if it would be driver issue then probably it would end in whole system crash, not only desktop. Looks like it needs to be fixed on GNOME side. I can only confirm that your changes make GLFW and libdecoration working. I hope it will land soon, it makes GLFW Wayland experience much better on GNOME. Yeah, it's not closely integrated with desktop theme like on compositors with xdg-decorations support but it's much better to have custom decorations than no full decorations at all. I had issues with moving and resizing with vanilla GLFW decorations and with libdecorations they are gone and everything works perfectly. Before this will land officially I'm going to use my custom build in my projects. Good job!

@rmader Thank you for investigating this. Good to know it's known issue and there are plans to fix it. I agree, clients should never make compositor crash. It should be fixed on GNOME side.

rmader:
@DragonSWDev: the leak would still make the application crash so we should make sure it doesn't happen. The fact that it can also crash the system compositor just makes it much worth than it should be :)

FTR: the fix would need to be in the kernel and would benefit other DEs as well, most notably KDE as it also works on making use of cgroups.

christianrauch:
> is this still an active work in progress?

Do you mean the decoration library or this very PR?

The library is mostly feature-complete. I haven't received any feedback on this PR from @elmindreda, so I am just updating the API from time to time to make it compile with the upstream.

christianrauch:
@elmindreda @linkmauve Can someone approve the CI workflow?

DragonSWDev:
It is worth mentioning that SDL merged libdecor support and libdecor landed in Fedora 34 repository. More distributions will probably follow this as well.

vchernin:
@christianrauch me and @yavko both run into blurriness when running this patch at HiDPI on GNOME. It doesn't seem to exist without this patch.

[Here's a diff](https://github.com/vchernin/minecraft-wayland-misc/commit/bad6665a4c0204ec66d25158c2800ce883a86357?short_path=53f499b#diff-53f499be91bd991915f5bf5e2e55f2e9b8858bd706ccf8755ed9743d10f35cf1) (click Swipe at the bottom and zoom in with your browser). I took it a while ago but the issue seems to still exist, with this patch Minecraft is slightly, but noticeably blurrier than without this patch.

Haven't yet tried running an SDL app or something to confirm whether it's a general libdecor bug, I suppose that's a good next step if there's no ideas. FYI the patches stored in this [helpful repo](https://github.com/Admicos/minecraft-wayland/tree/libdecoration) were used. Note an older version of glfw was used, since this patch hasn't been rebased in a while.

Regardless thanks for working on this, hopefully at some point this does get merged.

christianrauch:
I stopped working on this since it's not clear if this gets merged eventually. If there is interest by the upstream developers (@elmindreda), I would probably look into this again in more detail.

libdecor supports HiDPI. That means that the UI elements/surfaces are rendered at the scale that is set for the display. Hence the decoration UI elements should not appear blurry. However, for the "content" surface, the client is responsible for managing the surface/buffer scale depending on which display the surface is shown.
Check out the [example](https://gitlab.gnome.org/jadahl/libdecor/-/blob/master/demo/demo.c). It should render the checker pattern at the correct size without upscaling.

vchernin:
A couple perhaps notable observations:

Initially when launching the game the decoration elements are blurry. But the mouse cursor is the correct size for both the decoration and "content" surface.

After the loading bar appears, when hovering over the "content" surface (i.e. the game itself), the cursor appears larger than normal. When hovering over the libdecor UI elements, the cursor appears smaller (the correct size compared to GTK apps and hovering over nothing).

https://user-images.githubusercontent.com/38842733/145725869-b20df185-6756-42e2-9d99-7f9380aa95fd.mp4

Interestingly at about 11 seconds the cursor is the correct size everywhere, but I think that might be since I hovered over the window buttons, and temporarily the cursor thought it was on a different surface than it was actually on (might be an incorrect explanation).

I can't reproduce any of this with the libdecor quick start from the readme on this same setup.

Gnome shell/mutter 41.1 on Fedora 35.

yavko:
> A couple perhaps notable observations:
> 
> Initially when launching the game the decoration elements are blurry. But the mouse cursor is the correct size for both the decoration and "content" surface.
> 
> After the loading bar appears, when hovering over the "content" surface (i.e. the game itself), the cursor appears larger than normal. When hovering over the libdecor UI elements, the cursor appears smaller (the correct size compared to GTK apps and hovering over nothing).
> 
>  cursor-inconsistent-size-demo-libdecor.mp4 
> Interestingly at about 11 seconds the cursor is the correct size everywhere, but I think that might be since I hovered over the window buttons, and temporarily the cursor thought it was on a different surface than it was actually on (might be an incorrect explanation).
> 
> I can't reproduce any of this with the libdecor quick start from the readme on this same setup.
> 
> Gnome shell/mutter 41.1 on Fedora 35.

for me the mouse cursor gets way smaller and on HIDPI the ui is blurry

jgcodes2020:
is this PR going to be placed in the [Wayland backend](https://github.com/glfw/glfw/projects/2)?

elmindreda:
Thank you so much for this PR! I definitely want to merge this for 3.4. The current client-side decorations are unusable and libdecor seems like the best available solution, especially now that it has proper packages.

The only major change I'd suggest is that the choice between libdecor and XDG decoration be made at run-time, with libdecor dynamically loaded when needed. That said, these bits of GLFW have been changing a lot lately and might again soon, so this might be a bad time to rebase. Also note that rebase status won't prevent this from being accepted.

Thank you for your patience! I'm still in the process of taking on more maintenance of the Wayland code and currently can't give a proper code review, but will as soon as I'm able.

jwinarske:
@elmindreda My only request with this PR is that is can be disabled at build, opposed to runtime detect.  The use cases involve embedded Linux where one most likely does not want Wayland window decorations.

elmindreda:
@jwinarske The libdecor API is small and the bits used by GLFW could be declared in an internal header, as is already done with some other APIs. That way there's no external compile-time dependency even when support is enabled.

orowith2os:
Coming from https://github.com/glfw/glfw/issues/1639, how goes this? Is there anything that needs to be done, tested, etc?

tauin:
how would you get this to use different libdecor plugins? e.g GNOME titlebars

orowith2os:
> how would you get this to use different libdecor plugins? e.g GNOME titlebars

Libdecor will automatically use the GTK plugin on GNOME, from what I can tell. Can't speak for any other plugins, though.

christianrauch:
> how would you get this to use different libdecor plugins? e.g GNOME titlebars

Every plugin defines a priority and then the plugin with the highest priority is loaded. On the current master (https://gitlab.freedesktop.org/libdecor/libdecor) the GTK plugin takes precedence over the cairo plugin. Hence, the GTK plugin will be loaded by default when available. This does not depend on the desktop shell yet.

TheBrokenRail:
> > how would you get this to use different libdecor plugins? e.g GNOME titlebars
> 
> Libdecor will automatically use the GTK plugin on GNOME, from what I can tell. Can't speak for any other plugins, though.

One caveat: if you're running the `libdecor` demos from the source-tree ([using `meson devenv`](https://gitlab.freedesktop.org/libdecor/libdecor#quick-start)), it will [force the `cairo` plugin](https://gitlab.freedesktop.org/libdecor/libdecor/-/blob/25f29fefd58587aa12389c05c082087390f0b170/src/plugins/cairo/meson.build#L26).

christianrauch:
> One caveat: if you're running the `libdecor` demos from the source-tree ([using `meson devenv`](https://gitlab.freedesktop.org/libdecor/libdecor#quick-start)), it will [force the `cairo` plugin](https://gitlab.freedesktop.org/libdecor/libdecor/-/blob/25f29fefd58587aa12389c05c082087390f0b170/src/plugins/cairo/meson.build#L26).

This is just because this part is missing for the GTK plugin. If you add it there, the GTK will take precedence again. Anyway, this is only for the devenv and is irrelevant for the final distributed releases. If you think this is a limiting factor, then open an issue. I can add it if required.

elmindreda:
Thank you everyone for your patience. Please have a look at the [libdecor-r2](https://github.com/glfw/glfw/tree/libdecor-r2) branch for my proposal for initial libdecor support. It should do everything this PR does, while adding (optional) dynamic loading and implementing libdecor paths for a few more GLFW features. It's a re-implementation that came out of learning the libdecor API, so any bugs in it are my fault and not @christianrauch's.

libdecor is a really neat library and fit in easily once a few things in GLFW were cleaned up. I'm likely using it incorrectly in places, but it mostly Works on My Machine now. Tested with libdecor 0.1.0-3build1 on Ubuntu 22.04.

I kept the existing client-side decorations as a fallback of last resort, as in their current state I think they're better than no decorations at all, but with libdecor installed you should never see them.

Currently toggling `GLFW_DECORATED` isn't working well with Cairo decorations and isn't working at all with XDG decorations, and I haven't been able to figure out why.

I don't feel entirely confident about the workaround I added to make sure libdecor's globals are in place if some other part of `glfwInit` fails and the libdecor context is destroyed before the regular event pump starts. This is needed because the 0.1.0 Cairo plugin destroys its shm global unconditionally.

christianrauch:
@elmindreda Can you open a PR with your branch where this can be reviewed?

PR #1696, opened 2020-05-18
Add keyboard layout callback and name query

### LINKED ISSUE #1201: Keyboard layout changed callback

elmindreda:
This adds a global callback for when the keyboard layout changes and a query for the localized, human-readable name of the current layout.

On macOS it also triggers when the mode changes for an IME, for example switching between hiragana and katakana, while on other platforms they are both referred to as `Japanese` or similar. I don't know whether or how to fix that inconsistency.

On Wayland the name query returns data cached from received events but on other platforms it queries the system directly.

This PR is intended to fix #1201.

This functionality may be affected by IME support work before 3.4 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,10 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+ - Added `glfwGetKeyboardLayoutName` for querying the name of the current
+   keyboard layout (#1201)
+ - Added `glfwSetKeyboardLayoutCallback` and `GLFWkeyboardlayoutfun` for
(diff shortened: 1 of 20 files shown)


norcalli:
Does the IME inconsistency on other platforms mean that it won't get notified at all on a change like hiragana/katakana or that the names will just appear to be the same?

PR #1698, opened 2020-05-22
Add statement to docs for glfwTerminate

tombsar:
At the moment it is safe to call `glfwTerminate` even if the library is not currently initialized (it returns instantly with no side effects), but this is not mentioned in the documentation. Unless you plan to change this behaviour in future (and I don't think you should), it would be great to make this a promise so that users don't need to add their own redundant conditional.


elmindreda:
Thank you!

PR #1700, opened 2020-05-25
Fix /w4 compile warnings

ben1:
These changes simply avoid re-declaring variables that were already declared. In one case the type of the variable was different too, so the name of the variable was also changed to reflect its usage.


elmindreda:
Thank you!

PR #1701, opened 2020-05-25

Luflosi:
The `atomCount` variable has the type `unsigned long`, so the `for` loop iterating over it should use the same type.

Closes #1701.


elmindreda:
Thank you!

PR #1702, opened 2020-05-25
Replace %m conversion specifier with %s and strerror()

Luflosi:
When compiling with `-Wall` and `-pedantic-errors`, gcc complains with
```
warning: ISO C does not support the '%m' gnu_printf format [-Wformat=]
```
because the `%m` conversion specifier is a GNU extension.

Closes #1702.

I'm not sure if this is actually a problem for most people. This was a problem for kitty but since it has its own fork of GLFW, it isn't a big deal.


Luflosi:
Wow, that was fast!

linkmauve:
Well sure, it was an easy and uncontroversial change to review. :)

PR #1703, opened 2020-05-25
Use __typeof__() instead of void*

Luflosi:
Preserve the real type of a pointer instead of casting to `void*`.

Closes #1703.

Can `__typeof__()` be used on all supported platforms?

### DIFF
--- a/src/internal.h
+++ b/src/internal.h
@@ -218,7 +218,7 @@ typedef void (APIENTRY * PFN_vkVoidFunction)(void);
 // Swaps the provided pointers
 #define _GLFW_SWAP_POINTERS(x, y) \
     {                             \
-        void* t;                  \
+        __typeof__(x) t;          \
         t = x;                    \
         x = y;                    \
         y = t;                    \


Luflosi:
Apparently it's not supported everywhere. Is there a way to achieve the same thing that works on all platforms?

elmindreda:
Not that I know of. Is this intended to fix the issue of casts between function and non-function pointers?

Luflosi:
This caused warnings in kitty:
```
glfw/input.c:1019:5: error: assigning to 'void *' from 'GLFWkeyboardfun' (aka
      'void (*)(struct GLFWwindow *, struct GLFWkeyevent *)') converts between
      void pointer and function pointer [-Werror,-Wpedantic]
    _GLFW_SWAP_POINTERS(window->callbacks.keyboard, cbfun);
    ^                   ~~~~~~~~~~~~~~~~~~~~~~~~~~
glfw/internal.h:233:11: note: expanded from macro '_GLFW_SWAP_POINTERS'
        t = x;                    \
          ^ ~
```
But since there seems to be no good way to fix this on all platforms, I'll close this PR.

elmindreda:
We could remove the macro altogether for one extra line per setter function. I would accept a PR that did that.

PR #1708, opened 2020-06-02
Reorder platform fields in mappings.h.in

tombsar:
The upstream SDL_GameControllerDB README (https://github.com/gabomdq/SDL_GameControllerDB) has a convention that the `platform` entry should always be at the end of the mapping string. Would it make sense to follow that with glfw's own mappings?

I haven't regenerated `mappings.h`, but I could do that and add to this PR if requested.

### DIFF
--- a/src/mappings.h.in
+++ b/src/mappings.h.in
@@ -61,13 +61,13 @@
 const char* _glfwDefaultMappings[] =
 {
 @GLFW_GAMEPAD_MAPPINGS@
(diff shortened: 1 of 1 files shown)


tombsar:
Sorry, it's maybe not so obvious since SDL2's own controllermap tool puts `platform` at the start of the string...

elmindreda:
I think it's good to not necessarily use the same order as our upstream source so our parser doesn't get too complacent.

Issue #1695, opened 2020-05-18
Robustness of GL header include guard


tombsar:
Yesterday I tried to integrate glad into a project using OpenGL ES, and had to spend an hour diagnosing the most frustrating compile errors of my life (thanks gcc!) caused by, it turns out, inclusion of both glad.h (explicitly by me) and gl.h (implicitly by glfw).

I understood from the documentation (https://www.glfw.org/docs/latest/context_guide.html#context_glext_auto) that:

> By including the glad header before the GLFW header, it suppresses the development environment's OpenGL or OpenGL ES header.

However I don't see any mechanism in glfw3.h that enforces this; it looks as though in the absence of any `GLFW_INCLUDE_*` defines, GL/gl.h is included unconditionally (https://github.com/glfw/glfw/blob/master/include/GLFW/glfw3.h#L210).

gl.h has an include guard to prevent duplicate inclusion, but the guard has a different name for each version of OpenGL. Hence, the inclusion of GLES/gl.h (or equivalent glad.h) does not prevent the inclusion of GL/gl.h.

Easy to work around once I found the issue, (`#define GL_INCLUDE_NONE`), but it would be great if I didn't need to do that.

At the moment I'm wondering why `GL_INCLUDE_NONE` isn't the default behaviour, making the user specifically ask if they want glfw to include an OpenGL header. ~~Obviously that's a breaking change, so could only be considered for glfw 4.~~ Significant change, but not necessarily **breaking**.

In the meantime, I think we need at least an update to the documentation to make this behaviour more obvious, if not an actual fix.

Thoughts?

elmindreda:
Agreed, `GLFW_INCLUDE_NONE` is a much better default. That decision carried over from 2.x and I didn't reexamine it until 3.0 was already out.

Let's see what we can do about it in 3.x.

elmindreda:
Since mixing any of these headers is a bad idea, we could perhaps add the falsely promised mechanism. Detect the canonical inclusion guards early on and define `GLFW_INCLUDE_NONE` if any are present.

Switching the default would be the cleaner solution but it would also break the compilation of a lot of tutorial code, including our widely copied [minimal example](https://www.glfw.org/documentation.html).

elmindreda:
Sorry, that was mostly just restating what you had already said.

Does this seem reasonable?
```diff

-#elif !defined(GLFW_INCLUDE_NONE)
+#elif !defined(GLFW_INCLUDE_NONE) && \
[...]

PR #1712, opened 2020-06-05
Fixed: GLFW_INCLUDE_GLU is not working related to issue #1695.

tatsy:
I found that the fix proposed in issue #1695 prevents `glu.h` from being included even when `GLFW_INCLUDE_GLU` is specified.

In this PR, I separated the section for `glu.h` from the section where the standard GL.h is included.


elmindreda:
Thank you!

PR #1717, opened 2020-06-12
Should pass hdc when invoke eglGetDisplay at win32

halx99:
Such as EGL provide by google AngleProject, see https://github.com/google/angle/blob/master/src/libANGLE/renderer/d3d/d3d11/Renderer11.cpp function ```Renderer11::initialize```

### DIFF
--- a/src/egl_context.c
+++ b/src/egl_context.c
@@ -396,9 +396,7 @@ GLFWbool _glfwInitEGL(void)
     }
 
 #if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


halx99:
@elmindreda We need ```ReleaseDC``` after ```eglGetDisplay```

elmindreda:
The GLFW window class is `CS_OWNDC`, meaning [`ReleaseDC` is a no-op](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-releasedc).

PR #1720, opened 2020-06-18
Update documentation for OPENGL_DEBUG_CONTEXT hint

tombsar:
The documentation currently says that the GLFW_OPENGL_DEBUG_CONTEXT hint is ignored if an OpenGL ES context is requested, but I don't think that matches the implemented behaviour.

### DIFF
--- a/docs/window.dox
+++ b/docs/window.dox
@@ -408,8 +408,7 @@ Forward-compatibility is described in detail in the
 @anchor GLFW_OPENGL_DEBUG_CONTEXT_hint
 __GLFW_OPENGL_DEBUG_CONTEXT__ specifies whether to create a debug OpenGL
 context, which may have additional error and performance issue reporting
-functionality.  Possible values are `GLFW_TRUE` and `GLFW_FALSE`.  If OpenGL ES
-is requested, this hint is ignored.
+functionality.  Possible values are `GLFW_TRUE` and `GLFW_FALSE`.
 
 @anchor GLFW_OPENGL_PROFILE_hint
 __GLFW_OPENGL_PROFILE__ specifies which OpenGL profile to create the context
--- a/docs/window.dox
+++ b/docs/window.dox
@@ -406,9 +406,14 @@ Forward-compatibility is described in detail in the
 [OpenGL Reference Manual](https://www.opengl.org/registry/).
 
 @anchor GLFW_OPENGL_DEBUG_CONTEXT_hint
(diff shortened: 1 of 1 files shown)


docs/window.dox
```diff
@@ -408,8 +408,7 @@ Forward-compatibility is described in detail in the
 @anchor GLFW_OPENGL_DEBUG_CONTEXT_hint
 __GLFW_OPENGL_DEBUG_CONTEXT__ specifies whether to create a debug OpenGL
```
elmindreda, line 409: This should list both APIs or neither.

docs/window.dox
```diff
@@ -408,8 +408,7 @@ Forward-compatibility is described in detail in the
...
 context, which may have additional error and performance issue reporting
-functionality.  Possible values are `GLFW_TRUE` and `GLFW_FALSE`.  If OpenGL ES
-is requested, this hint is ignored.
```
elmindreda, line 412: 👍

docs/window.dox
tombsar: True. I'll reword it.

tombsar:
> Possibly we should add a preferred GLFW_DEBUG_CONTEXT alias as well?

I'd be in favour of that, although I think it should be `GLFW_CONTEXT_DEBUG` for consistency with the other hints?

Up to you whether you want to introduce the new hint at the same time as this documentation hotfix, or leave it for a future release.

elmindreda:
> I'd be in favour of that, although I think it should be `GLFW_CONTEXT_DEBUG` for consistency with the other hints?

Oh, yes, definitely, thank you!

> Up to you whether you want to introduce the new hint at the same time as this documentation hotfix, or leave it for a future release.

Separate would be best, I think.

tombsar:
I struggled to find a wording I was happy with, so I ended up borrowing the phrase "debug mode" from the OpenGL wiki. I also added a reference to the Khronos extension definition

What do you think? Did I change it too much?

I'm a bit lost with Github's review system, so hopefully this all works smoothly.

elmindreda:
> What do you think? Did I change it too much?

It's great! Will merge asap.

elmindreda:
Thank you! This was merged as 8a69a0d7e545a88de3dd88710e4a881932ca07ed with corresponding fixes in 4e3b43383a1e738dfcf47205671246f2008a8d94 (that I didn't think to check for before writing b420ca74047e26d03bb107b66c50108b153b4ce1).

PR #1738, opened 2020-07-23
Return NULL instead of GLFW_FALSE where appropriate

Luflosi:
Related to 782e6b6cefc9ac2a2b6779f9e658ab948809253b.

I think this commit is correct because these functions return `NULL` in other cases as well and never `GLFW_FALSE`.


elmindreda:
Thank you!

PR #1739, opened 2020-07-24
Use switch statement instead of if-else-chain for cursor shapes

Luflosi:
I propose replacing the if-else-chains for the cursor shapes with switch statements. Since the C language provides this construct for this exact purpose, I think we should use it. Is there any reason not to do that?


Luflosi:
It's been almost three months. Any chance to get this reviewed this year or at least tell me a rough time estimate?

elmindreda:
Thanks for the bump. Adding this to my queue. Should be merged soon.

elmindreda (approved):

PR #1745, opened 2020-08-08
Make _glfwPlatformInitJoysticks() return the correct type in null_joystick

Luflosi:
This code was introduced in 782e6b6cefc9ac2a2b6779f9e658ab948809253b.
All the other instances of `_glfwPlatformInitJoysticks()` also return `GLFWbool`.

Closes #1745.


elmindreda:
Thank you!

PR #1756, opened 2020-08-31
Add glfw3.hpp wrapper a la vulkan.hpp

MTCoster:
**This is currently in a proof-of-concept state, and is by no means ready to merge.**

I was inspired by [vulkan.hpp](https://github.com/KhronosGroup/Vulkan-Hpp) to create a C++-style wrapper around `glfw3.h`: `glfw3.hpp`.

This allows for the use of paradigms central to C++ (such as RAII) when using GLFW.

I’m open to suggestions for how this should look (I do have some ideas of my own already), and more than willing to be responsible for its upkeep indefinitely.

### DIFF
--- a/include/GLFW/glfw3.hpp
+++ b/include/GLFW/glfw3.hpp
@@ -1,0 +1,177 @@
+#ifndef _glfw3_hpp_
+#define _glfw3_hpp_
+
+#ifndef GLFW_HPP_NAMESPACE
+#define GLFW_HPP_NAMESPACE glfw
+#endif
+
+#ifdef GLFW_HPP_NO_STD_STRING
+#define GLFW_HPP_STRING char const *
+#else
+#define GLFW_HPP_STRING std::string
+#endif
+
+#include <GLFW/glfw3.h>
+
+#include <functional>
+#include <optional>
+#include <string>
+#include <system_error>
+#include <vector>
+
+namespace GLFW_HPP_NAMESPACE {
+
+    namespace _private {
+#ifdef GLFW_HPP_NO_STD_STRING
+        inline GLFW_HPP_STRING makeString_(char const *str) { return str; }
+        inline char const *cString_(GLFW_HPP_STRING const &str) { return str; }
+#else
+        inline GLFW_HPP_STRING makeString_(char const *str) { return std::string(str); }
+        inline char const *cString_(GLFW_HPP_STRING const &str) { return str.c_str(); }
+#endif
+    }
+
+    using ErrorFun = GLFWerrorfun;
+
+    struct Size {
+        int width;
+        int height;
+    };
+
+    enum class ErrorCode {
+        noError = GLFW_NO_ERROR,
+        notInitialized = GLFW_NOT_INITIALIZED,
+        noCurrentContext = GLFW_NO_CURRENT_CONTEXT,
+        invalidEnum = GLFW_INVALID_ENUM,
+        invalidValue = GLFW_INVALID_VALUE,
+        outOfMemory = GLFW_OUT_OF_MEMORY,
+        apiUnavailable = GLFW_API_UNAVAILABLE,
+        versionUnavailable = GLFW_VERSION_UNAVAILABLE,
+        platformError = GLFW_PLATFORM_ERROR,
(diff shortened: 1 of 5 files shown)


janekb04:
@MTCoster Hi, I also like the idea to have an official interface for C++ in GLFW. I would like to just quickly note that there are now [two C++ wrappers](https://www.glfw.org/community.html#wrappers) for GLFW. I made one of them ([GLFWPP](https://github.com/janekb04/glfwpp)) and it seems a bit similar to your proposal.

include/GLFW/glfw3.hpp
```diff
@@ -0,0 +1,177 @@
...
+                Error::getAndThrow_();
+            }
+        }
+
+        explicit GLFW(ErrorFun errorCallback) : GLFW() {
+            this->errorCallback = errorCallback;
+            glfwSetErrorCallback(this->errorCallback);
+        }
+
+        virtual ~GLFW() {
```
gracicot, line 154: Very small nitpick: the `GLFW` type does not seem like a polymorphic type, but has a virtual destructor. Unless you explicitly intend users to extend and use the `GLFW` type in a polymorphic context, I would remove the `virtual` there.

include/GLFW/glfw3.hpp
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+
+        friend class GLFW;
+        friend class Window;
+    };
+
+    class Window {
+    public:
+        Window(Size const &size, GLFW_HPP_STRING const &title) :
```
gracicot, line 115: Here the lifetime of a `Window` object is tied the whether the window is closed or not, but the window is leaked. This does not follow the rule of 5, as you initialize a resource in the constructor.

include/GLFW/glfw3.hpp
```diff
@@ -0,0 +1,177 @@
...
+        }
+    };
+
+    inline std::error_code make_error_code(ErrorCode code) {
+        return {static_cast<int>(code), ErrorCategory()};
+    }
+
+    class Error : public std::system_error {
+    public:
+        ErrorCode const code;
```
gracicot, line 84: If you can change the member to not be const and add a `code()` function to get the error code, a bit like boost is doing, that would be super good!

gracicot:
I added a few comments that would improve the interface a bit.

include/GLFW/glfw3.hpp
janekb04: Perhaps the `final` specifier could be also added to prevent the class from being inherited from.

MTCoster:
> @MTCoster Hi, I also like the idea to have an official interface for C++ in GLFW. [...] I made one of them ([GLFWPP](https://github.com/janekb04/glfwpp)) and it seems a bit similar to your proposal.

Hi @janekb04, I don't think your wrapper existed when I first proposed this, but it looks almost exactly how I imagined mine would eventually! I think instead of reinventing the wheel, I would prefer to support getting your solution upstreamed as `glfw3.hpp` (if you are willing to contribute it of course).

Alternatively, I'd be happy to resume work on a version designed specifically to be part of the core GLFW library if you would prefer to keep your work separate. Either way; I'm glad there's support in the community for such an undertaking!

janekb04:
@MTCoster After giving it a thought, I would be very glad to have my library included as part of GLFW. It would certainly require some changes, but I don't think they would be substantial. I would mostly be just making it more in-line with modern C++ like by adding move semantics and possibly restructuring it to be in a single file or creating separate `cpp` file(s).

Arr4y:
One of my favorite bindings for glfw is vkfw (https://github.com/Cvelth/vkfw). One of the main reasons is that is has great support for Vulkan. I didn't realize it at first but it gives my access to the hpp Vulkan headers. One of the main advantages is that it took significantly less lines of code to do things.

I am wondering if there's any plans to support Vulkan hpp headers with this?

janekb04:
@Arr4y Hi, my wrapper already supports Vulcan-Hpp.

Arr4y:
> 
> 
> @Arr4y Hi, my wrapper already supports Vulkan-Hpp.

Awesome. I'm looking forward to this. Thank you for the heads up.

PR #1757, opened 2020-09-01
[docs] Better scaling on mobile and small screens.

siavashserver:
Hopefully the docs will become easier to read on mobile devices.

Related: glfw/website#70
Follow up to : glfw/website#71

PR #1759, opened 2020-09-03
[docs] Migrating to Sass for stylesheets.

siavashserver:
Migrating to Sass for docs as well to unify the workflow.

Related: glfw/website#70
Follow up to : glfw/website#71 and glfw/glfw#1757


linkmauve:
Since this seems to be done mostly for variables, may I suggest using [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) instead? They don’t require any compilation step, and make it easier to change them at runtime (for instance for a dark theme, if we ever get one in the future).

siavashserver:
I still rather to keep sass/less in my toolchain for stuff like mixins,
etc; they come in handy 😋

In other words sass/less and css vars can coexist and complement each other.

linkmauve:
I see no problem with that! :D

elmindreda:
Merged, thank you!

Issue #1676, opened 2020-04-18
Connecting/Disconnecting Monitor on Windows 10 rewrites properties in Primary monitor


nvmsocool:
I am using the latest release in 32 bit mode on a 64 bit Windows 10 machine (with no other libraries) using Visual Studio 2019, and found that disconnecting or connecting a monitor causes issues with the properties of the primary monitor. The properties I noticed were the working width and height were set to 0, and the content scale was set to a very large number.

I have attached the sample code that recreates the issue.  
[GLFWPOC.txt](https://github.com/glfw/glfw/files/4495998/GLFWPOC.txt)

Steps to recreate:
1) Run the application
2) Right click to display the working width and height of the primary monitor
3) Connect a monitor
4) Right click again to see that the width and height are now 0.

Left clicking at any time will try to go into fullscreen, which behaves unpredictably after the connect/disconnect issue happens.

elmindreda:
Thank you for reporting this! It should be fixed now with 166611cef5967313e6a3d50849e57ddbcdc33f26 by @dougbinks.

PR #1761, opened 2020-09-04
Win32 update monitor handles when monitors are added or removed

dougbinks:
On Windows, when a monitor is added or removed (or in some cases just a display setting change is made), the monitor display handle can change.

Currently GLFW does not update handles for monitors whose names remain the same after a monitor change, which results in calls to some functions which need the monitor handle, such as ```glfwGetMonitorWorkArea``` to fail.

This PR has the minimal change needed to fix this by updating the monitor handle. A larger code change could be made which would minimize the computation, but as monitor changes are infrequent and also take some time this is a decent enough and minimal solution.

This should fix issue #1676

### DIFF
--- a/src/win32_monitor.c
+++ b/src/win32_monitor.c
@@ -185,6 +185,8 @@ void _glfwPollMonitorsWin32(void)
                            display.DeviceName) == 0)
                 {
                     disconnected[i] = NULL;
+                    // handle may have changed, update
+                    EnumDisplayMonitors(NULL, NULL, monitorCallback, (LPARAM) _glfw.monitors[i]);
                     break;
                 }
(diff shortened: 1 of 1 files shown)


dougbinks:
Just a quick note to say that should a more optimal performance version be required which only calls EnumDisplayMonitors once, I would be happy to make the change - it would only add a few more lines of code.

elmindreda (approved):

elmindreda:
Thank you! This has been merged as 166611cef5967313e6a3d50849e57ddbcdc33f26.

This should be sufficiently optimal until the changes land that make disconnected monitor objects more robust.

Issue #51, opened 2013-06-17
Default window created with 16-bit color with OS X / AMD Radeon HD 6750M


jstoecker:
Using GLFW3, the default pixel format for the window is 16-bit color on OS X 10.8.4 with an AMD Radeon HD 6750M. If I force use of integrated graphics (Intel HD Graphics 3000) I get 32-bit color. I used the code found here for testing:

http://www.glfw.org/docs/3.0/quick.html

This issue does not occur with GLFW 2.7. I was able to fix the problem by adding
glfwWindowHint(GLFW_ALPHA_BITS, 8) before window creation. Without this, kCGLPFAAlphaSize shows as 1 instead of 8 which is stated as the default value.

![glfw-bug-osx](https://f.cloud.github.com/assets/3835701/660503/9e289ad6-d702-11e2-9773-096c066b785d.png)

PR #1769, opened 2020-09-18
Desktop Environment freezes when GLFW Window width or height is >32767

satej1210:
I found this bug while working with the GLFW library with Open3D. 

Basically, if you set the width or height to 32768 in the `boing.c` example, the entire desktop would freeze. While stepping through the code, the program would freeze at `XCheckTypedWindowEvent` in [x11_window.c:114](https://github.com/glfw/glfw/blob/96f9f5c4b970140ab5380fb9886ba787c8698937/src/x11_window.c#L114).

I thought it made sense to add the `assert` statements to check for `width > 32767` and `height > 32767` because if you remove the asserts that are already present in the function, and the if statement to check if `width or height <= 0`, then the desktop environment will freeze in a similar way when you set the `width or height > 32767`.

OS: `Linux 5.4.0-47-generic #51~18.04.1-Ubuntu SMP Sat Sep 5 14:35:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
xserver-xorg-core version: `1.20.8-2ubuntu2.2~18.04.3`
Release or commit: 96f9f5c4b970140ab5380fb9886ba787c8698937 (latest commit)
Error messages: None because the system needs to be restarted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,8 @@ information on what to include when reporting a bug.
    combinaitons (#1598)
  - [X11] Bugfix: Keys pressed simultaneously with others were not always
    reported (#1112,#1415,#1472,#1616)
+ - [X11] Bugfix: Desktop Environment freezes when window width or height is
+   over 32767
  - [Wayland] Removed support for `wl_shell` (#1443)
  - [Wayland] Bugfix: The `GLFW_HAND_CURSOR` shape used the wrong image (#1432)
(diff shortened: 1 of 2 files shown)


satej1210:
Probably worth mentioning that this bug only occurs on xserver-xorg-core version `1.20.8-2ubuntu2.2~18.04.3` (which is on Ubuntu 18.04), but does not show up on `1.20.8-2ubuntu2.4` (on Ubuntu 20.04)

PR #1774, opened 2020-09-29
remove ECM dependency

jwinarske:
This PR removes the ECM build dependency for Wayland.

This PR was built and tested on Ubuntu 18.04.5 LTS and Yocto Dunfell.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,10 +52,6 @@ src/glfw_config.h
 src/glfw3.pc
 src/glfw3Config.cmake
 src/glfw3ConfigVersion.cmake
-src/wayland-pointer-constraints-unstable-v1-client-protocol.h
-src/wayland-pointer-constraints-unstable-v1-protocol.c
-src/wayland-relative-pointer-unstable-v1-client-protocol.h
-src/wayland-relative-pointer-unstable-v1-protocol.c
 
 # Compiled binaries
 src/libglfw.so
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,20 +191,18 @@ endif()
 # Use Wayland for window creation
 #--------------------------------------------------------------------
 if (_GLFW_WAYLAND)
-    find_package(ECM REQUIRED NO_MODULE)
-    list(APPEND CMAKE_MODULE_PATH "${ECM_MODULE_PATH}")
 
-    find_package(Wayland REQUIRED Client Cursor Egl)
-    find_package(WaylandScanner REQUIRED)
-    find_package(WaylandProtocols 1.15 REQUIRED)
+    include(FindPkgConfig)
+    pkg_check_modules(Wayland REQUIRED
+        wayland-client>=0.2.7
+        wayland-cursor>=0.2.7
+        wayland-egl>=0.2.7
+        xkbcommon)
 
     list(APPEND glfw_PKG_DEPS "wayland-client")
 
     list(APPEND glfw_INCLUDE_DIRS "${Wayland_INCLUDE_DIRS}")
-    list(APPEND glfw_LIBRARIES "${Wayland_LIBRARIES}")
-
-    find_package(XKBCommon REQUIRED)
-    list(APPEND glfw_INCLUDE_DIRS "${XKBCOMMON_INCLUDE_DIRS}")
+    list(APPEND glfw_LIBRARIES "${Wayland_LINK_LIBRARIES}")
 
     include(CheckIncludeFiles)
     include(CheckFunctionExists)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -45,30 +45,44 @@ if (_GLFW_X11 OR _GLFW_WAYLAND)
 endif()
 
 if (_GLFW_WAYLAND)
-    ecm_add_wayland_client_protocol(GLFW_WAYLAND_PROTOCOL_SOURCES
-        PROTOCOL
-        "${WAYLAND_PROTOCOLS_PKGDATADIR}/stable/xdg-shell/xdg-shell.xml"
-        BASENAME xdg-shell)
-    ecm_add_wayland_client_protocol(GLFW_WAYLAND_PROTOCOL_SOURCES
-        PROTOCOL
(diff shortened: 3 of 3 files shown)


CMakeLists.txt
```diff
@@ -191,20 +191,18 @@ endif()
...
 
-    find_package(Wayland REQUIRED Client Cursor Egl)
-    find_package(WaylandScanner REQUIRED)
-    find_package(WaylandProtocols 1.15 REQUIRED)
+    include(FindPkgConfig)
+    pkg_check_modules(Wayland REQUIRED
+        wayland-client>=0.2.7
+        wayland-cursor>=0.2.7
+        wayland-egl>=0.2.7
+        xkbcommon)
```
gracicot, line 200: Is there a reason to change from CMake packages to pkg-config? CMake packages also supports specifying the minimum version.

src/CMakeLists.txt
```diff
@@ -45,30 +45,44 @@ if (_GLFW_X11 OR _GLFW_WAYLAND)
...
-        PROTOCOL
-        "${WAYLAND_PROTOCOLS_PKGDATADIR}/unstable/pointer-constraints/pointer-constraints-unstable-v1.xml"
-        BASENAME pointer-constraints-unstable-v1)
-    ecm_add_wayland_client_protocol(GLFW_WAYLAND_PROTOCOL_SOURCES
-        PROTOCOL
-        "${WAYLAND_PROTOCOLS_PKGDATADIR}/unstable/idle-inhibit/idle-inhibit-unstable-v1.xml"
-        BASENAME idle-inhibit-unstable-v1)
+    find_program(WAYLAND_SCANNER_EXECUTABLE NAMES wayland-scanner)
+    pkg_check_modules(WAYLAND_PROTOCOLS REQUIRED wayland-protocols>=1.15)
+    pkg_get_variable(WAYLAND_PROTOCOLS_BASE wayland-protocols pkgdatadir)
```
gracicot, line 50: Same here. Is there any advantages using pkg-config instead of the CMake native packages?

CMakeLists.txt
jwinarske: It works by default for a host build, easy to cross compile on host (set PKG_CONFIG_DIR env variable), and it doesn't require an extra dependency.

src/CMakeLists.txt
jwinarske: The primary advantage is that it doesn't require having to install ECM, which is not provided as a package for most distros, nor is it included in the cmake as a dependent repo.  This directly improves the developer experience.  The alternative would be to add ECM as a dependent repo, and pull it "automajically", this is a bit clunky as it only applies to Wayland.  So the trade off is use pkg config and drop the ECM dependency.  Which works as expected.

gracicot: Noted, thanks.

src/CMakeLists.txt
```diff
@@ -45,30 +45,44 @@ if (_GLFW_X11 OR _GLFW_WAYLAND)
...
+              COMMAND ${WAYLAND_SCANNER_EXECUTABLE} private-code
+              < ${protocol_file} > ${output_file}.c
+              DEPENDS ${protocol_file})
+          list(APPEND GLFW_WAYLAND_PROTOCOL_SOURCES ${output_file}.c)
+    endmacro()
+
+    set(GLFW_WAYLAND_PROTOCOL_SOURCES)
+    wayland_generate(
+        ${WAYLAND_PROTOCOLS_BASE}/stable/xdg-shell/xdg-shell.xml
+        ${CMAKE_SOURCE_DIR}/src/wayland-xdg-shell-client-protocol)
```
linkmauve, line 69: You should never put generated files in the repository itself if possible, the user might build it from a read-only filesystem or anything like that. Instead, keep it in `${CMAKE_BINARY_DIR}/src/` as it does currently.

.gitignore
```diff
@@ -52,10 +52,13 @@ src/glfw_config.h
...
-src/wayland-pointer-constraints-unstable-v1-client-protocol.h
-src/wayland-pointer-constraints-unstable-v1-protocol.c
-src/wayland-relative-pointer-unstable-v1-client-protocol.h
-src/wayland-relative-pointer-unstable-v1-protocol.c
+src/wayland-pointer-constraints-unstable-v1-client-protocol.*
+src/wayland-relative-pointer-unstable-v1-client-protocol.*
+src/wayland-idle-inhibit-unstable-v1-client-protocol.*
+src/wayland-viewporter-client-protocol.*
+src/wayland-xdg-decoration-client-protocol.*
+src/wayland-xdg-shell-client-protocol.*
```
linkmauve, line 60: That way you can drop those.

src/CMakeLists.txt
```diff
@@ -45,30 +45,44 @@ if (_GLFW_X11 OR _GLFW_WAYLAND)
...
+
+    macro(wayland_generate protocol_file output_file)
+          add_custom_command(OUTPUT ${output_file}.h
+              COMMAND ${WAYLAND_SCANNER_EXECUTABLE} client-header
+              < ${protocol_file} > ${output_file}.h
+              DEPENDS ${protocol_file})
+          list(APPEND GLFW_WAYLAND_PROTOCOL_SOURCES ${output_file}.h)
+
+          add_custom_command(OUTPUT ${output_file}.c 
+              COMMAND ${WAYLAND_SCANNER_EXECUTABLE} private-code
```
linkmauve, line 60: This commit also fixes the issue of ECM building it as `code` instead of `private-code`, and thus generating warnings.

linkmauve (changes requested):
Thanks a lot for this commit, it’s a very welcome change to drop this dependency!

Btw, which distributions don’t ship ECM?  In my experience all of ArchLinux, Debian, Ubuntu, Fedora, OpenSuse, etc. do provide it.

jwinarske:
Well I was using the wrong package name...  It is in fact there :)  I rarely use the CMake that comes with a distro, so I would have to manually override it either way.  I'll update PR in a couple of hours.

linkmauve (approved):

linkmauve:
Thanks, and merged!

PR #1779, opened 2020-10-03
Wayland: Remove FindWaylandProtocols.cmake and FindXKBCommon.cmake

caramelli:
With the removal of the ECM (Extra CMake Modules) dependency, FindXKBCommon.cmake and FindWaylandProtocols.cmake can be removed.

Nicolas Caramelli


linkmauve:
Indeed, thanks!

PR #1781, opened 2020-10-04
triangle-vulkan test: Update shader code

caramelli:
Updated shader code with the output of the glslangValidator tool to avoid the SPIRV error "Vulkan spec doesn't allow BuiltIn VertexId/InstanceId to be used."

Nicolas Caramelli


elmindreda (approved):

elmindreda:
Thank you! This has been merged as 8d47dc5894623bc7a1447236fd9ba62ecd70bc97.

PR #1782, opened 2020-10-08
create missing directory and use proper binary folder

jwinarske:
Fixes item misses in last PR.
- If ran on clean tree, it would fail as nothing was creating the src directory.
- If using add_directory scheme, build fails as source files were going to incorrect folder.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -50,38 +50,38 @@ if (_GLFW_WAYLAND)
     pkg_get_variable(WAYLAND_PROTOCOLS_BASE wayland-protocols pkgdatadir)
 
     macro(wayland_generate protocol_file output_file)
-          add_custom_command(OUTPUT ${output_file}.h
-              COMMAND ${WAYLAND_SCANNER_EXECUTABLE} client-header
-              < ${protocol_file} > ${output_file}.h
-              DEPENDS ${protocol_file})
-          list(APPEND GLFW_WAYLAND_PROTOCOL_SOURCES ${output_file}.h)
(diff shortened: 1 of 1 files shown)


elmindreda:
I can verify the second issue you list but not the first one.

jwinarske:
Try creating a new build folder.  There was certainly a case I hit where it failed to copy due to a non-existent folder

elmindreda:
Hmm. Tried it again now. It seems the `${GLFW_BINARY_DIR}/src` directory is created during CMake generation when files like `glfw3.pc` and `glfw_config.h` are created, while the Wayland proxy sources are generated as part of the build, which is necessarily later.

I will merge the binary directory fixes in the meantime. Thank you for those!

elmindreda:
All the path handling in the previous PR used unquoted variable references, so if there was whitespace in any path then that may explain why you got those errors.

jwinarske:
In my case there were no white spaces in the path (only if I need to build something on Windows, which aside from quotes, will also most likely require path conversion).  When using "add_custom_command" the target directory isn't created auto-magically, it needs to be created manually.  If you attempt to build this project standalone, it builds fine as the folder required for the "add_custom"command" exists.  My repro case adds glfw3 via "add_subdirectory", opposed to "ExternalProject_Add".  This should not repro for standalone or "ExternalProject_Add" cases.

elmindreda:
Still can't reproduce this. I tried with the [starter project](https://github.com/juliettef/GLFW-CMake-starter) now, which adds GLFW as a subdirectory. Once I'd applied the binary directory fixes above it works as expected. Tested with the make and Ninja generators.

PR #1784, opened 2020-10-14
Wayland: Assume 96 DPI if physical size is <= 0

caramelli:
This fallback is similar to the one performed in the "X11: Assume 96 DPI if RandR Monitor Size is zero" commit: https://github.com/glfw/glfw/commit/e96dc5d2199032d565b55c4177820ab28b1d67c9

Nicolas Caramelli

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -1038,6 +1038,8 @@ int _glfwPlatformInit(void)
     char *cursorSizeEnd;
     long cursorSizeLong;
     int cursorSize;
+    int i;
+    _GLFWmonitor* monitor;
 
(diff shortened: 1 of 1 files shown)


linkmauve:
Thanks! This indeed helps on projectors and other outputs which don’t have an intrinsic size.

elmindreda:
Would it be possible to do this in `outputHandleGeometry` in `wl_monitor.c` instead of having it as a separate pass?

caramelli:
The outputHandleGeometry() callback is called before the outputHandleMode() callback which provides width and height
But yes we can do something in the outputHandleMode() callback, that's another solution!
I created another pull request if you prefer to use this method: https://github.com/glfw/glfw/pull/1792

PR #1792, opened 2020-10-19
Wayland: Move the 96 DPI assumption into the outputHandleMode() callback

caramelli:


linkmauve:
Hi, sorry for not reviewing this PR before it was merged, but it relies on the assumption that the `mode` event will be sent after the `geometry` event on a given `wl_output`, which is all but guaranteed.

The proper way to do so would be to store these properties in a temporary struct, and apply them in the `done` handler (if the protocol version is ≥ 2, otherwise to do it every time a `mode` or `geometry` event is received) so that it can react to live changes. Those are arguably quite rare, a normal monitor doesn’t often change its physical size for instance, but it could perfectly do in a VR setup, or in other kinds of non-conventional monitors.

caramelli:
If we consider Weston (reference implementation of a Wayland compositor), the outputHandleGeometry() callback is called before the outputHandleMode():

	wl_output_send_geometry(resource,
				output->x,
				output->y,
				head->mm_width,
				head->mm_height,
				head->subpixel,
				head->make, head->model,
				output->transform);
	if (version >= WL_OUTPUT_SCALE_SINCE_VERSION)
		wl_output_send_scale(resource,
				     output->current_scale);

	wl_list_for_each (mode, &output->mode_list, link) {
		wl_output_send_mode(resource,
				    mode->flags,
				    mode->width,
				    mode->height,
				    mode->refresh);
	}

Yes, the PR is based on this consideration.

linkmauve:
[The protocol](https://gitlab.freedesktop.org/wayland/wayland/blob/master/protocol/wayland.xml#L2448) is the source of truth defining what an acceptable behaviour can be, and I couldn’t find any mention of an order there, except that `done` gets emitted after all the other ones are sent, so I’d advise not relying on implementation details of a single compositor.

caramelli:
OK, so we can do something in the outputHandleDone() callback as you suggest.
Wayland versions without "done" event support are very old (this raises me questions about CMakeLists.txt: the pkg_check_modules for Wayland with version 0.2.7 is a strange version number ...)

I have tested the following code:

static void outputHandleDone(void* data, struct wl_output* output)
{
    struct _GLFWmonitor *monitor = data;
    int i;

    for (i = 0; i < monitor->modeCount; ++i)
    {
        if (i == monitor->wl.currentMode && (monitor->widthMM <= 0 || monitor->heightMM <= 0))
        {
            // If Wayland does not provide a physical size, assume the default 96 DPI
            monitor->widthMM  = (int) (monitor->modes[monitor->wl.currentMode].width * 25.4f / 96.f);
            monitor->heightMM = (int) (monitor->modes[monitor->wl.currentMode].height * 25.4f / 96.f);
            break;
        }
    }

    _glfwInputMonitor(monitor, GLFW_CONNECTED, _GLFW_INSERT_LAST);
}

Maybe we need to do a new PR

linkmauve:
This one looks more correct, thanks!

elmindreda:
I hope b925a54ef11adab250e4e9592afe918c1f8caebb is a step in the right direction.

higfxback:
OK, I tested the code corresponding to this last commit, no issue for me!

Nicolas Caramelli

PR #1793, opened 2020-10-23
Init hint to pick VK_KHR_xlib_surface instead of VK_KHR_xcb_surface

caramelli:
If libX11-xcb.so is available on the system, the VK_KHR_xcb_surface extension is used and there is no way to force the use of the VK_KHR_xlib_surface extension.
This change adds the GLFW_NO_XCB environment variable to do this, but a better workaround might exist.

Nicolas Caramelli

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1250,6 +1250,11 @@ extern "C" {
  *  macOS specific [init hint](@ref GLFW_COCOA_MENUBAR_hint).
  */
 #define GLFW_COCOA_MENUBAR          0x00051002
+/*! @brief X11 specific init hint.
+ *
+ *  X11 specific [init hint](@ref GLFW_X11_XCB_hint).
+ */
+#define GLFW_X11_XCB                0x00051003
 /*! @} */
 
 #define GLFW_DONT_CARE              -1
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1252,9 +1252,9 @@ extern "C" {
 #define GLFW_COCOA_MENUBAR          0x00051002
 /*! @brief X11 specific init hint.
  *
- *  X11 specific [init hint](@ref GLFW_X11_XCB_hint).
+ *  X11 specific [init hint](@ref GLFW_X11_XCB_VULKAN_SURFACE_hint).
  */
-#define GLFW_X11_XCB                0x00051003
+#define GLFW_X11_XCB_VULKAN_SURFACE 0x00052001
 /*! @} */
 
 #define GLFW_DONT_CARE              -1
--- a/src/internal.h
+++ b/src/internal.h
@@ -248,7 +248,9 @@ struct _GLFWinitconfig
         GLFWbool  menubar;
         GLFWbool  chdir;
     } ns;
-    GLFWbool      x11_xcb;
+    struct {
+        GLFWbool  xcbVulkanSurface;
+    } x11;
 };
 
 // Window configuration
--- a/src/internal.h
+++ b/src/internal.h
@@ -248,6 +248,7 @@ struct _GLFWinitconfig
         GLFWbool  menubar;
         GLFWbool  chdir;
     } ns;
+    GLFWbool      x11_xcb;
 };
 
 // Window configuration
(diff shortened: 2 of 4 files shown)


caramelli:
Reading through the internal implementation of glfw, it seems to me that using a hint instead of an environment variable would be quite appropriate. So I updated the change.

Nicolas Caramelli

elmindreda:
Is there an issue with VK_KHR_xcb_surface that this is working around or is it a matter of preference?

caramelli:
The uses of VK_KHR_xcb_surface and VK_KHR_xlib_surface in GLFW are both fine: VK_KHR_xcb_surface is used by default (if libX11-xcb.so is available on the system), and there is no issue with that!

But VK_KHR_xcb_surface and VK_KHR_xlib_surface have a specific implementation in the Vulkan ICD (Installable Client Driver), with sometimes different behaviors or performances depending on the internal implementation of the ICD. Yes, the purpose of the proposed change is to have the ability to select or test the VK_KHR_xlib_surface extension available in the ICD.

Nicolas Caramelli

caramelli:
Do you still have reservations about this change?

elmindreda:
I'm concerned that the hint name is too generic. Perhaps something like `GLFW_X11_XCB_VULKAN_SURFACE`?

caramelli:
Thanks for your comment.
In the current implementation, XCB is only used to create the Vulkan surface, but yes, XCB could be used for other things. 

OK, I updated the hint name with GLFW_X11_XCB_VULKAN_SURFACE

include/GLFW/glfw3.h
```diff
@@ -1250,6 +1250,11 @@ extern "C" {
  *  macOS specific [init hint](@ref GLFW_COCOA_MENUBAR_hint).
  */
 #define GLFW_COCOA_MENUBAR          0x00051002
+/*! @brief X11 specific init hint.
+ *
+ *  X11 specific [init hint](@ref GLFW_X11_XCB_VULKAN_SURFACE_hint).
+ */
+#define GLFW_X11_XCB_VULKAN_SURFACE 0x00051003
```
elmindreda, line 1257: Please open a new sub-range for X11, i.e. `0x00052001`.

src/x11_init.c
```diff
@@ -813,11 +813,15 @@ static GLFWbool initExtensions(void)
...
     }
 
+    if (_glfw.hints.init.x11_xcb_vulkan_surface)
+    {
 #if defined(__CYGWIN__)
-    _glfw.x11.x11xcb.handle = _glfw_dlopen("libX11-xcb-1.so");
+        _glfw.x11.x11xcb.handle = _glfw_dlopen("libX11-xcb-1.so");
 #else
-    _glfw.x11.x11xcb.handle = _glfw_dlopen("libX11-xcb.so.1");
+        _glfw.x11.x11xcb.handle = _glfw_dlopen("libX11-xcb.so.1");
```
elmindreda, line 821: Please replace any tab characters with spaces.

src/internal.h
```diff
@@ -248,6 +248,7 @@ struct _GLFWinitconfig
         GLFWbool  menubar;
         GLFWbool  chdir;
     } ns;
+    GLFWbool      x11_xcb_vulkan_surface;
```
elmindreda, line 251: Please use an anonymous struct `x11` and camel case like the sections above.

elmindreda (changes requested):
Only minor tweaks needed, otherwise it looks good and is ready to merge.

src/x11_init.c
caramelli: If I'm not wrong, these are already space characters: so I haven't changed this part.

src/internal.h
caramelli: Yes, of course. It should be OK.

caramelli:
If you see others changes required for this PR, do not hesitate to let me know!

src/x11_init.c
elmindreda: Sorry, my bad. I misread the diff.

caramelli:
GLFW_X11_XCB_VULKAN_SURFACE hint is now used in https://github.com/KhronosGroup/Vulkan-Samples
Thanks for having merged this!

Issue #1790, opened 2020-10-18
Cursor position snapping to center when CURSOR_DISABLED is set


tomc1998:
On Linux, Ubuntu 18

The `cursor` example shows this, commit 0ef149c8f2451fbc8e866834675a075cfc295b6c (current master, at the time of posting)

## Steps to reproduce:

Compile the library:

```
mkdir build && cd build
cmake ..
make -j
```

Run the example:

```
tests/cursor
```

Press 'D' to 'disable' the cursor (hide the cursor & 'grab' it into the window)

Every second or so, the cursor position snaps to the center of the window. This sometimes doesn't happen - i'm not sure what causes this. Not moving the mouse at all stops the snapping - the snapping only happens after / during a mouse movement.

This is a pretty brutal bug, as any 1st person 'mouselook' applications are broken under this. There's no workaround, other than ignoring all mouse inputs in the center of the screen, because there's no way to hide the mouse & 'do it yourself' (as the mouse will just get stuck on the edges of the screen, afaik there's no way to get mouse movement deltas)

I'm using the tiling window manager 'i3', however I have disabled 'mouse warping' (the behaviour that snaps the cursor to the center of the window when a window is focused) with `mouse_warping none` in my I3 config - the behaviour persists.

Is there any workaround for this at least?

Thanks!

elmindreda:
Please post the GLFW version string returned by `glfwGetVersionString` and, if possible, some output from the test program that shows the error.

tomc1998:
```
5.947: Cursor position: 320.000000 474.000000 (+0.000000 -1.000000)
5.964: Cursor position: 320.000000 473.000000 (+0.000000 -1.000000)
5.964: Cursor position: 320.000000 472.000000 (+0.000000 -1.000000)
5.980: Cursor position: 320.000000 471.000000 (+0.000000 -1.000000)
5.997: Cursor position: 320.000000 470.000000 (+0.000000 -1.000000)
5.997: Cursor position: 320.000000 469.000000 (+0.000000 -1.000000)
6.014: Cursor position: 320.000000 530.000000 (+0.000000 +61.000000)
```

Here's some output, you can see that the cursor snaps back to 320,530 after a short time, resulting in a huge delta

Version string is `3.4.0 X11 GLX EGL OSMesa clock_gettime evdev`

I also just noticed I got these errors as it starts running, using the 'cursor' example in the glfw repo:

```
Error: X11: Standard cursor shape unavailable
[...]

PR #1795, opened 2020-10-25
Added dumb fix for disabled cursor snapping to center

tomc1998:
This is a fix for https://github.com/glfw/glfw/issues/1790

The cursor callback was being called, snapping the cursor back to the center of the window, when the cursor was disabled.

This PR disables calling the cursor callback for 'EnterNotify' X11 events, which seemed to be the root of the snapping. This seems to fix all the issues I was having.

Unsure whether this is the correct solution, since I don't understand the purpose of the GLFW cursor warping - with this change, the virtual cursor position will grow and grow to quite absurd sizes. This might not be the correct fix!

### DIFF
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -1452,7 +1452,11 @@ static void processEvent(XEvent *event)
                 updateCursorImage(window);
 
             _glfwInputCursorEnter(window, GLFW_TRUE);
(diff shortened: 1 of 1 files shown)


elmindreda:
Thank you, will review.

Issue #1783, opened 2020-10-09
USE_MSVC_RUNTIME_LIBRARY_DLL unnecessary for cmake 3.15 and above


jpr89:
In cmake 3.15 "CMAKE_MSVC_RUNTIME_LIBRARY" was introduced to solve the problem that GLFW's USE_MSVC_RUNTIME_LIBRARY_DLL  addresses.

Cmake Documentation:
https://cmake.org/cmake/help/latest/release/3.15.html#variables
https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#variable:CMAKE_MSVC_RUNTIME_LIBRARY

I propose 2 things.

1.)
If the client is using cmake 3.15 or above then don't bother making this variable at all. It really only confuses things since the official cmake approach will override whatever GLFW does anyway. This creates unnecessary build complexity/confusion.
Since users will have a cmake-gui option that doesn't do anything.

2.)
If this variable continues to exist prefix it with GLFW_* to make it clear what project it belongs to. In cmake-gui/ccmake you can order cache variables by Groups (Groups essentially just be prefixes). You can see what I mean in the picture below.
 
![image](https://user-images.githubusercontent.com/31327577/95607904-62f41680-0a2a-11eb-89f3-3702caaeed66.png)

In larger cmake projects, not using grouped entries confuses regular developers. Personally the first time I saw it, I thought it was part of cmake like BUILD_SHARED_LIBS.

elmindreda:
> If the client is using cmake 3.15 or above then don't bother making this variable at all. It really only confuses things since the official cmake approach will override whatever GLFW does anyway. This creates unnecessary build complexity/confusion.
Since users will have a cmake-gui option that doesn't do anything.

Agreed, will fix.

> If this variable continues to exist prefix it with GLFW_* to make it clear what project it belongs to. In cmake-gui/ccmake you can order cache variables by Groups (Groups essentially just be prefixes). You can see what I mean in the picture below.

The name was picked because a number of other projects are also using it, so it may be a good idea to keep the current name.

hdf89shfdfs:
"The name was picked because a number of other projects are also using it, so it may be a good idea to keep the current name."

Out of curiosity what other projects? The only I'm aware that does this is GLFW.

hdf89shfdfs:
I've created a pull request:
https://github.com/glfw/glfw/pull/1796

PR #1796, opened 2020-10-26
Modernize MSVC runtime library code

jpr89:
This commit is in response to this issue:
https://github.com/glfw/glfw/issues/1783


YMba9g8j9CJp0wLoQf5y:
Any update on this?

hdf89shfdfs:
Is there anything I need to do to make this change more palatable @elmindreda ?

elmindreda:
@hdf89shfdfs Sorry, my arm stopped working for a month. I like this solution a lot, except possibly that half of the logic is in a separate, new file.

hdf89shfdfs:
@elmindreda hope you feel better.

I understand what you mean but I'm not really a fan of having all the logic inside the main CMakeLists.txt since I think it makes reading the cmake harder.

Generally in my projects I like having a generic helper module but I didn't see one that already existed and looking in your cmake modules folder they all seem to be single purpose so I thought it fit the mold.

What about having a cmake module glfw.cmake? And put helper functions/macros inside there instead?

hdf89shfdfs:
I've created a new commit showing what a more general cmake helper module would look like.

hdf89shfdfs:
Please let me know what you think @elmindreda

hdf89shfdfs:
Hey @elmindreda  do you have any idea why the linux builds would be failing? I'm struggling to figure it out.

CMakeLists.txt
```diff
@@ -1,20 +1,14 @@
 cmake_minimum_required(VERSION 3.1...3.17 FATAL_ERROR)
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/Cmake/")
```
gracicot, line 3: Your problem is here. It works on windows because filesystems are case insensitive, but on Linux they are. Do on Linux this line is meaningless since `Cmake` is not a directory, but `CMake` is. Either way this is a typo.

hdf89shfdfs: Thank you! I'll update this right away!

hdf89shfdfs:
@elmindreda let me know if this change is acceptable. 

I really think it helps the readability quite a lot :)

But that's an opinion. The main thing is to fix the problem.

If the new cmake module isn't acceptable I can move all of this logic into the main Cmakelists.txt

hdf89shfdfs:
> 
> 
> @elmindreda let me know if this change is acceptable.
> 
> I really think it helps the readability quite a lot :)
> 
> But that's an opinion. The main thing is to fix the problem.
> 
> If the new cmake module isn't acceptable I can move all of this logic into the main Cmakelists.txt

Any update on this?

moritz-h:
I have found this PR, while having the problem, that the current "regex replace" based method does not work anymore when using modern CMAKE_MSVC_RUNTIME_LIBRARY in a top level project. I think it is a nice start, but also I see a few problems with the current implementation of this PR.

- Please don't force a user to globally set CMAKE_MSVC_RUNTIME_LIBRARY by throwing a warning. I think it should be perfectly fine if a user set CMAKE_MSVC_RUNTIME_LIBRARY as target property instead of globally.
- Please don't force static linking in the standalone case. The standalone case may be used to build GLFW standalone to later use find_package from another project. Therefore the user should be able to select linkage type.

I would suggest to keep the USE_MSVC_RUNTIME_LIBRARY_DLL option for all CMake versions, but for CMake 3.15+, instead of changing the CMAKE_C_FLAGS the CMAKE_MSVC_RUNTIME_LIBRARY property can be set on the glfw target.

elmindreda:
@moritz-h Agreed, that seems to be the better solution. I've also added a warning for when the `USE_MSVC_RUNTIME_LIBRARY_DLL` option has no effect.

PR #1798, opened 2020-10-28
wl: Fix destroying CSDs in the correct order

LeonardKoenig:
On Wayland we implement Client-Side Decorations if the compositors do not
implement SSDs.  In that case, the destructors of the surfaces were called
in the wrong order, leading to a dereference of an already freed object.
We need to first destroy the subsurface before destroying the parent surface.

Related PR on kitty: https://github.com/kovidgoyal/kitty/pull/3066
Related issue on kitty: https://github.com/kovidgoyal/kitty/issues/3051

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ information on what to include when reporting a bug.
  - [Wayland] Bugfix: Retrieving partial framebuffer size would segfault
  - [Wayland] Bugfix: Scrolling offsets were inverted compared to other platforms
    (#1463)
+ - [Wayland] Bugfix: Client-Side Decorations were destroyed in the wrong worder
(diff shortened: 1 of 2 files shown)


christianrauch:
I think that the currently implemented CSD are unusable, at least on GNOME Shell. I proposed https://github.com/glfw/glfw/pull/1693 to implement "proper" client-side decorations on Wayland, including the usual functionality like min/max/close and additional fixes for resizing glitches.

Luflosi:
I fully agree with you and would love to see your PR merged. However think it would be best to merge this simple fix first since it shouldn't take much time to review.

elmindreda (approved):

PR #1801, opened 2020-11-02
docs: fix simple typo, collapseable -> collapsable

timgates42:
There is a small typo in deps/nuklear.h.

Should read `collapsable` rather than `collapseable`.


linkmauve:
I’d even go for “collapsible”, which seems to be recommended by various dictionaries.

Issue #1648, opened 2020-02-18
menubar not clickable on macOS Catalina until it lost and regained focus.


zhuah:
It may related to #1165 and  this [commit](https://github.com/glfw/glfw/commit/9aa15aa710ff5adbdbccb124da96864df84b2479)

andydotxyz:
This is an issue that I was hoping to solve for Fyne as well (fyne-io/fyne#505) but I can't even figure how to begin debugging.

elmindreda:
If possible, please bundle your macOS GUI applications.

That said, this is clearly a bug and hopefully it can be fixed.

andydotxyz:
I completely agree. This is really only a developer facing issue. But I still wnated to try and help fix it ;)

PR #1802, opened 2020-11-05
Fix for #1648: menubar not clickable on macOS Catalina

richardwilkes:
NSApp setActivationPolicy was being called too soon when the app was not
bundled and launched from the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ information on what to include when reporting a bug.
    (#1635)
  - [Cocoa] Bugfix: Failing to retrieve the refresh rate of built-in displays
    could leak memory
+ - [Cocoa] Bugfix: Menubar was not clickable on macOS 10.15+ until it lost and
+   regained focus (#1648)
  - [X11] Bugfix: The CMake files did not check for the XInput headers (#1480)
  - [X11] Bugfix: Key names were not updated when the keyboard layout changed
    (#1462,#1528)
@@ -428,6 +430,7 @@ skills.
  - Waris
  - Jay Weisskopf
  - Frank Wille
+ - Richard A. Wilkes
  - Tatsuya Yatagawa
  - Ryogo Yoshimura
  - Lukas Zanner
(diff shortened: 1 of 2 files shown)


andydotxyz:
Tested this and it fixes the issue for me thanks!

elmindreda:
Thank you for this!

Is this behavior documented anywhere? Would it be safe to set the policy from `applicationDidFinishLaunching:`? It appears to work, but Cocoa.

richardwilkes:
The (bad) behavior seemed to be introduced in macOS Catalina. There was no documentation at the time as to why it changed nor have I found any since. As for trying to make the call from applicationDidFinishLaunching, I did try that and it didn't have the desired effect for some reason.

elmindreda:
Thanks! I'll merge the current version and we'll hope for clarity later.

elmindreda (approved):

elmindreda:
The `dispatch_get_main_queue` function does not appear to be available on 10.8. Does it work for you to instead call `setActivationPolicy:` after `run` in `_glfwPlatformInit`?

richardwilkes:
Looks like it wasn't introduced until 10.10. :-(

I believe all that really needs to be done is to have that call to setActivationPolicy run on the event thread, after the event loop is executing.

elmindreda:
cdc3b71a90434fae3255edf4f152da0971733499 is my proposal for a backwards compatible fix.

richardwilkes:
Hmm... I'm not sure that will actually fix the problem. Earlier experiments showed it had to be done as the result of an event coming in on the event loop, not anywhere else. I had also tried on some of the callbacks, like applicationDidFinishLaunching, to no avail.

Have you tested this change and seen it work both incorrectly without and then correctly with?

elmindreda:
> Have you tested this change and seen it work both incorrectly without and then correctly with?

Yup, the bug is consistently reproducible on my machine and cdc3b71a90434fae3255edf4f152da0971733499 Works on My Machine™ when launched from a terminal, Finder and Xcode.

richardwilkes:
Fantastic. Glad to hear that works.

elmindreda:
Anyone following this thread, please test cdc3b71a90434fae3255edf4f152da0971733499 and report the results.

andydotxyz:
yes that works for me too.

PR #1804, opened 2020-11-18
Machine shutdown callback

torkeldanielsson:
Adds a machine shutdown callback. Windows only at the moment. (Anyone know if the API for this on linux is easy? Haven't looked into adding it on linux at all, sorry...)

It was added to our (voysys.com) fork of glfw due to a customer request. I thought I would drop a PR here in the upstream, maybe this is functionality that someone else can find useful?

From the included addition to the documentation:
```
If you wish to be notified when the user attempts to shut down the machine, or 
interrupt a shutdown, then set a machine shutdown callback.

@code
glfwSetMachineShutdownCallback(window, machine_shutdown_callback);
@endcode

The callback function is called when GLFW detects that the machine is shutting down.
It can be used for example to save data to disk in order to minimize risk of data loss.
Another use is to interrupt the shutdown (return false).

@code
int machine_shutdown_callback(GLFWwindow* window)
{
    initiate_save_of_important_data_to_disk();
    if (prevent_machine_shutdown())
        return GLFW_FALSE;
    else
        return GLFW_TRUE;
}
@endcode
```

### DIFF
--- a/docs/window.dox
+++ b/docs/window.dox
@@ -618,8 +618,8 @@ void window_close_callback(GLFWwindow* window)
 }
 @endcode
 
-If you wish to be notified when the user attempts to shut down the machine, set a 
-machine shutdown callback.
+If you wish to be notified when the user attempts to shut down the machine, or 
+interrupt a shutdown, then set a machine shutdown callback.
 
 @code
 glfwSetMachineShutdownCallback(window, machine_shutdown_callback);
@@ -627,11 +627,16 @@ glfwSetMachineShutdownCallback(window, machine_shutdown_callback);
 
 The callback function is called when GLFW detects that the machine is shutting down.
 It can be used for example to save data to disk in order to minimize risk of data loss.
+Another use is to interrupt the shutdown (return false).
 
 @code
-void machine_shutdown_callback(GLFWwindow* window)
+int machine_shutdown_callback(GLFWwindow* window)
 {
     initiate_save_of_important_data_to_disk();
+    if (prevent_machine_shutdown())
+        return GLFW_FALSE;
+    else
+        return GLFW_TRUE;
 }
 @endcode
 
--- a/docs/window.dox
+++ b/docs/window.dox
@@ -618,6 +618,23 @@ void window_close_callback(GLFWwindow* window)
 }
 @endcode
 
+If you wish to be notified when the user attempts to shut down the machine, set a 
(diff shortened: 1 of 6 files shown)


tombsar:
We could look into simply trapping the SIGTERM signal on Linux for this purpose?

tombsar:
I'm struggling to find up-to-date information on how to detect this on Linux. Will run some tests with `dbus-monitor` to see what messages get passed around. `SIGTERM` and `SIGKILL` will be sent to the program when it needs to terminate, but a) that can be for reasons other than system shutdown, and b) it's probably too late to do anything about it by that point.

torkeldanielsson:
> 
> 
> I'm struggling to find up-to-date information on how to detect this on Linux. Will run some tests with `dbus-monitor` to see what messages get passed around. `SIGTERM` and `SIGKILL` will be sent to the program when it needs to terminate, but a) that can be for reasons other than system shutdown, and b) it's probably too late to do anything about it by that point.

I looked a bit at it now too. Asking applications for permission before shutdown, and letting the applications interrupt the shutdown, seems to be a Windows thing. On Linux the signals being sent is more of a "you are to shut down, NOW". I would be hesitant to add signal listening to this - it doesn't seem like something that is done already by glfw?

tombsar:
Yes and no. At least historically, shutdowns on Linux were broadcast with a bit of warning, and users could cancel a pending shutdown by running `shutdown -c` in a shell. I'd expect there to be a systemd equivalent of this, but perhaps not.

I don't believe glfw does any signal handling at the moment (although `signal.h` does seem to be included), so it might be better to leave it alone.

At the moment, on Windows does shutting down the machine set the `glfwGetWindowShouldClose` flag? A compromise might be to add a function whereby the application can query what caused the window to need to close.

torkeldanielsson:
> At the moment, on Windows does shutting down the machine set the `glfwGetWindowShouldClose` flag? A compromise might be to add a function whereby the application can query what caused the window to need to close (e.g. "system shut down" vs "user action") instead of adding a new callback.

It does not set the glfwGetWindowShouldClose flag. Windows asks applications for permission to shut down, so an application can intercept a shutdown and cancel it (it is possible to change settings and disable this so the PC always shuts down regardless). This PR suggests using the return bool from the callback to signal whether to let the shutdown proceed or not.

PR #1824, opened 2020-12-30
Update CMakeLists.txt

atkurtul:
This was preventing building with clang on windows. Should work as well with clang-cl


hdf89shfdfs:
> 
> 
> This was preventing building with clang on windows. Should work as well with clang-cl

This was added to fix clang-cl on windows. You are re-introducing the issue.

https://github.com/glfw/glfw/commit/ac627706ef93362caf08bd3bf249656efd9151e8#diff-25a6634263c1b1f6fc4697a04e2b9904ea4b042a89af59dc93ec1f5d44848a26

atkurtul:
Clang with GNU interface treats forwards slashes as file paths. so there is no way to compile it with clang unless the cmake generator is Visual Studio.

elmindreda:
This should be fixed now with 061a0263a9783c1442ad96a061c717c167ab4a76. Thank you for reporting this!

PR #1836, opened 2021-01-18
8bitdo Controller mappings

pmprog:
Addition of 8bitdo controller mappings (pulled from Raylib, with a config update for M30)


elmindreda:
Thank you for your patience. These mappings should be included now with afa78b427028d416a5b79bdd13908412f6f80294, which pulled in the then current set of mappings from [SDL_GameControllerDB](https://github.com/gabomdq/SDL_GameControllerDB). If any are still missing, please submit them to that project and they will be included in our next mappings update.

PR #1859, opened 2021-03-05
Suggestion update minimum to 3.13

jpr89:
For the upcoming 3.4 release why not update the CMake minimum?

3.13 came out Nov 21, 2018

Currently CMake 3.20 is out. Glfw is 19 versions behind.

3.13 brings a lot of quality of life updates for users.


uthidata:
Note that this fix https://github.com/glfw/glfw/pull/1874 requires at least CMake 3.15 to work, so if the minimum CMake version changes at all, I think it should change to 3.15.

Issue #1145, opened 2017-11-23
fatal error C1091: compiler limit: string exceeds 65535 bytes in length


chaoticbob:
Not sure if this was present before, but I just ran into this error:
````
C:\code\glfw\src\mappings.h(302): fatal error C1091: compiler limit: string exceeds 65535 bytes in length
````

Haven't found a good workaround yet. 

Using Visual Studio 2015 and target platform is x64. 

**Update:** just looked at the history of the file and the changes appear to be recent?

elmindreda:
Fix up shortly.

chaoticbob:
Thanks for the super quick fix!

rayvling:
@elmindreda 
could you share how to fix up? thanks

PR #1868, opened 2021-03-21
Cache mappings on disk

linkmauve:
Parsing this file was taking the most time out of `glfwInit()`, or about 14ms on my PinePhone before the start of this series.  In the case where the string is already in cache, it now takes only 0.95ms.

It would be possible to make that even faster, and additionally to keep that page shared with other GLFW processes, by using `mmap()` instead, but for now I opted to not change the memory representation.

I also changed the mappings.h format to be a single `\n`-separated string instead of multiple different strings, so that we can call `glfwUpdateGamepadMappings()` once instead of once per entry, this alone was a ~2ms win.

Part of the implementation is shared with https://github.com/xkbcommon/libxkbcommon/pull/220

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -34,9 +34,11 @@
 #error "libxxhash is required for the compose cache"
 #endif
 
+#include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #define XXH_INLINE_ALL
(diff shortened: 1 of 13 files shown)


linkmauve:
I’ve updated this PR with two new commits, using the `mmap()` approach suggested in the first comment.

tombsar:
> changed the mappings.h format to be a single \n-separated string instead of multiple different strings

Does this run the risk of generating overlong string literals that could be rejected by the compiler? I remember there was a previous issue related to that (https://github.com/glfw/glfw/issues/1145).

linkmauve:
> > changed the mappings.h format to be a single \n-separated string instead of multiple different strings
> 
> Does this run the risk of generating overlong string literals that could be rejected by the compiler? I remember there was a previous issue related to that (#1145).

Good catch, I didn’t realise I was basically reverting this change.  I’ve instead replaced it with a concatenation of all the strings just before calling `glfwUpdateGamepadMappings()`. :)

src/cache.c
```diff
@@ -0,0 +1,201 @@
...
+    int length;
+    char *hash_part;
+    XXH128_canonical_t hash;
+
+    /* Mostly copied over from xkb_context_include_path_append_default(). */
+    home = getenv("HOME");
+    xdg = getenv("XDG_CACHE_HOME");
+    if (xdg != NULL)
+    {
+        length = strlen(xdg) + strlen("/glfw/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX") + 1;
```
aloucks, line 92: Why is the filename `XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX` ?

linkmauve: That is just a placeholder, I could have used `+ 32` instead (maybe that’s more readable?), the actual filename gets generated from the hash of the input on lines 118-122.

linkmauve: I’ve just pushed a change with the `+ 32`, which is indeed less surprising.

elmindreda:
I don't think it's a good idea for GLFW to start caching data to disk if it can possibly be avoided. The recent improvement to the handling of mappings seem to have helped with performance somewhat.

If performance does become an issue I would prefer either delaying the parsing until we need the mappings, only parsing entries as needed or re-using the existing parser and have `mappings.h` contain pre-parsed `_GLFWmapping` structs.

PR #1874, opened 2021-03-28
Fix failure to build when using GCC-like Clang front-end

nmcclatchey:
At least on Windows, the Clang compiler has two front-end variants:
* clang-cl uses a command-line interface that imitates MSVC.
* clang and clang++ use a command-line interface that imitates GCC.

[CMake version 3.15 gained support for the latter variant of the Clang compiler](https://cmake.org/cmake/help/latest/release/3.15.html), but glfw contains a CMake file that uses `CMAKE_C_SIMULATE_ID` to detect the frontend; however, this variable is misleading when using clang on Windows with the GCC-like front-end. This pull request updates the test to use the (undocumented) variable `CMAKE_C_COMPILER_FRONTEND_VARIANT` when available (CMake 3.15+), which correctly identifies the command-line variant that must be used for compilation.

In short, this pull request permits building glfw on Windows build systems using clang and clang++, instead of clang-cl.


elmindreda:
This should be fixed now with 061a0263a9783c1442ad96a061c717c167ab4a76. Thank you for reporting this!

PR #1899, opened 2021-05-12
Fix Wayland cursor offset when the cursor shape changes

### LINKED ISSUE #1706: Non-arrow cursors are offset from the hotspot in Wayland
Not sure if there is a bug already filed on this, but it is a bit hard to notice.

**OS and version:** Ubuntu 20.04 or 18.04 running "Ubuntu on Wayland" desktop environment
**Release or commit:** libglfw3-wayland/focal,now 3.3.2-1 amd64
**Error messages:** There are no error messages or unusual events, but the minimal app mentioned below can easily show the problem. 

cursorswitch

Basically, when switching from standard arrow to I-beam, the I-beam does not seem to be aligned with the arrow's hotspot. The app allows changing the cursor shapes using the number keys.

```
#include <GLFW/glfw3.h>
#include <iostream>

using namespace std;

bool s_quit = false;

void keyCallback(GLFWwindow* window, int key, int scancode, int action, int mods)
{
    if (key == GLFW_KEY_Q && action == GLFW_PRESS)
    {
        s_quit = true;
    }
    if (key == GLFW_KEY_1)
    {
        auto cursor = glfwCreateStandardCursor(GLFW_ARROW_CURSOR);
        glfwSetCursor(window, cursor);
    }
    if (key == GLFW_KEY_2)
    {
        auto cursor = glfwCreateStandardCursor(GLFW_IBEAM_CURSOR);
        glfwSetCursor(window, cursor);
    }
    if (key == GLFW_KEY_3)
    {
        auto cursor = glfwCreateStandardCursor(GLFW_CROSSHAIR_CURSOR);
        glfwSetCursor(window, cursor);
    }
    if (key == GLFW_KEY_4)
    {
        auto cursor = glfwCreateStandardCursor(GLFW_HAND_CURSOR);
        glfwSetCursor(window, cursor);
    }
    if (key == GLFW_KEY_5)
    {
        auto cursor = glfwCreateStandardCursor(GLFW_HRESIZE_CURSOR);
        glfwSetCursor(window, cursor);
    }
    if (key == GLFW_KEY_6)
    {
        auto cursor = glfwCreateStandardCursor(GLFW_VRESIZE_CURSOR);
        glfwSetCursor(window, cursor);
    }
}

void cursorPosCallback(GLFWwindow* window, double xpos, double ypos)
{
    cout << "Cursor position is: " << xpos << ", " << ypos << endl;
}


int main()
{
    glfwInit();
    glfwWindowHint(GLFW_VISIBLE, GLFW_TRUE);
    GLFWwindow* window = glfwCreateWindow(1920, 1080, "My Title", glfwGetPrimaryMonitor(), nullptr);
    if (!window)
    {
        cerr << "Can't create window!" << endl;
        glfwTerminate();
        return 1;
    }

    glfwSetKeyCallback(window, keyCallback);
    glfwSetCursorPosCallback(window, cursorPosCallback);

    while (!s_quit)
    {
        glfwWaitEvents();
    }

    glfwDestroyWindow(window);
    glfwTerminate();

    return 0;
}
```

**Steps to Reproduce**
1. Compile the program above and link against the wayland version of `libglfw3`
2. Run the program, observe the cursor
3. Press 2 to change the cursor to an I-beam, observe its hotspot.
4. Press 1 to change back to the arrow, observe its hotspot.

**Expected behaviour**
The I-beam's hotspot should align with the arrow's tip

**Actual behaviour**
The I-beam is offset from the arrow's tip. This seems to happen with other shapes as well. This does NOT happen in X11, so it seems to be a Wayland issue.

warisb:
The wayland protocol spec
https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_pointer

States that `set_cursor` must be called with the serial number of the
`enter` event. However, GLFW is passing in the serial number of the latest
received event, which does not meet the protocol spec.

As a result, `set_cursor` calls were simply ignored by the compositor.

This fix complies with the protocol more closely by specifically caching the
`enter` event serial, and using it for all `set_cursor` calls.

Fixes #1706

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ information on what to include when reporting a bug.
  - [Wayland] Bugfix: Client-Side Decorations were destroyed in the wrong worder
    (#1798)
  - [Wayland] Bugfix: Monitors physical size could report zero (#1784,#1792)
+ - [Wayland] Bugfix: Non-arrow cursors are offset from the hotspot (#1706)
  - [POSIX] Bugfix: `CLOCK_MONOTONIC` was not correctly tested for or enabled
  - [NSGL] Removed enforcement of forward-compatible flag for core contexts
  - [NSGL] Bugfix: `GLFW_COCOA_RETINA_FRAMEBUFFER` had no effect on newer
@@ -455,7 +456,7 @@ skills.
  - Torsten Walluhn
  - Patrick Walton
  - Xo Wang
- - Waris
+ - Waris Boonyasiriwat
  - Jay Weisskopf
  - Frank Wille
  - Andy Williams
@@ -470,6 +471,8 @@ skills.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Koray Kilinc
+ - Joseph Chua
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -105,6 +105,7 @@ static void pointerHandleEnter(void* data,
 
     window->wl.decorations.focus = focus;
     _glfw.wl.serial = serial;
+    _glfw.wl.pointerEnterSerial = serial;
     _glfw.wl.pointerFocus = window;
 
     window->wl.hovered = GLFW_TRUE;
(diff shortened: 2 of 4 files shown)


warisb:
Note that we found that this bug affected cursor modes (Disable, Hidden, Normal) as well, so if anyone noticed a problem related to that, this PR should also fix it.

warriormaster12:
Only seems to work with Opengl but not with Vulkan

elmindreda:
This has been merged as e7758c506dfcf258566f866c761c1b6bd7d98011. Thank you for the fix!

PR #1908, opened 2021-06-01
Fix some keys not repeating in Wayland

koraykilinc:
According to https://xkbcommon.org/doc/current/group__components.html#ga9d7f998efeca98b3afc7c257bbac90a8, xkb_keymap_key_repeats requires keymap and keycode as input.

```
int xkb_keymap_key_repeats( struct xkb_keymap * keymap,
			    xkb_keycode_t key
)
```

However, in wl_init.c:inputChar, we are passing in xkb_keysym_t which was a type mismatch.

This results in some keys not repeating when they should and vice versa.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ information on what to include when reporting a bug.
  - [Wayland] Bugfix: Client-Side Decorations were destroyed in the wrong worder
    (#1798)
  - [Wayland] Bugfix: Monitors physical size could report zero (#1784,#1792)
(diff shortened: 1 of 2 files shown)


warisb:
I think this one is a no-brainer. Simply fixing a protocol misuse. I encourage maintainers to take a look asap.

elmindreda:
This has been merged as 216d5e8402513b582563d5b8433fefb449a1593e. Thank you for the fix!

Issue #1448, opened 2019-03-16
Ability to get title of GLFWwindow


Xydez:
I have a project where I need to get the title of a GLFWwindow, and I think this would be an easily implemented feature. I am unfortunately not very familiar with the inner workings of GLFW, so I can't implement it myself.

LucasGithub123:
For now, I would just use the window user pointer as a workaround. You can store a pointer to any data you want with [glfwSetWindowUserPointer](https://www.glfw.org/docs/latest/group__window.html#ga3d2fc6026e690ab31a13f78bc9fd3651) and retrieve it anytime later with [glfwGetWindowUserPointer](https://www.glfw.org/docs/latest/group__window.html#ga17807ce0f45ac3f8bb50d6dcc59a4e06).

It's probably best to put your data in a struct, so you can add more data to it later on, such as a version number or something.
Here's  an example program that stores the title and an int in the window user pointer, then when the user presses T, it retrieves the pointer and uses it to print the title to the terminal.

```C
#include <stdio.h>
#include <GLFW/glfw3.h>

typedef struct {
    char* title;
    int myVersion;
} MyWindowData;

void key_callback(GLFWwindow* window, int key, int scancode, int action, int mods) {
    if(action == GLFW_PRESS ) {
        if(key == GLFW_KEY_T) { // READ DATA HERE
            MyWindowData* data = (MyWindowData*)glfwGetWindowUserPointer(window);
            printf("%s version %d\n", data->title, data->myVersion);
        }else if(key == GLFW_KEY_Q) {
            glfwSetWindowShouldClose(window, 1); // type Q to quit
        }
    }
}

int main() {
    if (!glfwInit()) return 1;
    char title[] = "My Cool Game";
    GLFWwindow* window = glfwCreateWindow(1280, 720, title, NULL, NULL);
    glfwSetKeyCallback(window, key_callback);
    MyWindowData data = { title, 42 };
    glfwSetWindowUserPointer(window, &data); // STORE DATA HERE
    while( !glfwWindowShouldClose(window) ) {
        // render() or whatever...
        glfwSwapBuffers(window);
        glfwPollEvents();
    }
    glfwDestroyWindow(window);
    glfwTerminate();
}
```

elmindreda:
@Xydez It would not be difficult to add. May I ask why you need this?

Xydez:
[...]

PR #1909, opened 2021-06-03
Add get window title function [help needed]

Bayemite:
This is addressing issue #1448, adding a function that queries the window title.
The implementation prefers calls to platform specific functions where available.

TODO:
It would be nice if someone could help out with the Cocoa and Wayland implementations- I have no access to these platforms.


src/win32_window.c
```diff
@@ -1462,6 +1462,41 @@ void _glfwPlatformDestroyWindow(_GLFWwindow* window)
         DestroyIcon(window->win32.smallIcon);
 }
 
+char* _glfwPlatformGetWindowTitle(_GLFWwindow* window)
+{
+    int count = GetWindowTextLengthW(window->win32.handle);
+    if(count == 0)
+    {
+        int error;
+        SetLastError(0);
```
intrigus, line 1471: Shouldn't this be before the call to `GetWindowTextLengthW`?

Bayemite: Yes, it should be that way...
I was probably taking the documentation too literally, even when it made no sense.

ws909:
Why was the Cocoa implementation removed?

dougbinks:
I have started work on a revised version of this PR, which will retain author credits for @Bayemite.

Currently I have rebased the code and am in progress with switching from the old style platform functions to the new style before testing and reviewing the code on each platform.

PR #1913, opened 2021-06-10
Win32: Expose native HDC with glfwGetWGLContextDC

tksuoran:
OpenXR requires both HDC and HGLRC, so I added getter for HDC.


tksuoran:
I suppose `GetDC(glfwGetWin32Window())` would do so this is probably not needed.

elmindreda:
Yes, `GetDC` will work fine and will return the same DC as the GLFW function would, as our window gets its own private one. That needs to be noted (at least) in the documentation for the Win32 native access functions, though.

elmindreda:
Closing this as fixed by the documentation update showing the `GetDC` solution above.

PR #1923, opened 2021-06-23
Consistent subsections in compile guide

tombsar:
In the compile guide, the "Dependencies for Linux and OSMesa" section looks to be one level too high in the hierarchy. I've moved it to be in line with the similarly-named sub-sub-sections.


elmindreda:
Well spotted, thank you!

PR #1925, opened 2021-06-27
Add glfwGetGLXFBConfig

knokko:
**Description**
This commit adds a new function called `glfwGetGLXFBConfig` that will return the `GLXFBConfig` that was chosen to create a given window.

**Motivation**
This is needed to create an OpenXR session with OpenGL rendering on some Linux systems: It is needed to set the `glxFBConfig` value of the `XrGraphicsBindingOpenGLXlibKHR` struct.

**Review notes**
I am not an experienced C developer and this is the first time I (try to) contribute to GLFW, so I might have missed something important. Also, I don't know a good way to test whether this works correctly; I only checked that it still compiles and that some of the examples still work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ information on what to include when reporting a bug.
  - [EGL] Added ANGLE backend selection via `EGL_ANGLE_platform_angle` extension
    (#1380)
  - [EGL] Bugfix: The `GLFW_DOUBLEBUFFER` context attribute was ignored (#1843)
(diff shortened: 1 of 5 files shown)


elmindreda:
Looks good, will add this. Thanks!

I'm currently restructuring the internals of context creation and this will be even easier to implement after.

sdegrande:
I have the same need...

sdegrande:
Many thx for the merge, @elmindreda

Issue #279, opened 2014-04-20
cannot build glfw as shared and another cmake library as static


onqtam:
The only way (reasonable) to build glfw as a shared lib through cmake is to set BUILD_SHARED_LIBS to ON.
This has the effect of changing every library that does not explicitly specify itself as STATIC or SHARED to be SHARED so it is not suitable when multiple cmake files are in the same "cache domain" of this variable.
PLEASE provide a different way to specify how glfw is built. Maybe add a supplementary option to test against, one that is only for glfw?

if (BUILD_SHARED_LIBS OR GLFW_BUILD_AS_SHARED)
    set(_GLFW_BUILD_DLL 1)
endif()
...
if(GLFW_BUILD_AS_SHARED)
    add_library(glfw SHARED ${glfw_SOURCES} ${glfw_HEADERS})
else()
    add_library(glfw ${glfw_SOURCES} ${glfw_HEADERS})
endif()

onqtam:
still thinking about this..

mcclure:
I have just got burned by this issue... GLFW "namespaces" its other CMake variables so it's strange this one is not "namespaced". This is a real inconvenience with the way CMake works, a CMakeLists include_directorying another CMakeLists has no good way to work around namespace stomping performed by the included CMake file

hdf89shfdfs:
Personally I'd rather GLFW not do this.

When you set BUILD_SHARED_LIBS the expectation is that every library that didn't manually specify it's type will conform to what you said.

Creating custom conventions is never ideal, and age very badly in cmake.
(Speaking as someone who is revamping my company's cmake infrastructure...)

Here is a way you can handle this situation in cmake, without having to modify the cmake code.

```
set(BUILD_SHARED_LIBS_OLD ${BUILD_SHARED_LIBS})

# OFF or ON whatever you want.
set(BUILD_SHARED_LIBS OFF)

add_subdirectory(glfw)

set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_OLD})
```

elmindreda:
Thank you for your feedback and your patience.

You can now (finally) control the library type of GLFW separately with the optional `GLFW_LIBRARY_TYPE` CMake variable added by d83849792b0cf3e40c402f299c867ac6e1fd0e20. When set this will override the value of `BUILD_SHARED_LIBS` without affecting other libraries.

PR #1928, opened 2021-07-01
Rename BUILD_SHARED_LIBS to GLFW_BUILD_SHARED_LIBS

Zino2201:
So lately I was trying to use GLFW in my project combined with other libs.
I needed GLFW to be built as a shared library, but the option `BUILD_SHARED_LIBS` actually sets a CMake flag (https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html).
This flag turns all libraries that don't explicitly specify themselves as `STATIC` or `SHARED` into shared libraries, conflicting with some libraries.

This addresses #279.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,14 +30,14 @@ for:
             - GENERATOR: MinGW Makefiles
     build_script:
         - set PATH=%PATH:C:\Program Files\Git\usr\bin=C:\MinGW\bin%
-        - cmake -S . -B build -G "%GENERATOR%" -DBUILD_SHARED_LIBS=%BUILD_SHARED_LIBS%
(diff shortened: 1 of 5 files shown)


elmindreda:
Thank you for this PR. This feature has been requested several times over the years and should be available in a usable form now.

The (optional) CMake variable `GLFW_LIBRARY_TYPE` added by d83849792b0cf3e40c402f299c867ac6e1fd0e20 will now let you control the library type of just GLFW. When set it will override the value of `BUILD_SHARED_LIBS` for GLFW but won't affect other libraries.

Note that this change does contain a CMake variable named `GLFW_BUILD_SHARED_LIBS`, but this one should not be set by the user. It's just used to avoid shadowing the value of `BUILD_SHARED_LIBS` from a higher-level project.

PR #1931, opened 2021-07-08
Clarify "as you would" for beginners.

### LINKED ISSUE #1088: Make the compiling step more explicit
Hi! 

I rarely use cmake/make on my UNIX distributions, and as such I have limited knowledge to how to use said programs. I am a web dev trying to learn WebGL through OpenGL, and the formulation of the step "Compiling the library" on the page [Compiling GLFW](http://www.glfw.org/docs/latest/compile_guide.html#compile_deps_x11) is a bit obscure. 

I suggest that rather than the current:

```md
## Compiling the library
You should now have all required dependencies and the project files or makefiles necessary to compile GLFW. Go ahead and compile the actual GLFW library with these files, as you would with any other project.

...
```

The following be used:

```md
## Compiling the library
You should now have all required dependencies and the project files or makefiles necessary to compile GLFW. You can now compile the GLFW library with these files. You can generally do so with 

\```
cd <glfw-root-dir>
make install
\```
```

Cheers! 👍

davidvmckay:
fixes #1088

### DIFF
--- a/docs/compile.dox
+++ b/docs/compile.dox
@@ -179,10 +179,12 @@ You should now have all required dependencies and the project files or makefiles
 necessary to compile GLFW.  Go ahead and compile the actual GLFW library with
 these files, as you would with any other project:
 
+e.g. on linux or unix,
 @code{.sh}
 cd <glfw-root-dir>
 make install
 @endcode
+or on Windows, open the .sln file in Visual Studio and build via the file menu.
 
 Once the GLFW library is compiled, you are ready to build your applications,
 linking it to the GLFW library.  See @ref build_guide for more information.
--- a/docs/compile.dox
+++ b/docs/compile.dox
@@ -177,7 +177,12 @@ development environment, move on to @ref compile_compile.
 
 You should now have all required dependencies and the project files or makefiles
 necessary to compile GLFW.  Go ahead and compile the actual GLFW library with
-these files, as you would with any other project.
+these files, as you would with any other project:
+
+@code{.sh}
+cd <glfw-root-dir>
+make install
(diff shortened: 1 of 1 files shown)


aloucks:
I think it would be more helpful to have the `cmake` one-liners.

gracicot:
Usually after doing generating cmake command, you can do `cmake --build .` in the build directory, and it will invoke whatever buildsystem cmake generated.

davidvmckay:
Thank you @aloucks and @gracicot -- those are nice suggestions.  However, that didn't seem to work for me with the current state of `master` in this repository -- SO I'm just trying to improve documentation for users of the repo as its build process stands today.

elmindreda:
@davidvmckay Thank you for this PR. That sentence really did need usage examples. This fix led to me doing a refresh of most of the guide (3b9597048259a198f8f27db7d5eb0eab7de0d3b3). I removed the reference to installation for now. Installation and its pros and cons needs its own section.

If building the main branch with `cmake --build` doesn't work then please report that as a bug in a new issue.

@aloucks @gracicot CMake one-liners are great. In the update above I show both the make and CMake build lines, partly to demonstrate that people don't _need_ to build via CMake, if that makes sense.

This is only an initial refresh and I would love feedback on it.

davidvmckay:
> This fix led to me doing a refresh of most of the guide ([3b95970]
> This is only an initial refresh and I would love feedback on it.

@elmindreda 
Wow, that is a fantastic amount of enhancement to the build guide.  I can't wait to see it published to the main site.  Also loving the updates to the project board.

PR #1951, opened 2021-08-20
fix bit shift runtime error

Slemmie:
#### Stumbled upon an issue while using GLFW in a project:

src/x11_window.c:1274:72: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'

Had a look inside the file. it seems that either the value (1 << 31) is a mere 'inf value', or, perhaps a check to make sure 'diff' itself wont overflow on its own. Proposed change is to simply lower the value a little (by half, for ease). If the value must be exactly (1 << 31), though, perhaps using some other datatype like a long value is better, or just change the condition to (diff <= (1 << 31) - 1) which is achievable without ever casting away from int, maybe use 'INT_MAX'.

I encountered the error after adding glfwCallbacks and then pressing the same keyboard key twice; to be exact, the first instance of pressing some key for the second time during runtime triggered the error, after that the error wouldn't reappear until relaunching the program.

EDIT: I am using Ubuntu 20.04


src/x11_window.c
```diff
@@ -1271,7 +1271,7 @@ static void processEvent(XEvent *event)
                 //       (the server never sends a timestamp of zero)
                 // NOTE: Timestamp difference is compared to handle wrap-around
                 Time diff = event->xkey.time - window->x11.keyPressTimes[keycode];
-                if (diff == event->xkey.time || (diff > 0 && diff < (1 << 31)))
+                if (diff == event->xkey.time || (diff > 0 && diff < (1 << 30)))
```
Hejsil, line 1274: I've hit this issue as well. Idk exactly why this value is used, but if it has to be that way we should probably just use the correct type. (`Time` is either an `unsigned long` or `unsigned int`)
```suggestion
                if (diff == event->xkey.time || (diff > 0 && diff < ((Time)1 << 31)))
```

Slemmie: Good point!

elmindreda:
Sorry for the delay. This has been merged now as b54fb0af10bb0167df67fece984350e66bf569ce. Thank you for the fix!

PR #1952, opened 2021-08-22
Enable joystick support on FreeBSD using the Linux driver.

### LINKED ISSUE #1892: FreeBSD gamepad support
It'd be nice to also have joystick/gamepad support on FreeBSD.

I got a patch for go-gl/glfw here: https://github.com/go-gl/glfw/pull/318/commits/2ab0057315e35fcfe7cd0515951a167ec949dd7a

However it likely can't be used as go-gl imports glfw 3.3. So would it be possible to get the changes from there into glfw 3.3-stable and also master, and then we could make go-gl import the upstreamed changes?

divVerent:
This is arguably incomplete; it supports joysticks via webcamd (which uses Linux
drivers running in userspace), but not via uhidd (which is FreeBSD's own thing).

Nevertheless a strict improvement.

Tested using a gamepad I have that is supported by the "xpad" Linux driver.

Note: adds build dependency on devel/evdev-proto and devel/libinotify.
To actually use it, need to install multimedia/webcamd at runtime (if not
installed or not running, no joysticks will be detected).

Fixes #1892.


divVerent:
Note: I have not tested this; my game is written in Go, which is based on the 3.3 release series of glfw, and this change is a port-forward of my 3.3 changes I actually did test. Joystick APIs however have changed significantly from 3.3 to master.

Nevertheless, the changes are trivial except for the added build dependency.

Will check later if I can make a local go-gl/glfw build based on the current master branch.

divVerent:
Now tested the changes properly: copying the glfw master source into go-gl/glfw/v3.3/glfw/glfw works just fine.

FrostKiwi:
Works for me for both a DualShock 3 and a 360 Controller. Big <3 <3

Curious how this will interact with bluetooth controllers. I am intersted in getting Bluetooth controllers to work, namely the DualShock 3, 4 and Dualsense. (PS3, PS4, PS5) Fun investigations ahead...

divVerent:
Sadly I can already tell you that webcamd does not support Bluetooth
devices. I do not have any myself, but it seems like they're supported
by ports that are neither webcamd nor uhidd, but implement the uhidd
protocol.

So eventually GLFW will also need the uhidd support; my PR only covers
the webcamd protocol at the moment, primarily because it's a trivial
change given the existing Linux implementation already works for it
when actually enabled.

PR #1972, opened 2021-10-06
Properly discard filtered events by XFilterEvent()

### LINKED ISSUE #1794: Key callback and char callback are called on different polls on some Linux machines
To reproduce, run the following code:

```c
#include <assert.h>
#include <stdio.h>
#include <stdlib.h>

#include <GLFW/glfw3.h>

int polln = 0;

void keyCall(GLFWwindow *win, int key, int scancode, int action, int mods) {
    printf("[%d] key: key=%d, sc=%d, act=%d, mod=%d\n", polln, key, scancode, action, mods);
}

void charCall(GLFWwindow *win, unsigned int codepoint) {
    printf("[%d] char: code=%u\n", polln, codepoint);
}

int main(int argc, char **argv) {
    assert(glfwInit());
    GLFWwindow *win = glfwCreateWindow(640, 480, "key test", NULL, NULL);
    assert(win);

    glfwSetKeyCallback(win, keyCall);
    glfwSetCharModsCallback(win, charCall);

    glfwMakeContextCurrent(win);

    while (!glfwWindowShouldClose(win)) {
        glfwSwapBuffers(win);
        glfwPollEvents();
        polln++;
    }

    glfwTerminate();
    return 0;
}
```

The code polls events and displays output in the key callback and char callback. The number in the brackets is the poll number (i.e. n-th poll)

## Actual behaviour

On some Linux machines, the key callback is called on one poll, and the char callback is called on the next poll:

```
[156] key: key=65, sc=38, act=1, mod=0
[157] char: code=97
[163] key: key=65, sc=38, act=0, mod=0
```

This behaviour seems to happen on GNOME but not KDE. On an affected machine, this happens consistently. This is what causes the Minecraft bug [MC-122477](https://bugs.mojang.com/browse/MC-122477).

My machine is affected: Pop!_OS 20.10. See the above Minecraft bug report for other users who have reported experiencing this issue.

## Expected behaviour

The key callback and char callback are called during the same poll:

```
[13659] key: key=65, sc=38, act=1, mod=0
[13659] char: code=97
[14837] key: key=65, sc=38, act=0, mod=0
```

Indeed, this is the behaviour on Linux machines running KDE (and probably on Windows and macOS too, but I haven't tested).

RecursiveG:
According to https://www.x.org/releases/current/doc/man/man3/XFilterEvent.3.xhtml,
an event should be discarded if XFilterEvent returns `True`. Fix #1794.

### DIFF
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -1163,6 +1163,10 @@ static void processEvent(XEvent *event)
         keycode = event->xkey.keycode;
 
     filtered = XFilterEvent(event, None);
+    if (filtered)
+    {
+        return;
+    }
 
     if (_glfw.x11.randr.available)
     {


daipom:
This fix is necessary for input-method to work properly on X11. (#2130)

Since I thought I would need to make this kind of fix in #2130, I checked the behavior regarding this.
This fix only changes the timing of calling `_glfwInputKey`.

A simple key-press of `a` key consists of 6 events as follows.
(I added some debug-prints to `x11_window.c` and used `tests/events.c` to confirm the behavior)

Without this fix:

```
-- 1st event --
XFilterEvent: TRUE
Event-type: KeyPress
00000006 to 1 at 2.546: Key 0x0041 Scancode 0x0026 (A) (a) (with no mods) was pressed
-- 2st event --
XFilterEvent: TRUE
Skipped since XFindContext == 0
-- 3st event --
XFilterEvent: FALSE
Event-type: KeyPress
00000007 to 1 at 2.549: Character 0x00000061 (a) input
-- 4st event --
XFilterEvent: TRUE
Event-type: KeyRelease
00000008 to 1 at 2.624: Key 0x0041 Scancode 0x0026 (A) (a) (with no mods) was released
-- 5st event --
XFilterEvent: TRUE
Skipped since XFindContext == 0
-- 6st event --
XFilterEvent: FALSE
Event-type: KeyRelease
```

With this fix:

```
-- 1st event --
XFilterEvent: TRUE
Skipped by this fix
-- 2st event --
XFilterEvent: TRUE
Skipped by this fix
-- 3st event --
XFilterEvent: FALSE
Event-type: KeyPress
00000020 to 1 at 26.553: Key 0x0041 Scancode 0x0026 (A) (a) (with no mods) was pressed
00000021 to 1 at 26.553: Character 0x00000061 (a) input
-- 4st event --
XFilterEvent: TRUE
Skipped by this fix
-- 5st event --
XFilterEvent: TRUE
Skipped by this fix
-- 6st event --
XFilterEvent: FALSE
00000022 to 1 at 26.651: Key 0x0041 Scancode 0x0026 (A) (a) (with no mods) was released
```

Thus, this fix ensures that `_glfwInputKey` and `_glfwInputChar` occur simultaneously in a single `glfwPollEvents` or `glfwWaitEvents`, as mentioned in #1794.

This fix does not affect the repeat logic.

While it may affect several events other than KeyPress/Release, we should not process the filtered events without a special reason. (https://www.x.org/releases/current/doc/libX11/libX11/libX11.html#Event_Filtering_b)

daipom:
We can remove this if.

https://github.com/glfw/glfw/blob/679106e8b027b74f8dc1d57c38bafc7c0794c3c7/src/x11_window.c#L1286-L1318

We can remove this too.

https://github.com/glfw/glfw/blob/679106e8b027b74f8dc1d57c38bafc7c0794c3c7/src/x11_window.c#L1559-L1560

j-romchain:
why hasn't this been merged yet???

j-romchain:
it says there are no conflicts and its just been sitting here for YEARS without merging, why?

lixiangwuxian:
Is glfw not actively maintained anymore?
Seems fyne in x11 need this patch to make ime works functionally.
@elmindreda may a review please

ashie:
@elmindreda

Would you consider adding this PR to the [3.6 milestone](https://github.com/glfw/glfw/milestone/26)?

In https://github.com/glfw/glfw/pull/2130#issuecomment-1922075877, you mentioned that the next minor release would focus on keyboard input improvements (which now appears to have shifted to 3.6). Given that, I think this PR is also relevant to the goals of that release.

Additionally, this change improves the behavior of #2130, which is already part of the 3.6 milestone, while still being a worthwhile fix on its own.

For those reasons, I think this PR may be worth considering for inclusion in the 3.6 milestone.

PR #1982, opened 2021-10-28
Fix source comment typo

luzpaz:


elmindreda:
Thank you for the fix!

PR #1989, opened 2021-11-11
Fix undefined behaviour in X11 keypress handling

silversquirl:
`1 << 31` is undefined because 1 is an int, so the shift would set the sign bit

Related to (and discovered in the same way as) #1986


elmindreda:
Thank you for the PR! This should finally be fixed now with b54fb0af10bb0167df67fece984350e66bf569ce from #1951.

PR #1992, opened 2021-11-15
update docstring which inaccurately says GLFW_CONTEXT_REVISION is a window hint

slimsag:
This docstring previously indicated that `GLFW_CONTEXT_REVISION` was a window
hint and attribute, but in fact it is only a window attribute (there is no code
which uses this constant in any other context.)

We noticed this in https://github.com/hexops/mach/pull/71/files#r749741814

Signed-off-by: Stephen Gutekanst <stephen@hexops.com>


elmindreda (approved):
👍🏻

elmindreda:
Thank you for the fix!

Issue #1985, opened 2021-10-31
macOS < 12 not supported anymore?


vincent1114:
![image](https://user-images.githubusercontent.com/50698701/139578692-4a14310d-cd0c-478b-ba0d-068f6e59489b.png)

On glfwinit(), segfaults on this line.
I don't see losing support for macOS 12 on release note, so I assume this was probably missed.

saschasc:
Something like this could probably work.

```
#if (MAC_OS_X_VERSION_MAX_ALLOWED < 120000) // Before macOS 12 Monterey
  #define kIOMainPortDefault kIOMasterPortDefault
#endif
```

I had to add the following definition to CMakeLists.txt to make it compile after upgrading to XCode 13.1 (using llvm@12).

`add_definitions("\"-DNS_FORMAT_ARGUMENT\(A)=\"")`

jeeb:
@vincent1114 can you check if #1994 helps? In the end it seems like `MACH_PORT_NULL` is the same as the value of this extern symbol (looking at the IOKit header comments, as well as https://developer.apple.com/documentation/iokit/kiomasterportdefault?language=objc ).

Kind of hilarious that this is an extern symbol and not just an enum or define, so instead of just checking whether your SDK is new enough, you also have to take into account target if you want to utilize the new version of this symbol :)

saschasc:
@jeeb For me it works. But I still have to add

`add_definitions("\"-DNS_FORMAT_ARGUMENT\(A)=\"")`

when using LLVM 12.

jeeb:
With Xcode 13.1 on macOS 11.6 (and thus targeting 11.6 by default) I can just do the following (at least with the current master)
```
rm -rf macos_build/ && mkdir -p macos_build/ && cd macos_build/
cmake -G "Ninja"  ..
ninja
```
which outputs the following
```
-- The C compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
[...]

PR #1994, opened 2021-11-18
cocoa: fixup move to non-deprecated default IOKit mach port

jeeb:
f75c251deccde745d3d3f313aa81955095369d6f was a valiant attempt, but unfortunately  [`kIOMasterPortDefault`](https://developer.apple.com/documentation/iokit/kiomasterportdefault?language=objc) and [`kIOMainPortDefault`](https://developer.apple.com/documentation/iokit/kiomainportdefault?language=objc) are not enum values, but instead external `mach_port_t` symbols.

Thus just checking if the SDK is new enough to support this symbol's definition is not enough, and will cause compilation warnings with f.ex. SDK from Xcode 13.1 and the target being macOS <12.0. This is due to the new symbol being utilized unconditionally without `@available` runtime conditions (`unguarded-availability-new` warnings, which I wonder why they're not just errors by default). And during runtime it doesn't find the symbol and crashes :)

This PR contains the "three stages of grief" I went through when figuring out how to best fix this, and thus I am posting them so that one of them could be picked for upstream inclusion.

1. Just switch the #define check to target version instead of SDK version.
   
   **Pros**: compiles without warnings for both targets.
   **Cons**: if the old external symbol is removed, a build built against an older target doesn't work on a newer system any more.
2. Properly utilize `@available` for runtime checking and utilization of new or old symbol depending on its availability.
   
   **Pros**: compiles without warnings for both targets, utilizes new symbol for new OS versions (although I think I am missing another elif for target > 12.0 where only the new symbol is utilized).
   **Cons**: requires both ifdefs and a function.
3. Notice a comment in the definition of the symbol that:

   > When specifying a primary port to IOKit functions, the NULL argument indicates "use the default". This is a synonym for NULL, if you'd rather use a named constant.

   ... and thus just utilizing MACH_PORT_NULL should be good enough?
   
   **Pros**: simplest if it works.
   **Cons**: I am not sure if I got proper values out of my macOS 11.6 system with either `kIOMasterPortDefault` or `MACH_PORT_NULL` within libplacebo's plplay that I was testing things against.

I'm pretty sure versions 1 and 2 should work as expected (I will have to adjust the ifdeffery if that ends up being the preferred route), for version 3 I would like someone who does get values out of these APIs to verify that `MACH_PORT_NULL` works just like documentation hints that it does.

### DIFF
--- a/src/cocoa_monitor.m
+++ b/src/cocoa_monitor.m
@@ -58,7 +58,7 @@ static char* getMonitorName(CGDirectDisplayID displayID, NSScreen* screen)
     io_service_t service;
     CFDictionaryRef info;
 
-    if (IOServiceGetMatchingServices(kIOMainPortDefault,
+    if (IOServiceGetMatchingServices(MACH_PORT_NULL,
                                      IOServiceMatching("IODisplayConnect"),
                                      &it) != 0)
     {
@@ -231,7 +231,7 @@ static double getFallbackRefreshRate(CGDirectDisplayID displayID)
     io_iterator_t it;
     io_service_t service;
 
-    if (IOServiceGetMatchingServices(kIOMainPortDefault,
+    if (IOServiceGetMatchingServices(MACH_PORT_NULL,
                                      IOServiceMatching("IOFramebuffer"),
                                      &it) != 0)
     {
--- a/src/cocoa_platform.h
+++ b/src/cocoa_platform.h
@@ -44,13 +44,6 @@ typedef void* id;
 // NOTE: Many Cocoa enum values have been renamed and we need to build across
 //       SDK versions where one is unavailable or the other deprecated
 //       We use the newer names in code and these macros to handle compatibility
-#if MAC_OS_X_VERSION_MAX_ALLOWED < 120000
- #define kIOMainPortDefault kIOMasterPortDefault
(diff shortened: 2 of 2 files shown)


jeeb:
Alright, did a short verification and it seems like the note in the documentation holds true, and `MACH_PORT_NULL` can be utilized instead of this extern symbol.

Code:
```c
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <IOKit/IOKitLib.h>

int main(int argc, char const *argv[])
{
	const mach_port_t null_port = MACH_PORT_NULL;

	fprintf(stderr, "MACH_PORT_NULL: %u, kIOMasterPortDefault: %u\n", null_port, kIOMasterPortDefault);
	return 0;
}
```

Compilation
```
cc -o iokit_test iokit_test.c  -framework IOKit
```

Result:
```
./iokit_test
MACH_PORT_NULL: 0, kIOMasterPortDefault: 0
```

I will simplify this PR then to only have the last commit :) .

jeeb:
Alright, simplified and added a reference to #1985 .

elmindreda (approved):
Looks good!

elmindreda:
Thank you!

PR #2018, opened 2021-12-29
Windows: avoid modifying the global keyboard state with ToUnicode

lb90:
Since Windows 10 version 1607 we can make use of a new flag that
makes ToUnicode operate on a process-private keyboard state.

Reference:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-tounicode

Note: this is a coordinated effort across various open source projects

* [GTK](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4195)
* [Chromium](https://chromium-review.googlesource.com/c/chromium/src/+/3360877)

### DIFF
--- a/src/win32_init.c
+++ b/src/win32_init.c
@@ -487,13 +487,13 @@ void _glfwUpdateKeyNamesWin32(void)
 
         length = ToUnicode(vk, scancode, state,
                            chars, sizeof(chars) / sizeof(WCHAR),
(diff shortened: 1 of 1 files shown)


elmindreda:
This should be fixed now with ad01c1b614868c3cbc79306aa6a19c9fc06f34a6. Thank you!

PR #2024, opened 2022-01-04
X11: Fix segfault when _glfw.x11.display is higher than 1024

illustris:
FD_SETSIZE is hardcoded to 1024 in glibc. See the bugs section of the select call for more info.
In some cases, it is possible for _glfw.x11.display to be larger than FD_SETSIZE, causing a segfault or buffer overflow.
Replacing select() with poll() seems to be the recommended solution:
- https://www.man7.org/linux/man-pages/man2/select.2.html#BUGS
- https://access.redhat.com/solutions/488623
- https://stackoverflow.com/questions/7976388/increasing-limit-of-fd-setsize-and-select/7977082#7977082

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,7 @@ information on what to include when reporting a bug.
  - [X11] Bugfix: Icon pixel format conversion worked only by accident, relying on
    undefined behavior (#1986)
  - [X11] Bugfix: Dynamic loading on OpenBSD failed due to soname differences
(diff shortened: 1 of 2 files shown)


illustris:
Whoops... didn't notice the above commit wasn't on glfw/glfw. Reopening to track that commit getting merged here.

elmindreda:
Sorry for the confusion. Forgot that GH notifies of issue tags in forks. Thank you for the PR! This should be fixed now with d3e4fcf8b7608e7b6f6cf1c102b1e28c478f5a6c.

PR #2033, opened 2022-01-25
Fix #1281: glfwPostEmptyEvent sometimes doesn't wake up glfwWaitEvents on X11

### LINKED ISSUE #1281: glfwPostEmptyEvent sometimes fails to unblock glfwWaitEvents on Ubuntu 18.04

joaodasilva:
This is a rare race that triggers often when running GLFW programs on GitHub Actions with Xvfb.

The underlying issue is internal to Xlib: XSendEvent races with an external select() call and may get dispatched before select() has blocked on the display fd.

The fix observes a pipe in the select() call, and writes to that pipe from glfwPostEmptyEvent to make select() return.

The original issue reproduces locally once per minute when running the windowjs/windowjs tests in a loop;
The bug hasn't triggered after running for several hours with the fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+ - Bugfix: fixed glfwPostEmptyEvent not unblocking glfwWaitEvents on X11 (#1281)
  - Added `GLFW_PLATFORM` init hint for runtime platform selection (#1958)
  - Added `GLFW_ANY_PLATFORM`, `GLFW_PLATFORM_WIN32`, `GLFW_PLATFORM_COCOA`,
    `GLFW_PLATFORM_WAYLAND`, `GLFW_PLATFORM_X11` and `GLFW_PLATFORM_NULL` symbols to
--- a/src/x11_init.c
+++ b/src/x11_init.c
@@ -35,6 +35,7 @@
 #include <stdio.h>
 #include <locale.h>
 #include <unistd.h>
+#include <fcntl.h>
 
 
 // Translate the X11 KeySyms for a key to a GLFW key code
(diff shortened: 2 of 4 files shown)


joaodasilva:
Hi GLFW maintainers. This is my first PR on GitHub; please advise if this isn't the right way to contribute fixes.

I'm using GLFW for windowjs.org and spent quite some time chasing this down, mainly due to flakes in the GitHub Actions test runners due to the issue.

The discussion in #1281 already mentioned using a self-pipe to fix the issue but activity seems stale; I've just updated and polished the patch, and addressed just the issue at hand (i.e. this doesn't replace select() with poll() to reduce risk).

This is applied manually to windowjs/windowjs now, would be great to have it fixed upstream :-)

elmindreda:
Thank you for the PR! This should be fixed now with cd22e2849512a88d0ab77bc7a3458646625f2c50, which was partly based on it.

PR #2035, opened 2022-01-30
Platform: Prefer Wayland over X11

cbix:
When GLFW_ANY_PLATFORM is used (which is the default) native
Wayland would previously never be chosen if XWayland is also available.

Another idea (TBD) would be to control this behavior by an environment variable (see Qt or Firefox) and otherwise default to X11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ information on what to include when reporting a bug.
  - [Wayland] Added dynamic loading of all Wayland libraries
  - [Wayland] Added support for key names via xkbcommon
  - [Wayland] Removed support for `wl_shell` (#1443)
+ - [Wayland] Prefer Wayland over X11 if both are available (#2035)
  - [Wayland] Bugfix: The `GLFW_HAND_CURSOR` shape used the wrong image (#1432)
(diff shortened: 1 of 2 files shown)


elmindreda:
> When GLFW_ANY_PLATFORM is used (which is the default) native Wayland would previously never be chosen if XWayland is also available.

This is on purpose, as the X11 backend is still much more complete than the Wayland one. Working on bringing the Wayland one up to speed now. The change in this PR will be made once the Wayland backend is sufficiently complete. (This is why Wayland is listed before X11 everywhere in the runtime platform selection other than in this array).

Thank you for the PR! Hope to be able to merge it soon.

> Another idea (TBD) would be to control this behavior by an environment variable (see Qt or Firefox) and otherwise default to X11.

Yeah, a `GLFW_PLATFORM` environment variable may be a good addition. Note that you can select Wayland already by un-setting the `DISPLAY` environment variable.
```sh
env -u DISPLAY ./my/app
```

cbix:
Thanks, looking forward to 3.4! :)

PR #2040, opened 2022-02-04
Wayland Drag and Drop

Pilzschaf:
Implements Drag and Drop for the Wayland backend.
This actually also fixes a bug in the previous implementation for the clipboard.
As the clipboard and drag and drop both use the wl_data_device, the dataDeviceHandleDataOffer callback is called for new clipboard and new drag and drop requests. So previously while an item has been hovered over the window the glfwGetClipboardString function would return the current drag and drop payload. The new version uses dataDeviceHandleSelection and dataDeviceHandleEnter to distinguish between clipboard and drag-and-drop data respectively.
Similar to the other backends, only the text/uri-list mimetype is handled. The uri-list is parsed with the parseUriList function which has been copied from the X11 backend. This function could be moved to a general utility file to remove this duplication.

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -773,7 +773,7 @@ static void dataDeviceHandleDataOffer(void* data,
                                       struct wl_data_device* dataDevice,
                                       struct wl_data_offer* id)
 {
-    wl_data_offer_add_listener(_glfw.wl.dataOffer, &dataOfferListener, NULL);
+    wl_data_offer_add_listener(id, &dataOfferListener, NULL);
 }
 
 static void dataDeviceHandleEnter(void* data,
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -26,6 +26,8 @@
 // It is fine to use C99 in this file because it will not be built with VS
 //========================================================================
 
+#define _GNU_SOURCE
+
 #include "internal.h"
 
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


src/wl_init.c
```diff
@@ -752,12 +816,91 @@ static void dataDeviceHandleMotion(void* data,
...
 {
+    int ret;
+    int fds[2];
+    size_t len = 0;
+    size_t size = 2048;
+    char* buffer = 0;
+
+    if(_glfw.wl.dragDataOffer)
+    {
+        ret = pipe2(fds, O_CLOEXEC);
```
ronchaine, line 827: I'm not really one to approve or reject PR's, but I wanted to comment:

Does this really need to be `O_CLOEXEC` (are we execing into anywhere?)

Without it the `_GNU_SOURCE` would not be required and this would be more
standards-compliant.

Pilzschaf: This code is based on the reading of the pipe for clipboard data in _glfwGetClipboardStringWayland in wl_window.c. There the O_CLOEXEC flag is also specified. So as this is already a requirement (at least for the wayland backend) I think it is best to also include the flag here.

ronchaine: Hmm, you are right. It also seems that both FreeBSD and OpenBSD seem to implement `pipe2` with `O_CLOEXEC` too so it shouldn't really be a future issue either.

elmindreda:
Thank you! There was a lot to address in the clipboard and drop paths so the final commit ended up a bit different but this PR was very helpful!

PR #2045, opened 2022-02-13
EGL: Add glfwGetEGLConfig

knokko:
This merge request adds the function `glfwGetEGLConfig`. This function is useful for creating an OpenGL rendering session in OpenXR.

The LWJGL project maintains its own GLFW fork that already has this commit. This works fine, but it could be useful for other people in the future if this merge request is accepted and this function is added to 'core GLFW'.

### DIFF
--- a/include/GLFW/glfw3native.h
+++ b/include/GLFW/glfw3native.h
@@ -517,6 +517,24 @@ GLFWAPI EGLContext glfwGetEGLContext(GLFWwindow* window);
  *  @ingroup native
  */
 GLFWAPI EGLSurface glfwGetEGLSurface(GLFWwindow* window);
+
+
+/*! @brief Returns the `EGLConfig` of the specified window.
+ *
+ *  @return The `EGLConfig` of the specified window, or `EGL_NO_SURFACE` if an
+ *  [error](@ref error_handling) occurred.
+ *
+ *  @errors Possible errors include @ref GLFW_NO_WINDOW_CONTEXT and @ref
+ *  GLFW_NOT_INITIALIZED.
+ *
+ *  @thread_safety This function may be called from any thread.  Access is not
+ *  synchronized.
+ *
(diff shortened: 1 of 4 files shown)


include/GLFW/glfw3native.h
```diff
@@ -517,6 +517,24 @@ GLFWAPI EGLContext glfwGetEGLContext(GLFWwindow* window);
...
  */
 GLFWAPI EGLSurface glfwGetEGLSurface(GLFWwindow* window);
+
+
+/*! @brief Returns the `EGLConfig` of the specified window.
+ *
+ *  @return The `EGLConfig` of the specified window, or `EGL_NO_SURFACE` if an
+ *  [error](@ref error_handling) occurred.
+ *
+ *  @errors Possible errors include @ref GLFW_NO_WINDOW_CONTEXT and @ref
```
Frostie314159, line 527: If glfw wasn't initialized yet this could also return `EGL_NO_SURFACE`

knokko: If GLFW isn't initialized yet (which I would consider as an error), this would indeed return `EGL_NO_SURFACE`. This is consistent with the behavior of `glfwGetEGLContext` and `glfwGetEGLSurface`.

theoparis:
Any updates on this? I am trying to make a minecraft mod that forces minecraft to use EGL instead of desktop opengl, and I don't want to have to rely on a 2 year old fork of glfw.

knokko:
I think you should also be able to use this fork: https://github.com/LWJGL-CI/glfw which is not running behind GLFW master
It looks like it still includes this PR: https://github.com/LWJGL-CI/glfw/commit/d163f8c7e0de2d8d895883feeaaafa05f82f29ec#diff-21938cdabc3c2b2f9924afd55e59cb74fb7da35f896e97f9c99783db9088f8a4R966

PR #2058, opened 2022-03-17
rgrgrgrg

LesterDigital:


gracicot:
I wonder why there is so many of those weird pull requests always adding premake files?

elmindreda:
@LesterDigital Thank you for this PR. I won't merge additional build setups unless they also come with an active maintainer, but I am working on making it easier to build GLFW without the official CMake files.

Issue #2036, opened 2022-01-30
Check if GLFW has been initialized


mlivesu:
Hi,
I maintain a library that exposes rendering facilities internally implemented with GLFW. I'd like to handle the calls to `glfwInit()` and `glfwTerminate()` directly from within the library, but since there are multiple entry points it would be necessary to know if the environment had already been initialized or not. 

I've seen that internally these macros implement similar checks
`_GLFW_REQUIRE_INIT()`
`_GLFW_REQUIRE_INIT_OR_RETURN(NULL)`
but they are not accessible from outside GLFW and I could not find an alternative public call in the API.  

A similar question was posted long time ago [here](https://discourse.glfw.org/t/check-if-glfw-has-been-initialized/367), but I am not aware of any improvement to this end. Do you think it would be possible to add a boolean function like `glfwIsInitialized()`? It is true that one could obtain the desired effect by externally storing the GLFW state in an external boolean variable, but in my specific case I'd rather avoid it because the library does not maintain an internal state.

Thanks!

ws909:
The two first lines of `glfwInit` are these, so there's no need for it.
```
if (_glfw.initialized)
        return GLFW_TRUE;

```

mlivesu:
I know I've been there, but this does not solve my problem. `glfwInit` will return `GLFW_TRUE ` regardless of whether the initialization happened at the current call or at some previous call, therefore from the outside it becomes impossible to distinguish between the two cases.

In my library I have multiple rendering facilities based on GLFW. In this case it is sometimes convenient to perform a test like `if (_glfw.initialized)`, to avoid terminating GLFW meanwhile somebody else is actually using it.

ws909:
If the client is not using GLFW, will your library initialize and terminate GLFW several times during the execution of the program? So each of your rendering methods will begin with `glfwInit`, and end with `glfwTerminate`? That sounds like a really bad solution, performance-wise.

mlivesu:
Yes and no. I have a GL window that is used for online rendering and in 99.9% of the cases is the function that will initialize and terminate GLFW.  In addition to this I have an offline rendering facility, based on `glfwWindowHint(GLFW_VISIBLE, GLFW_FALSE)`, that has a very narrow applicability and could be used once in a while either by a windowed application (i.e. meanwhile there is a GL window alive) or in a command line program. For this latter one, I'd need to know whether I should initialize and terminate GLFW or not, and I don't want to have an internal state that tells me so because the library is stateless.

[...]

PR #2065, opened 2022-03-30
Added glfwIsInitialized function

Frostie314159:
This adds a `glfwIsInitialized` function, that just accesses the internal `initialized` paramter of the `GLFWlibrary` object.

This makes it more direct and convenient to check if glfw is already initialized.

As requested in [glfw#2036](https://github.com/glfw/glfw/issues/2036)

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -2364,6 +2364,22 @@ GLFWAPI const char* glfwGetVersionString(void);
  */
 GLFWAPI int glfwGetError(const char** description);
 
+/*! @brief Returns the initialization state.
+*   This function returns whether the library has been initialized or not.
+*   @return `GLFW_TRUE` if initialized, or `GLFW_FALSE` otherwise.
+*
+*   @errors None.
+*   
+*   @remark This function may be called before @ref glfwInit.
+*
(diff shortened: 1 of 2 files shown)


wintertime:
There is more than one way to help users with the underlying problem:
1. I'd guess, having this function would be the simplest addition to code around mutliple libraries in one application not knowing if they still need to initialize GLFW. It would just be a higher burden on users to always check.
2. If `glfwInit` would just error out with something like `GLFW_ERROR_ALREADY_INITIALIZED` (instead of silently ignoring the call) that would solve the problem without having to add a new function. Users would then just check this and avoid calling `glfwTerminate` if some other part already initialized GLFW. I'm just not sure if that meets compatibility requirements. User code would be about as complicated as 1.
3. The most convenient way for users without making the implementation much more complicated than now would be to make `initialized` a counter. On `glfwInit` check the counter then only call the initialization code if it was 0, but always increase it by 1. On `glfwTerminate` error (or maybe ignore) if the counter is 0, decrease it by 1 and call termination code when it reaches 0 after it was decresed. That would be compatible with current user code and enable very easy use (just call `glfwInit` and `glfwTerminate` irregardless of other users).

PR #2066, opened 2022-03-31
Added a glfwGetVulkanResultString

Frostie314159:
This adds a public `glfwGetVulkanResultString` function, improves the already existing internal `_glfwGetVulkanResultString` function by adding all official error code descriptions and updating the internal [`VkResult`](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkResult.html#_description) enum to the Vulkan 1.3 definition
The internal function is just exposed by this patch, but preserved to not break the internal API.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -6212,6 +6212,23 @@ GLFWAPI const char** glfwGetRequiredInstanceExtensions(uint32_t* count);
 
 #if defined(VK_VERSION_1_0)
 
+/*! @brief Returns the result as a string.
+ *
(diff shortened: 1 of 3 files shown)


elmindreda:
Thank you for this PR but I will not merge this. As I see it utility function like this should not be in a library a slow-moving as GLFW currently is and would be better as a stand-alone library, possibly in header-only form. That way it could be useful to people that create their windows in other ways as well. GLFW takes care of (most? some?) platform-specific aspects of using Vulkan but platform-independent things like this do not need to be tied to it.

PR #2068, opened 2022-04-01
EGL: add context priority for the EGL context creation API

gregorburger:
Allows to specify context priority low, medium, high or realtime if
the extension is available. This is currently only implemented for
the EGL context creation API.

Currently only implemented for EGL, but equivalent extensions 
exist for GLX and Vulkan.


gregorburger:
I use it with nvidia drivers on desktop linux. I saw some commits that intel mesa drivers are supporting it as well. Could not figure out if amd drivers are supporting it.

gregorburger:
and as mentioned this could be extended to be used by vulkan context creation as well. 
https://www.khronos.org/blog/vulkan-1.3-and-roadmap-2022 
[VK_KHR_global_priority](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_KHR_global_priority.html)

ws909:
VK_KHR_global_priority and other interactions with Vulkan are outside the scope of GLFW. This is handled by client code.

PR #2069, opened 2022-04-01
Convert defines to enums

Frostie314159:
This PR attempts to migrate most defines to enums.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -539,7 +539,7 @@ enum GLFWkey{
  *
  *  @ingroup input
  *  @{ */
-enum GLFWkey_modifier{
+enum GLFWmodkey{
 /*! @brief If this bit is set one or more Shift keys were held down.
  *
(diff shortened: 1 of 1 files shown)


Frostie314159:
I just noticed the build errors. They come from a change I made aiming to make function attributes and return values more logical(i.e. uint32_t instead of int in places where it made sense). I'll make a separate PR with those changes sometime soon.

elmindreda:
Thank you for this PR but I will not merge this. The GLFW 3 API needs to remain stable, including the macros the public header provides. The next major version will likely use enums for numerical constants, which will require renaming symbols and changing function parameter types.

Frostie314159:
But with enums nothing should change since they refer to the same value. So this would be non-breaking.

Frostie314159:
> The GLFW 3 API needs to remain stable, including the macros the public header provides. 
I see what you mean, but I tested it with the provided tests and it worked without issues.
This change shouldn't even be noticed by most users.

PR #2070, opened 2022-04-02
Improve init and termination with counting

wintertime:
Counts additional calls to glfwInit and waits with termination until the same
amount of calls to glfwTerminate is done.
For normal users nothing should change.
For libraries/plug-ins they can now call glfwInit and glfwTerminate additional
times as long as the calls are balanced. This avoids the problem that they
would terminate GLFW completely when the main application is still using it.

This should fix the underlying problem of #2036 . It may still be useful to additionally merge #2065 .


nitrix:
Major buy-in on my side for this.

I have worked with a lot of bindings in Go and Rust that have been doing this sort of reference counting manually, even my own C code, just because the documentation for glfwInit() was underspecified.

It's going to make a lot of mistakes go away for people using the library the first time. Especially when GLFW gets wrapped in an abstraction and the application creates multiple windows.

Thank you @wintertime, hope we can get this approved.

nitrix:
Had a small discussion on IRC, logging here for the record.

Essentially, the backwards compatibility of GLFW is quite important and the current state of the PR would affect the behavior for developers that are relying on the multiple init (with the superfluous ones ignored) + single termination.

I proposed a new glfwInitHint(GLFW_REFCOUNTED, true); initialization flag that can be set prior to initialization, after which the rest of the PR can apply -- only conditionally to that flag being set though.

It seems reasonable and I can spend a bit of time on that approach depending on @wintertime's reponsiveness.

CC @elmindreda to confirm the viability of the idea.

wintertime:
Hi,

I saw the first, happy reply in email, but did not get a notification for the edit. It was pure luck I did not find it a month later.

I disagree, adding a hint to conditionally keep the bug (or strange behaviour oversight) would be a feasible approch:
- Currently, it would automatically fix the problem with an update, but with the conditional hint it would not. 
- For the majority use case of having a single init and single terminate nothing changes regarding compatibility.
- For the case of an application using init+terminate on demand repeatedly nothing changes, too.
- For the case of an application having an unknown number of plug-ins (or similar) and both the application and plug-ins needing GLFW, the old code only supports having init done by the application always and the plug-ins relying on it, but avoiding these calls. That would be possible already.
- If the application would not reliably call init, it would IMHO be a plug-in bug to haphazardly call init itself, because it can not know if it would later be the last one having to call terminate (unless using other unsupported hacks, because until the other MR is merged there is no function to test if initialization was done), therefore it would just never call terminate (as it cause very likely break another plug-in) and leak (nicer told "rely on the OS for cleanup"). If it never calls terminate anyway (or just the application before exit) the change would not influence compatibility.
- Lastly, I doubt the old code allows to coordinate with the app+multiple plug-ins to externally wrap around a scheme to reliably call terminate always but not too early. So I'd guess, there can be no compatibility problem, too. Or you have examples for this case actually? If yes, I'd think a cmake option would be the way to optionally reactivate the bug.

nitrix:
I prefer the PR as-is, but that'll be a tough sell to @elmindreda.
Either way, I'm offering my help too if there's a clear path forward.

PR #2103, opened 2022-05-06
Fix build error message with removed OSMesa option

linkmauve:
The correct hint wasn’t `GLFW_PLATFORM` but `GLFW_OSMESA_CONTEXT_API`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ include(GNUInstallDirs)
 include(CMakeDependentOption)
 
 if (GLFW_USE_OSMESA)
-    message(FATAL_ERROR "GLFW_USE_OSMESA has been removed; set the GLFW_PLATFORM init hint")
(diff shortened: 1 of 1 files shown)


elmindreda:
The main purpose of the `GLFW_USE_OSMESA` CMake option in 3.3 is to select the Null backend, and only briefly during the development of 3.3 was it to select OSMesa as context creation API.

Story time!

The Null backend was originally called OSMesa, because it was created by someone at Google for the sole purpose of being a testing stub with OSMesa for context creation. The CMake option for enabling this was `GLFW_USE_OSMESA` and it kept that name for compatibility reasons when the backend was renamed to Null. `GLFW_OSMESA_CONTEXT_API` was added later in 3.3, enabling runtime support for OSMesa for other backends. I would have added a better synonym for `GLFW_USE_OSMESA` in what will be 3.4 but then runtime platform selection allowed me to remove it entirely instead.

This why `GLFW_USE_OSMESA` points people at `GLFW_PLATFORM` and not `GLFW_OSMESA_CONTEXT_API`.

PR #2106, opened 2022-05-07
[CMAKE] Rename USE_MSVC_RUNTIME_LIBRARY_DLL to GLFW_USE_MSVC_RUNTIME_LIBRARY_DLL

sasichkamega:
More structured.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,7 +43,7 @@ cmake_dependent_option(GLFW_BUILD_WAYLAND "Build support for Wayland"
 
 cmake_dependent_option(GLFW_USE_HYBRID_HPG "Force use of high-performance GPU on hybrid systems" OFF
                        "WIN32" OFF)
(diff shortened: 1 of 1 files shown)


elmindreda:
In isolation I agree that this is a better name. The current name was picked because other projects (for example Bullet) were already using this one and it seemed possible that CMake would make this variable official, like they had already done for `BUILD_SHARED_LIBS`.

Instead CMake added [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html), which [should be used instead](https://www.glfw.org/docs/latest/compile#compile_options_win32) on newer versions of CMake.

At this point we only offer `USE_MSVC_RUNTIME_LIBRARY_DLL` for backward compatibility with existing projects that embed GLFW. Because of this I won't merge this PR.

PR #2107, opened 2022-05-08
Draft/RFC: Add window blur hint/attribute

### LINKED ISSUE #1823: Acrylic / Blurred Background Effect
I am a bit surprised that I wasn't able to find any other issues or discussions about this topic, especially seeing that we are able to have a transparent window with glfw, I don't believe it would be a long shot to ask for the prominent acrylic / frosted glass /blurred background effect that is now present in almost all modern backends.

- For Windows there is [Acrylic](https://docs.microsoft.com/en-us/windows/uwp/design/style/acrylic)
(Although it is mostly exposed with UWP, there seems to be [workarounds for win32](https://stackoverflow.com/questions/44000217/mimicking-acrylic-in-a-win32-app)) 
- On Mac there is [Vibrancy](https://developer.apple.com/documentation/appkit/nsvisualeffectview)
- Linux should be possible too.

If the [JavaScript people](https://github.com/AryToNeX/Glasscord) can do it, certainly we should be able to do it natively right?

Just to clarify what effect I am talking about:
Image

lsr0:
This adds very basic support for a new GLFW_BLURRED window
hint/attribute, and support in the X11 and null platforms. It is
intended to get feedback on whether this is the right API for this
feature, before diving into the other per-platform
implementations/stubs and documenting.

Note that setting blur at least on X11 is reasonably blind, we set a prop, but have no idea if a compositor is respecting it. As such I haven't tried to implement validated feedback here. We probably could do so for some platforms, however. It is of course likely meaningless on all platforms without transparency/opacity set and will state that in the documentation if & when.

Intended to fix #1823

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -927,6 +927,13 @@ extern "C" {
  */
 #define GLFW_MOUSE_PASSTHROUGH      0x0002000D
 
+/*! @brief Blur behind window hint and attribute
+ *
+ *  Blur behind [window hint](@ref GLFW_BLURRED_hint) or
+ *  [window attribute](@ref GLFW_BLURRED_attrib).
+ */
+#define GLFW_BLURRED                 0x0002000E
+
 /*! @brief Framebuffer bit depth hint.
  *
  *  Framebuffer bit depth [hint](@ref GLFW_RED_BITS).
--- a/src/internal.h
+++ b/src/internal.h
@@ -410,6 +410,7 @@ struct _GLFWwndconfig
     GLFWbool      focusOnShow;
     GLFWbool      mousePassthrough;
     GLFWbool      scaleToMonitor;
+    GLFWbool      blurred;
     struct {
         GLFWbool  retina;
         char      frameName[256];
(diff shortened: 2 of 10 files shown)


ws909:
Does this fill the entire window with a blurred background? How to control the colour and transparency of the blur? Does it work to render transparent OpenGL/Vulkan frame buffers on top of the window without issues? Any performance considerations?

I can probably help with a Cocoa implementation.

lsr0:
On the platforms I've seen this applies blur to all semi-transparent pixels of the window, essentially leaving control over blur areas to the alpha channel. Unfortunately it doesn't scale the blur with the opacity, however, so actually using transparency values anywhere near 0 doesn't look great.

Having separate control in the API for blur regions and/or colours would be great, but doesn't seem to be possible on many platforms.

There are certainly performance concerns to this, but they differ hugely by compositor.

The rationale for this simple proposal is for things like consoles and overlays, where legible transparency would be a great goal.

The default blur colour/transparency level, no idea how feasible that is cross platform. Some just don't have the concept.

PR #2108, opened 2022-05-10
Save and restore original X11 error handler

MartinPulec:
Greetings, I am proposing a patch to save&restore the X11 error handler that user might have set instead of restoring the default one. The point is that default X11 error handler is a bit ugly because it exits the program unconditionally, so the glfw user may override this (as GLFW also does).


elmindreda:
Oh, yes, this needs fixing. Thank you!

Issue #2097, opened 2022-05-02
Add IME support for each platform


daipom:
Hi, we are currently developing functions about IME support.

* https://github.com/clear-code/glfw/tree/3.4-2022-03-31+im-support

The main goal is to enable comfortable Japanese input on each platform by adding APIs to manage IME such as setting the candidate window position.
If possible, we would like to develop a form that can be merged into the upstream GLFW.

We are still in the early stage of implementation and will continue.
We would like to hear your view on the direction from this stage.

The functions we are developing are based on the PR: https://github.com/glfw/glfw/pull/658
The goal of this PR is the same as ours.

The APIs are currently the same as this PR and they are as follows.

* About preedit
    * `glfwSetPreeditCallback`: callback to notify the preedit text
    * `glfwSetPreeditCursorPos`: set the candidate window position
    * `glfwGetPreeditCursorPos`: get the candidate window position
    * `glfwResetPreeditText`: reset the preedit text
* About IME status
    * `glfwSetIMEStatusCallback`: callback to notify the change of ON/OFF of the IME
    * `glfwSetInputMode`: switch ON/OFF of the IME by using new mode `GLFW_IME`
    * `glfwGetInputMode`: get ON/OFF of the IME by using new mode `GLFW_IME`

We added the following fixes to the PR.

* Apply to the latest GLFW
* Fix the code styles and add some refactoring
* Add some improvements
    * Win32: use dynamic load for Imm32
    * Win32: enable to commit the results to the input
    * X11: use over-the-spot by default because on-the-spot of XIM doesn't work well
    * macOS: enable to manage the candidate window position
    * macOS: enable to clear preedit after the commit

# Details for each platform

You can confirm the behavior by `tests/ime`.
Right click to set that position to the position of the candidate window.

We are also developing a sample application using [raylib](https://github.com/clear-code/raylib/tree/better-ime-support), a videogames programing library using GLFW.

* https://github.com/clear-code/RaylibIMEInputSampleApp

## Win32

* Use Imm32.
* All APIs work fine on Windows10.

`tests/ime`

![glfw_test_ime_win](https://user-images.githubusercontent.com/12229857/166201487-8b2209f9-b7ba-4e96-9620-38e710ed5dfa.png)

Sample application

![sample_app_win](https://user-images.githubusercontent.com/12229857/166201505-52e0bf44-4dd2-4c5e-9ea4-8df701b6c145.png)

## X11

[...]

PR #658, opened 2015-12-11
Better IME support

shibukawa:
Add better IME support (#41) to implement on-the-spot IME conversion. My code is based on #643.

I added the following functions:
- glfwSetPreeditCallback() : callback function to notify preedit text
- glfwSetPreeditCursorPos() : teach text cursor position for IME windows
- glfwGetPreeditCursorPos() : get text cursor position for IME windows

They are not big changes. It keeps backward compatibility about API level.
To use IME feature, UI programmers should implement visualization code for preedit texts. 

I implemented platform specific code for Windows/X11/Mac OS X (not Wayland and Mir).

PR #2117, opened 2022-05-24
Draft: Add IME support for each platform

daipom:
This is a fix to the matter we discussed earlier in the issue: #2097.

This fix is based on @shibukawa's  #658 PR.

In that PR, there were some problems with each platform.
This PR fixed its major problems and now works mostly correctly on each platform.
(Minor fixes still need to be made, but we will continue to do so.)

The new APIs are the same as this PR and they are as follows.

* About preedit
    * `glfwSetPreeditCallback`: callback to notify the preedit text
    * `glfwSetPreeditCursorPos`: set the candidate window position
    * `glfwGetPreeditCursorPos`: get the candidate window position
    * `glfwResetPreeditText`: reset the preedit text
* About IME status
    * `glfwSetIMEStatusCallback`: callback to notify the change of ON/OFF of the IME
    * `glfwSetInputMode`: switch ON/OFF of the IME by using new mode `GLFW_IME`
    * `glfwGetInputMode`: get ON/OFF of the IME by using new mode `GLFW_IME`

Supported platforms: Win32, macOS, X11
(We need to consider supporting Wayland, but we would like to support these first.)

You can confirm the behavior by `tests/input_text`.

`tests/input_text` on X11:
input_text_x11

# Win32

* Use Imm32 (dynamic load)
* All APIs work fine on Windows10

# macOS

* All APIs work fine on M1 Monterey 12.3.1.

# X11

* Environment: Ubuntu 20.04 LTS, X11, ibus
* Adopted over-the-spot style by default because on-the-spot style of XIM doesn't work well.
    * You can change the style to on-the-spot by `glfwInitHint(GLFW_X11_ONTHESPOT, GLFW_TRUE)`.
* Over-the-spot style
    * Only `glfwSetPreeditCursorPos` and `glfwGetPreeditCursorPos` work and other APIs don't work.
    * This is OK for this style because the input method side manages the preedit, so the application side doesn't need to manage it except for the candidate window position.
* On-the-spot style
    * `glfwSetPreeditCursorPos` and `glfwGetPreeditCursorPos` don't work, so we can't manage the candidate window position.
    * All APIs of IME status doesn't work too.

# Remaining issues

* X11: Some keys (arrow, Enter, BackSpace, ...) are passed to applications during preediting
* Block-sizes of preedit-callback don't take into account the UTF-32 encoding
* The candidate window doesn't display correctly with fullscreen mode


daipom:
@ashie 's comment on #2097 about TSF support.

original:  https://github.com/glfw/glfw/issues/2097#issuecomment-1134159799

> > ## Windows: TSF support
> > In addition to Imm32, we need to consider supporting TSF (Text Service Framework).
> > [A comment on the original PR](https://github.com/glfw/glfw/pull/658#issuecomment-992132924) pointed out that the TSF needs to be supported instead of Imm32.
> > However, Imm32 seems to work well for now, so it may be enough to support only Imm32 at first?
> > Can we think of a direction to have this merged into upstream without TSF support, and then consider TSF later if necessary?
>
> I'm working on this task with @daipom
>
> TSF is nice to have, but we decided that we won't include it in our first pull request. Here is the reason:
>
> * Windows have a compatible layer "CUAS" (Cicero Unaware Application Support: TSF is formerly known as Cinero) that converts TSF <-> Imm32, so that we can use TSF input methods from Imm32 applications.
>       
>   * CUAS is enabled by default on Windows Vista or later.
>   * Although older CUAS was buggy (at least on Windows XP), recent version of CUAS seems stable enough.
>
> * Although Windows 8's Metro style didn't allow Imm32 and CUAS, Windows 10's tablet mode seems supporting it. We confirmed the actual behavior.
>
> * Although UWP applications can't use Imm32, It's not concerned with GLFW becuase GLFW is a library which depends on Win32.
>
> * We confirmed actual behavior with various input methods of Windows 10, and most of them works fine with Imm32 application.
>
>   * Microsoft IME (Japanese)
>   * ATOK Passport (Japanese)
>   * Google Japanese Input
>   * Microsoft Pinyin (Simplified Chinese)
>   * Microsoft Bopomofo (Traditional Chinese)
>   * Chinese Traditional Array (Traditional Chinese)
>
> * We don't use TSF specific features yet.
>
> * Since TSF requires COM (Component Object Model), we can't build it by MinGW.
>
>   * Even when we use Visual Studio, we have to greatly change the project configuration.
>
>
> We think the comment [`Imm32 is problematic on Window10 as Microsoft has dropped support of it.`](https://github.com/glfw/glfw/pull/658#issuecomment-992132924) is incorrect as mentioned above.
>
> refs:
>
> * https://bugs.chromium.org/p/chromium/issues/detail?id=225189#c1
>
> * https://nyaruru.hatenablog.com/entry/20070308/p1 (Written in Japanese)
>
> * https://nyaruru.hatenablog.com/entry/20070309/p1 (Written in Japanese)
>
> * https://game.watch.impress.co.jp/docs/20070927/wv.htm (Written in Japanese)
>
> * https://ascii.jp/elem/000/000/757/757721/2/ (Written in Japanese)

daipom:
@ashie 's comment on #2097 about Wayland support.

original:  https://github.com/glfw/glfw/issues/2097#issuecomment-1134355093

> > ## GNU/Linux: Wayland support
> > In addition to XIM, we need to consider supporting Wayland.
> > There is an idea supporting Fcitx directly in [a comment on the original PR](https://github.com/glfw/glfw/pull/658#issuecomment-165163366). However, we are not considering this direction because over-the-spot style seems to work well on XIM and we can support Wayland if needed.
> > What we would like to discuss is whether there is any possibility to go in the direction of only supporting XIM at first?
> 
> Wayland support is a bit complicated but we think we need to support it. Because recent major distributions are migrating from X11 to Wayland, and Wayland doesn't have global positioning protocol or API, so that it's always needed to support Wayland's text input protocol to show candidate window in correct position. Supporting Fcitx or IBus directly doesn't resolve this issue.
> 
> An complicated issue of Wayland is that there are some protocols and each of them are incompatible:
> 
> * [text-input-v1](https://cgit.freedesktop.org/wayland/wayland-protocols/tree/unstable/text-input/text-input-unstable-v1.xml)
> 
> * [text-input-v3](https://cgit.freedesktop.org/wayland/wayland-protocols/tree/unstable/text-input/text-input-unstable-v3.xml)
> 
> * [gtk-text-input](https://github.com/GNOME/gtk/blob/eb4d736a4dea14528d51475d97bbf013fb81571a/modules/input/gtk-text-input.xml)
> 
> 
> I'm not sure but it seems that gtk-text-input doesn't seem used in recent days (It's not included in GTK4, probably dropped). To support 2 major desktop environment "GNOME" and "KDE" on GNU/Linux, probably supporting text-input-v3 is enough. But another major window system "Weston" is also used widely on embedded systems and it supports only text-input-v1. So we may need to support both text-input-v1 and text-input-v3.
> 
> We have experience to implement them (e.g. https://github.com/ashie/gtk-wayland-textinputv1) so that we can implement it in GLFW as well, but not yet start to work on it.

daipom:
> Remaining issues
> * X11: Some keys (arrow, Enter, BackSpace, ...) are passed to applications during preediting

This is caused by `_glfwInputKey` being called even if the result of `XFilterEvent` is True.

* https://github.com/clear-code/glfw/blob/5d7a89e5690600e4bb88f06c5721a18f01fea131/src/x11_window.c#L1325
* https://github.com/clear-code/glfw/blob/5d7a89e5690600e4bb88f06c5721a18f01fea131/src/x11_window.c#L1437

I confirmed that inserting the following codes before the `_glfwInputKey` solved this issue.

```c
if (filtered)
    return;
```

Could you please teach me why `_glfwInputKey` is called even if the result of `XFilterEvent` is True?
There seems to be a reason for this and I am wondering how to solve it.

daipom:
> Remaining issues
>     * The candidate window doesn't display correctly with fullscreen mode

About the fullscreen issue on X11:

Changing `NET_WM_BYPASS_COMPOSITOR` value to `2` from `1` solves this issue on X11.

* `NET_WM_BYPASS_COMPOSITOR`: https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm45446104333040

```diff
diff --git a/src/x11_window.c b/src/x11_window.c
index 5eec344f..c2214509 100644
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -362,7 +362,7 @@ static void updateWindowMode(_GLFWwindow* window)
         // Enable compositor bypass
         if (!window->x11.transparent)
         {
-            const unsigned long value = 1;
+            const unsigned long value = 2;
 
             XChangeProperty(_glfw.x11.display,  window->x11.handle,
                             _glfw.x11.NET_WM_BYPASS_COMPOSITOR, XA_CARDINAL, 32,
```

This change may slightly worsen rendering performance.
Instead of always changing it to 2, should I make it possible to specify the value by `Hint` or something?

daipom:
> Remaining issues
>     * The candidate window doesn't display correctly with fullscreen mode

About the fullscreen issue on Windows and macOS:

It seems to be caused by the way creating a window on these platforms.

### Windows

On Windows, the IME display flashes the screen,
similar to the problem of transparency not working with fullscreen mode.

Weirdly, these problems occur when the window size exactly matches the screen size.
We can solve this problem by making the window height or width 1 larger than the screen size,
at the same time enabling transparency to work with fullscreen mode.

If you can tolerate this weird solution,
we can solve it by adding a new hint to make the window height or width 1 larger than the screen size.
(Maybe windows-only glfwWindowHint such as `GLFW_WIN32_FULLSCREEN_IME`)

### macOS

On macOS, the candidate window can't be displayed on fullscreen.
However, if we make it fullscreen by pressing the green button on the window, it works fine.
if we want IME input on fullscreen, we can use the green button to make it fullscreen instead of the GLFW function.

ashie:
> # Remaining issues
> 
>     * X11: Some keys (arrow, Enter, BackSpace, ...) are passed to applications during preediting
> 
>     * Block-sizes of preedit-callback don't take into account the UTF-32 encoding
> 
>     * The candidate window doesn't display correctly with fullscreen mode

In addition, we need to add a way to get caret position of preedit string.

daipom:
> In addition, we need to add a way to get caret position of preedit string.

I added this at https://github.com/glfw/glfw/pull/2117/commits/0ba4b3655707cce780e47679798f9e4d7023cc1a.

ronchaine:
Wayland IME is a bit of a mess.  Besides text-input-v1 and text-input-v3, there's also unstable input-method-v2, used by wlroots (which is pretty much everything not Weston, KDE or GNOME)

https://github.com/fcitx/fcitx5/issues/206  fcitx5 has some more details, and a blog post https://www.csslayer.info/wordpress/fcitx-dev/chrome-state-of-input-method-on-wayland/ has some good info as well (in Chinese).

ashie:
AFAIK input-method-v2 is a protocol for input methods, not affected to applications.

daipom:
Since I re-created this PR as #2130, I close this.

Issue #2121, opened 2022-05-27
Add a window hint for setting app_id on wayland


michaelpennington:
The equivalent of WM_CLASS for X11 is [app_id](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id). Allow setting this through a window hint string.

flysand7:
This option is not documented on the website with HTML docs, perhaps it should be added there too?

elmindreda:
@flysand7 [Done](https://www.glfw.org/docs/3.4/). Thanks!

PR #2122, opened 2022-05-27
Add a window hint for setting app_id on wayland

michaelpennington:
Simple change to support setting app_id on wayland windows, as specified in #2121 .

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -157,6 +157,7 @@ video tutorials.
  - Peoro
  - Braden Pellett
  - Christopher Pelloux
+ - Michael Pennington
  - Arturo J. Pérez
  - Vladimir Perminov
  - Anthony Pesch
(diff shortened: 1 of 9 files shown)


TheBrokenRail:
> @TheBrokenRail When would that be preferable over this window hint string?

Mainly so that you don't *have* to set the hint string. Right now on X11, everything just works by default because the `WM_CLASS` default is the window title. Of course, you can change this if you want to, but the default behavior makes things really simple: just set `StartupWMClass` to the window's title in your `.desktop` files and everything's good to go.

On Wayland, `app_id` isn't set at all right now, so you can't assign it to a `.desktop` file at all (at least on GNOME, `StartupWMClass` seems to correspond to the `app_id` on Wayland). And if this PR were merged as is, you'd still have to specify the `app_id` manually, rather than it defaulting to the window's title like X11.

PR #2128, opened 2022-06-10
[RFC] Drag & drop events

### LINKED ISSUE #1898: Enter event for drag'n'drop
Something that would be useful is to have support for an event when a user begins and ends a dragging operation.
My use case is I want to be able to hint/show where the user can drop something, which isn't possible today as the current drop callback happens at the very end.

Just for reference this is what Qt supports https://doc.qt.io/qt-5/qwidget.html#dragEnterEvent which is 4 "events" (in practice 4 C++ methods one can implement)

```C++
void QWidget::dragEnterEvent(QDragEnterEvent *event)
```
```
This event handler is called when a drag is in progress and the mouse enters this widget. The event is passed in the event parameter.

If the event is ignored, the widget won't receive any drag move events.

See the Drag-and-drop documentation for an overview of how to provide drag-and-drop in your application.
```
```C++
void QWidget::dragLeaveEvent(QDragLeaveEvent *event)
```
```
This event handler is called when a drag is in progress and the mouse leaves this widget. The event is passed in the event parameter.

See the Drag-and-drop documentation for an overview of how to provide drag-and-drop in your application.
```
```C++
void QWidget::dragMoveEvent(QDragMoveEvent *event)
```
```
This event handler is called if a drag is in progress, and when any of the following conditions occur: the cursor enters this widget, the cursor moves within this widget, or a modifier key is pressed on the keyboard while this widget has the focus. The event is passed in the event parameter.

See the Drag-and-drop documentation for an overview of how to provide drag-and-drop in your application.
```
```C++
void QWidget::dropEvent(QDropEvent *event)
```
```
This event handler is called when the drag is dropped on this widget. The event is passed in the event parameter.

See the Drag-and-drop documentation for an overview of how to provide drag-and-drop in your application.
```

I'm not sure if glfw needs to support all of the above, but they are all useful when you need to implement drag`n`drop

falbrechtskirchinger:
As discussed, I've started work on addressing #1898. I'd also like to lay the foundation for supporting other data types and at least support plain text in addition to paths.

I'd appreciate some feedback on the outline and answers to my questions below. What little code exists is not ready for review.

#### (Rough) outline of the proposed changes

* Add a window attribute that indicates whether a drop operation is in progress. To avoid any potential confusion with `glfwDragWindow`, I've decided against naming it `GLFW_DRAGGING`. I'd also like to distinguish between incoming and outgoing (should that ever be added) drag & drop operations. The current placeholder name is `GLFW_DND_IN_PROGRESS`.
* Add a callback for enter (and leave) events. The callback should receive information about the supported formats (text, paths, ...) and return (possibly via out parameter) whether to accept or reject the drop and what action might be taken (copy, link, or move).
* Add a callback for drag over events. Similar to the callback above but it also receives a cursor position. What does this mean for the regular cursor position callback? Which, if any, platforms keep generating regular updates? The same question applies to modifier keys, which are part of drag over events (on some if not all platforms).
* It should be possible to combine these two callbacks into one if desired.
* Add another (extended) drop callback (`GLWFdropfunex`?) that can receive more than a list of paths and returns (possibly via out parameter) the type of action taken (copy, link, or move). How different types of data are passed needs to be decided. Options are:
    ```c
    struct {
        int type; // GLFW_DND_TEXT, GLFW_DND_PATHS, ...
        int count;
        union {
            const char *text;
            const char **paths;
            // ...
        } data;
    };

    struct {
        int type;
        int count;
        void *data;
    };
    ```

#### Other open questions

* I've not paid any attention to ABI compatibility yet. Should I?
* When should Win32 DLLs be manually loaded and their symbols be retrieved via `GetProcAddress()`? Most are, which is why I've followed that practice for (most) new functions, but some functions are used without being explicitly loaded, e.g., `DragQueryPoint`, `DragQueryFileW`, `DragFinish`.

#### Status

* Win32: Dropping paths via `IDropTarget` has been implemented.

#### ToDo

- [ ] Complete implementation as outlined pending feedback.
- [ ] Documentation.
- [ ] Address items from contributing guidelines.
- [ ] Remove test code from `traingle-opengl.c`.
- [ ] Remove dead code (WM_DROPFILES, IDropTarget_QueryInterface, ...).

Closes #1898.


rankaisija64:
Looks really promising! It would be neat to see the idea of initiating drag and drop from glfw based application in the same way, having the formats, counts, data, etc set and providing data from the application to others.

ws909:
Would it not be more intuitive to use a `GLFW_DRAG_DROP_` prefix instead, or `GLFW_DD_`? `DND` sounds like a slang abbreviation for Dungeons and Dragons. It's a tiny bit distractive to me, but not a major issue; quite a tiny one in fact.

falbrechtskirchinger:
> Would it not be more intuitive to use a `GLFW_DRAG_DROP_` prefix instead, or `GLFW_DD_`? `DND` sounds like a slang abbreviation for Dungeons and Dragons. It's a tiny bit distractive to me, but not a major issue; quite a tiny one in fact.

I have no strong feelings about my chosen prefix. `DND` is a common abbreviation for drag and drop IMHO, but I don't mind changing it.

> Looks really promising! It would be neat to see the idea of initiating drag and drop from glfw based application in the same way, having the formats, counts, data, etc set and providing data from the application to others.

I'm trying to design the API with drag support in mind. Ideally, we'd add a proper abstraction for Windows' clipboard format and only expose MIME types in the API (like all other platforms do by default). I don't know how that would be received and I'm still waiting for any feedback or even interest from the maintainer before I proceed.

The project that inspired this PR is also on hold, so besides passive-aggressive digs, I'm not going to resort to begging for feedback at this point. ;-)

richardwilkes:
Is this effectively dead now?

I'm also interested full drag & drop support that I could use between windows of my application in addition to just dragging a file from the desktop.

I have a lot of experience on macOS and am willing to help there, if needed. My knowledge on other platforms is much less, although I can usually bumble along.

falbrechtskirchinger:
> Is this effectively dead now?

There doesn't seem to be any interest from the maintainers. I'd resume work on this once I get some feedback. It's also been a really long time, so I'd have to re-familiarize myself with everything.

The project I needed this for is on permanent hold and I have zero motivation to push the issue.

richardwilkes:
Well, that's disappointing. I'll eventually implement something for myself, but was hoping to see it in glfw proper so that I didn't have to constantly rework it as things changed.

emoon:
While I currently don't have time to write some of the code I would still love to see this feature being part of glfw

Brat-vseznamus:
Found this merge request. As a developer, I am very interested in this feature, because currently working on the same API, so I will be excited to see it come out.

Brat-vseznamus:
> Found this merge request. As a developer, I am very interested in this feature, because currently working on the same API, so I will be excited to see it come out.

If I can help somehow - reply on this comment

Renardjojo:
@elmindreda Can this merge request be inspected? It seems a lot of people would like this feature.

heyallnorahere:
Any plans to revive this feature? Would be very helpful

Issue #10, opened 2013-04-25
Gamma ramps of sizes other than 256 are not supported


elmindreda:
This can be fixed by changing the `GLFWgammaramp` struct to

``` C
typedef struct
{
    unsigned short* red;
    unsigned short* green;
    unsigned short* blue;
    unsigned int size;
} GLFWgammaramp;
```

However, this puts the responsibility of freeing the arrays returned by `glfwGetGammaRamp` on the client, forcing the choice between adding `glfwFreeGammaRamp` or endless complaints of memory leaks and (when beginners mix debug and release MS libcs) crashes.

PR #2130, opened 2022-06-13
Add IME support for each platform

daipom:
This PR is the re-creation of #2117 to rename and rebase.

This is a fix to the matter we discussed earlier in the issue: #2097.

This fix is based on @shibukawa's  #658 PR.

In that PR, there were some problems with each platform.
This PR fixed its major problems and now works correctly on each platform.

The new APIs are the same as this PR and they are as follows.

* About preedit
  * `glfwSetPreeditCallback`: callback to notify the preedit text
  * `glfwSetPreeditCursorRectangle`: set the preedit cursor area that is used to decide the candidate window
  * `glfwGetPreeditCursorRectangle`: get the preedit cursor area
  * `glfwResetPreeditText`: reset the preedit text
* About IME status
  * `glfwSetIMEStatusCallback`: callback to notify the change of ON/OFF of the IME
  * `glfwSetInputMode`: switch ON/OFF of the IME by using new mode `GLFW_IME`
  * `glfwGetInputMode`: get ON/OFF of the IME by using new mode `GLFW_IME`
* About the feature to take candidate list (Win32 ONLY)
  * `glfwSetPreeditCandidateCallback`: callback to notify the preedit candidate
  * `glfwGetPreeditCandidate`: get the current candiadate text of the specified index
  * This feature is currently only for Win32.
  * For a detailed explanation, please see below.

Supported platforms: Win32, macOS, X11, Wayland

You can confirm the behavior by `tests/input_text`. (See below for details).

`tests/input_text` on X11:
input_text_x11

# Win32

* Use Imm32 (dynamic load)
* All APIs work fine on Windows10

# macOS

* All APIs work fine on M1 Monterey 12.3.1.

# X11

* Environment: Ubuntu 20.04 LTS, X11, ibus
* Adopted over-the-spot style by default because on-the-spot style of XIM doesn't work well.
  * You can change the style to on-the-spot by `glfwInitHint(GLFW_X11_ONTHESPOT, GLFW_TRUE)`.
* Over-the-spot style
  * Only `glfwSetPreeditCursorPos` and `glfwGetPreeditCursorPos` work and other APIs don't work.
  * This is OK for this style because the input method side manages the preedit, so the application side doesn't need to manage it except for the candidate window position.
* On-the-spot style
  * `glfwSetPreeditCursorPos` and `glfwGetPreeditCursorPos` don't work, so we can't manage the candidate window position.
  * All APIs of IME status doesn't work too.

# Wayland support

* Support wayland protocols: `text_input_unstable_v3`, `text_input_unstable_v1`
  * https://cgit.freedesktop.org/wayland/wayland-protocols/tree/unstable/text-input/text-input-unstable-v3.xml
  * https://cgit.freedesktop.org/wayland/wayland-protocols/tree/unstable/text-input/text-input-unstable-v1.xml
* Currently, we support minimum required APIs for on-the-spot IME input: `glfwSetPreeditCallback`, `glfwSetPreeditCursorPos` and `glfwGetPreeditCursorPos`.

# Feature to take candidate list (Win32 ONLY)

Usually, the IME displays the candidate window, so the app side doesn't need to get canidate list.

However, sometimes, we need to display it on the app side.
For example, it could be for displaying it correctly on exclusive fullscreen, or for using a custom design.

This is especially common in games for Windows.

This feature is OFF by default.
We can use this by setting the `GLFW_MANAGE_PREEDIT_CANDIDATE` init hint to `GLFW_TRUE`.

We can confirm the behavior in `tests/input_text` with the `-c` option.

Currently, this feature is available only on Win32, since this feature is especially common on Windows.
We'd like to make it available on every platform if possible, but it may be difficult because of platform specific issues.

This feature is also involved in the issue with fullscreen and IME.
If fullscreen is strongly exclusive, then the candidate window of IME can't be displayed correctly.
This problem is especially serious in Windows, so this feature is one workaround for fullscreen on Windows.

We also have another fix that reduces fullscreen exclusivity for IME, but the fix for Windows is tricky... (please see `Related issues` below)
We would like to make this fix as another PR depending on the status of this PR, since this PR is already a large fix.

# Related issues

* X11: Some keys (arrow, Enter, BackSpace, ...) are passed to applications during preediting
  * This will be fixed by #1972
* The candidate window doesn't display correctly with fullscreen mode
  * We have fixed this issue in the following branch since this PR is already a large fix.
    * https://github.com/clear-code/glfw/tree/im-support-fullscreen 
  * We would like to make this fix as another PR depending on the status of this PR.

# `tests/input_text`

We have created a new test `input_text` for this feature.
Although this feature is available for various languages using IME, this test is currently intended for Japanese in font management.
We need to prepare a suitable font to test this feature.

The following is how to prepare the Japanese font.
Note: The font file needs to be TTF since `Nuklear` supports only TTF files.

**Win32**

* `Yu Mincho` font is used by default.
* This font is in the FOD packages on Windows, so it will be automatically added when you add Japanese to the input source. (Maybe you need windows-update after the setting).
  * https://learn.microsoft.com/en-us/typography/fonts/windows_10_font_list#japanese-supplemental-fonts
  * Or you can install it manually in `Manage optional features` of `Apps & features`.

**macOS**

* `Arial Unicode MS` font is used by default.
* We assume it is installed in macOS by default.
* If it is not installed, please install it in the `Font Book` app.

**X11 or Wayland**

* Please install the `fontconfig` library so that you can select fonts in GUI.
  * `$ sudo apt install fontconfig`
* Please install a Japanese font, such as `fonts-ipafont`.
  * `$ sudo apt install fonts-ipafont`
* Then you can select `IPAGothic` or `IPAMincho` in GUI.
* If they don't exist, please try the following command and launch the test again.
  * `$ fc-cache -fv`

**Specify the font-file path**

You can also specify a specific TTF font file.
You can comment out this line and set the full path of the font file.
Then the font is selected by default.

https://github.com/glfw/glfw/blob/1943dff6cf6f90c5a57cbc37010bab6d1632cefd/tests/input_text.c#L53

**Test other languages**

For now, we consider it sufficient if we can confirm this feature in Japanese.
The following is a note for future reference about what we need to do in order to test this feature in other languages.

* Fix Unicode range table:
  * https://github.com/glfw/glfw/blob/1943dff6cf6f90c5a57cbc37010bab6d1632cefd/tests/input_text.c#L97-L106
* Add the loaded fonts for each platform:
  * https://github.com/glfw/glfw/blob/1943dff6cf6f90c5a57cbc37010bab6d1632cefd/tests/input_text.c#L290-L313
  * On macOS, the current default font `Arial Unicode MS`  is assumed to be available in various languages other than Japanese, but if you need a language not supported in this font,  please add an appropriate font.
  * On X11 or Wayland, we can use `fontconfig` to select fonts by GUI, so we don't have to load the specific font in the code.

**Credits**

- Takuro Ashie
- Daijiro Fukuda
- Ryo Ichinose (from #658)
- Yasutaka Kumei (from #658)
- Yoshiki Shibukawa (from #658)
- xfangfang


daipom:
Sorry, I had to recreate the PR #2117.
Let us continue working here.

daipom:
> Remaining issues
>     * The candidate window doesn't display correctly with fullscreen mode

About the fullscreen issue on X11:

Changing `NET_WM_BYPASS_COMPOSITOR` value to `2` from `1` solves this issue on X11.

* `NET_WM_BYPASS_COMPOSITOR`: https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm45446104333040

```diff
diff --git a/src/x11_window.c b/src/x11_window.c
index 5eec344f..c2214509 100644
--- a/src/x11_window.c
+++ b/src/x11_window.c
@@ -362,7 +362,7 @@ static void updateWindowMode(_GLFWwindow* window)
         // Enable compositor bypass
         if (!window->x11.transparent)
         {
-            const unsigned long value = 1;
+            const unsigned long value = 2;
 
             XChangeProperty(_glfw.x11.display,  window->x11.handle,
                             _glfw.x11.NET_WM_BYPASS_COMPOSITOR, XA_CARDINAL, 32,
```

This change may slightly worsen rendering performance.
Instead of always changing it to 2, should I make it possible to specify the value by `Hint` or something?

daipom:
> Remaining issues
>     * The candidate window doesn't display correctly with fullscreen mode

About the fullscreen issue on Windows and macOS:

It seems to be caused by the way creating a window on these platforms.

### Windows

On Windows, the IME display flashes the screen,
similar to the problem of transparency not working with fullscreen mode.

Weirdly, these problems occur when the window size exactly matches the screen size.
We can solve this problem by making the window height or width 1 larger than the screen size,
at the same time enabling transparency to work with fullscreen mode.

If you can tolerate this weird solution,
we can solve it by adding a new hint to make the window height or width 1 larger than the screen size.
(Maybe windows-only glfwWindowHint such as `GLFW_WIN32_FULLSCREEN_IME`)

### macOS

On macOS, the candidate window can't be displayed on fullscreen.
However, if we make it fullscreen by pressing the green button on the window, it works fine.
I'm assuming that if we can make it fullscreen as well as when we press that green button, it will solve the problem.

daipom:
> > Remaining issues
> > 
> > * X11: Some keys (arrow, Enter, BackSpace, ...) are passed to applications during preediting
> 
> This is caused by `_glfwInputKey` being called even if the result of `XFilterEvent` is True.
> 
>     * https://github.com/clear-code/glfw/blob/5d7a89e5690600e4bb88f06c5721a18f01fea131/src/x11_window.c#L1325
> 
>     * https://github.com/clear-code/glfw/blob/5d7a89e5690600e4bb88f06c5721a18f01fea131/src/x11_window.c#L1437
> 
> 
> I confirmed that inserting the following codes before the `_glfwInputKey` solved this issue.
> 
> ```c
> if (filtered)
>     return;
> ```
> 
> Could you please teach me why `_glfwInputKey` is called even if the result of `XFilterEvent` is True? There seems to be a reason for this and I am wondering how to solve it.

This fix was already in existing PR: #1972
If this is merged, this issue is solved.

daipom:
We would like to separate the fullscreen-IM-support from this PR. (Wayland-support too).

So all work on this PR is completed now.

daipom:
I have rebased to the latest mater.

daipom:
Thank you so much for your review in advance!

daipom:
I have rebased to the latest mater.

daipom:
We refactored some variable names.

daipom:
We refactored some variables again.

ashie:
>     * Wayland support
>       
>       * We will handle this problem in the future PR

We've add a commit to support wayland to this PR:
https://github.com/glfw/glfw/pull/2130/commits/238d5b395b45c7ae76b3efb4ef56a4783e7ce71e

daipom:
We fixed several bugs.

We will reorganize this branch soon because the number of commits has become too large.

daipom:
I have reorganized the commits.

daipom:
I add an explanation of how to use `tests/input_text` in order to test this feature to the initial comment.

* https://github.com/glfw/glfw/pull/2130#issue-1269235140

If you have any questions about this feature or need more explanation, please let us know!

daipom:
We have improved the implementation of IME status APIs for macOS.

daipom:
I have fixed some APIs.

daipom:
We have rebased this branch again.

* Add guard logic for `_glfw_calloc()` and `_glfw_realloc()`.
* Fix missing initialization of `preedit.cursorWidth`.
* Improve on-the-spot logic for X11, although it is not used by default.
* Refactor test code.

daipom:
We have rebased this branch again.

* Improve document.

hujun-open:
I hope this get merged soon, proper IME support is really important for non-English user, specially CJK users

daipom:
Sorry, now we are going to merge this feature into this PR.
We will fix the failures on CI soon.

> * Win32: Add a new API to get the candidate list.
>   
>   * We have added this feature in the following branch.
>     
>     * https://github.com/clear-code/glfw/tree/im-support-drawing-candidate
>   * Usually, the IME displays the candidate window, so the app side doesn't need to get the list.
>   * However, sometimes, we need to display it on the app side.
>     
>     * For example, it could be for displaying it correctly on exclusive fullscreen, or for using a custom design.
>     * This is especially common in games for Windows.
>   * We would like to make this fix as another PR depending on the status of this PR and the fullscreen issue above.

daipom:
We add the feature to take candidate list (Win32 ONLY).
I have updated the initial comment on this PR, so please see it about this feature.

If you would like this feature to be separated into another PR, please let me know.
As for this feature, it is summarized in the last commit and can be easily removed.

daipom:
I have rebased to the latest mater to solve the conficts.

xfangfang:
@daipom 
Thank you very much for this pr. I found some problems on MacOS. The detailed description is written here:
https://github.com/clear-code/glfw/pull/2

daipom:
I have rebased this branch again.

* Squash the fix of https://github.com/clear-code/glfw/pull/2 into 75cd5f95c03dfe0a42effc3f046a5753cfa9d53a.

Thanks so much @xfangfang !

bjorndm:
Thanks doe making this! I urgently need this patch for a Japanese healthcare application.

daipom:
I have rebased this branch to the latest mater to solve the conflicts.

daipom:
Following https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-feature,
I'm adding Changelog and Credits.

Sorry for missing this.

daipom:
Added credits and changelog entries.

Please let me know if there are any points that need explanation or correction, or if there are any areas where the PR should be separated :)

ws909:
You should add the changelog additions to the top of the list.

daipom:
> You should add the changelog additions to the top of the list.

@ws909 Thanks! I have fixed it!

fldx:
Hello, I have a question regarding the implementation of the `glfwSetInputMode: switch ON/OFF of the IME by using new mode GLFW_IME` feature. Has this feature been implemented? I cannot find the definition of `GLFW_IME`.

daipom:
@fldx Please note that it doesn't work on X11.

kappa-maintainer:
I tried this PR in lwjgl3, it fixed every IME-related error I found in current release. Hope it can be merged soon!

fldx:
> @fldx Please note that it doesn't work on X11.

I found it, thank you. The implemented code seems to have not been merged into the master branch.

daipom:
> > @fldx Please note that it doesn't work on X11.
> 
> 
> 
> I found it, thank you. The implemented code seems to have not been merged into the master branch.

Yes, it is newly implemented in this PR, so doesn't exist in the master branch.

ashie:
I've rebased this pull request on the latest master.

lilydjwg:
Hi, I just tried this patch with Minecraft. Unfortunately it crashed with a null pointer error (SIGSEGV; I don't know where, sorry) when I tried to type in the chat (the input method's window appeared and no text was seen in the game).

elmindreda:
Hello everyone. Thank you for your patience. We're currently finishing the last bits of 3.4, which is focused on Wayland support. The next minor release, 3.5, will focus on improvement to keyboard input and this PR is a large part of that.

I will be moving this over to the 3.5 milestone shortly. That doesn't mean it's being pushed back; it means it's next.

ashie:
@lilydjwg Could you tell me about your environment?

* How did you install Minecraft and GLFW with IME support?
* Which OS are you using? e.g.) Fedora, Ubuntu, etc...
* Which version of your OS?
* Which desktop environment are you using? e.g.) GNOME, KDE, sway, weston, etc...
* Which input method are you using? e.g.) IBus, Fcitx5, etc...
* Which input method engine are you using? e.g.) ibus-chewing, fcitx5-chewing, etc..

kappa-maintainer:
> Hi, I just tried this patch with Minecraft on Wayland. Unfortunately it crashed with a null pointer error (SIGSEGV; I don't know where, sorry) when I tried to type in the chat (the input method's window appeared and no text was seen in the game).

You shouldn't add this by yourself, please use the latest LWJGL3 snapshot, they already merged this PR in their glfw branch

lilydjwg:
I'm using PrismLauncher with self-compiled glfw on Arch Linux. I'm using fcitx5 on Wayfire git with some IM fixes.

I don't see snapshot versions of LWJGL 3 in my launcher.

kappa-maintainer:
Maybe you compiled it wrong, try the build I compiled earlier:
https://github.com/CleanroomMC/glfw/releases/tag/3.4.1-1
This is not up-to-date since LWJGL did this officially
You could get snapshot lwjgl here
https://www.lwjgl.org/customize
iirc sonatype is hosting it too

lilydjwg:
OK, I've got the crashing location (I needed to `-DCMAKE_BUILD_TYPE=RelWithDebInfo`).

```
[...signal handler...]
#9  0x0000000000000000 in ?? ()
#10 0x00007f4f3a292843 in textInputV3CommitString (data=0x7f4fb57b0b00, textInputV3=<optimized out>, text=<optimized out>) at .../glfw/src/wl_window.c:2163
2163            window->callbacks.character((GLFWwindow*) window, codepoint);
```

`window->callbacks.character` is null.

I also tried the snapshot version of lwjgl from https://www.lwjgl.org/customize. Same crash (only without symbols).

ashie:
> ```
> [...signal handler...]
> #9  0x0000000000000000 in ?? ()
> #10 0x00007f4f3a292843 in textInputV3CommitString (data=0x7f4fb57b0b00, textInputV3=<optimized out>, text=<optimized out>) at .../glfw/src/wl_window.c:2163
> 2163            window->callbacks.character((GLFWwindow*) window, codepoint);
> ```
> 
> `window->callbacks.character` is null.

Hmm, it seems that we need `NULL` pointer guard here as win32 backend does.
Thanks for your report!

lilydjwg:
Adding a guard prevents the crash, but no text can be committed.

ashie:
> Adding a guard prevents the crash, but no text can be committed.

`window->callbacks.character == NULL` means that the application doesn't set the callback.
I'm not sure but probably it's hard to input text from IME in this case, need to modify the application.

lilydjwg:
I see, but minecraft can work with XIM on Xwayland (even without this patch). Does it use another mechanism?

ashie:
> Hmm, it seems that we need `NULL` pointer guard here as win32 backend does. Thanks for your report!

Addressed in 5ad451f2eaa6b5c3c3718257f5d52ddfddfd78bb

ashie:
> I see, but minecraft can work with XIM on Xwayland (even without this patch). Does it use another mechanism?

I'm not sure yet.

kappa-maintainer:
For using this patch under wayland, XMODIFIERS @im=fcitx is needed

ashie:
> For using this patch under wayland, XMODIFIERS @im=fcitx is needed

It means GLFW works with X11 backend and uses XIM as input method.
It's not expected for this patch although it's fine for users to use the X11 backend to avoid problems with the wayland backend.
This patch intend to use text-input-v3 or text-input-v1 protocol on Wayland backend.

VVD:
Are there any plans to release a new version with these patches soon?

Silverlan:
> Are there any plans to release a new version with these patches soon?

I'm curious as well, are the patches stable yet, or are there still some unresolved issues?

VVD:
These patches are required for the latest version of lwjgl to run Minecraft 1.20.x.

elmindreda:
@VVD @Silverlan Yes, 3.4 is out so I've started reviewing and preparing to merge this.

VVD:
```
if (GLFW_BUILD_X11 OR GLFW_BUILD_WAYLAND)
    find_package(Fontconfig)
    if (FONTCONFIG_FOUND)
        target_compile_definitions(input_text PRIVATE FONTCONFIG_ENABLED)
        target_link_libraries(input_text fontconfig)
    endif()
endif()
```
This is wrong.
If fontconfig installed it just grab it silently. But if isn't - it ignores fontconfig.
Where is option to force turn off search fontconfig even if it's installed?

ashie:
Ah, it would be better to be loaded dynamically and make runtime option, not build time.
It 's used only in an example program input_text, and just only an optional feature to list fonts.
We'll fix it later.
Thank you for pointing out it!

VVD:
> Ah, it would be better to be loaded dynamically and make runtime option, not build time. It 's used only in an example program `input_text`, and just only an optional feature to list available fonts. We'll fix it later. Thank you for pointing out it!

And `input_text` not installed with install target, but build without errors.

ashie:
> Ah, it would be better to be loaded dynamically and make runtime option, not build time.

Done.

VVD:
ping

xfangfang:
@ashie Hi, I`m using the Chinese input method and have noticed that `glfwSetPreeditCursorRectangle` occasionally fails under Windows. I am not sure if other languages may have similar issues.
I just submitted a PR, with specific descriptions in: https://github.com/clear-code/glfw/pull/4

ashie:
Ah yes, `ImmSetCompositionWindow()` should be handled as well. Thanks for you feedback!
Most modern applications supports on-the-spot style pre-edit text so that composition window isn't used so often in usual, at least in Japan. Anyway, we should also handle `ImmSetCompositionWindow()` correctly.

daipom:
@xfangfang @ashie Thanks!
As I could not assume the case where the composition window is used, it is disabled by the following code:
https://github.com/glfw/glfw/blob/0c57adf4787b5801e6660f028a8f5a89ead7b9b2/src/win32_window.c#L910-L912

If `ImmSetCompositionWindow()` resolves the issue, it could mean that the position of the composition window needs to be managed regardless of whether it is enabled or disabled.
Or, if some input methods need to use the composition window, we may also need a feature to enable it.

HappyRespawnanchor:
Will this be merged soon? Proper IME support is very important for CJK users

xfangfang:
@ashie @daipom 
Hi, I just made some adjustments to the implementation for Wayland, the experience has improved for my application, but I'm not sure if this is what this PR requires. For a detailed description, please see: https://github.com/clear-code/glfw/pull/5

VVD:
Can you merge at least "preedit" part?

Houfeng:
This CR has been created for two years, can it still be merged?

kappa-maintainer:
I tried to use `glfwSetInputMode` to block IME, but it didn't work as I expected
Some example of this:
Windows: https://github.com/reserveword/IMBlocker/blob/master/common/src/main/java/io/github/reserveword/imblocker/common/IMManagerWindows.java
Mac: https://github.com/reserveword/IMBlocker/blob/master/common/src/main/java/io/github/reserveword/imblocker/common/IMManagerMac.java
X11: https://github.com/Yesterday17/KokoaLinux/blob/master/libkokoa/library.c

ghostflyby:
The cocoa implementation is switching between ASCII and first system language input source instead of toggling text input, which may cause untended IME changes for multilingual IME users or when switching between windows. Is that intended?

ashie:
> I tried to use `glfwSetInputMode` to block IME, but it didn't work as I expected Some example of this: Windows: https://github.com/reserveword/IMBlocker/blob/master/common/src/main/java/io/github/reserveword/imblocker/common/IMManagerWindows.java Mac: https://github.com/reserveword/IMBlocker/blob/master/common/src/main/java/io/github/reserveword/imblocker/common/IMManagerMac.java Wayland: text-input-v3 has enable and disable while v1 has activate and deactivate, tested and it worked

Thanks for your comment, I've checked it.

Unfortunately we don't intend  to behave as you expect for the current implementation of `glfwSetInputMode(GLFW_IME, true or false)`. We intend pass through `activate` or `deactivate` state to an input method context attached by system default, we don't intend to force disable input method.

For example we use [`ImmSetOpenStatus()`](https://learn.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immsetopenstatus) on Windows while [`ImmAssociateContext(hwnd, NULL)`](https://learn.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immassociatecontext) is used in the example you pointed out. If an application doesn't care it, an input method context will be attached by default when a window is created. We don't go against it as same as current GLFW.

We think force disabling IM isn't needed in most cases as most applications doesn't care it.
Because when an IM context is deactivated, it is expected to act almost as if it does not exist.
On the other hand, I know that such requirement (disable IM completely) also exists. So we can add another API to achieve it.
But it would be better to do it in another pull request to prevent this big change from growing further.

ashie:
@ghostflyby 

> The cocoa implementation is switching between ASCII and first system language input source instead of toggling text input, which may cause untended IME changes for multilingual IME users or when switching between windows. Is that intended?

Do you mean `_glfwSetIMEStatusCocoa()`?
https://github.com/glfw/glfw/pull/2130/files#diff-b37cd88f1337abc4ba24aaef1fffbbdcb1d7fbae309d299bed6a140ec5ee77c9R2012-R2069

Although we don't think current implementation of `_glfwSetIMEStatusCocoa()` is desired, probably macOS doesn't provide any other suitable way such as `ImmSetOpenStatus()` on Windows.

In addition, as I described in https://github.com/glfw/glfw/pull/2130#issuecomment-2591745482, we don't intend to disable IM completely in this pull request. Although we are not against adding such an API, we don't want to grow this pull request further.

ghostflyby:
Perhaps I have a better solution on macOS.


Instead of always calling `[self interpretKeyEvents:@[event]];` , store the imestatus set in a state,.

inside `- (void)keyDown:(NSEvent *)event`, check state and if enabled, call `[self interpretKeyEvents:@[event]]; `, if not, call insertText using [event.characters](https://developer.apple.com/documentation/appkit/nsevent/characters?language=objc) and selectedRange.

Then neither TIS nor ime switches is necessary.

idea tested with a simple swiftui view, and generally works.

```swift
import SwiftUI
import AppKit

struct ContentView: View {
    @State private var imeStatus: Bool = false

    var body: some View {
        VStack(spacing: 20) {
            Toggle("Enable IME", isOn: $imeStatus)
                .padding()

            CustomTextView(imeStatus: $imeStatus)
                .frame(height: 150)
                .border(Color.gray, width: 1)
                .padding()
        }
        .padding()
    }
}

struct CustomTextView: NSViewRepresentable {
    @Binding var imeStatus: Bool

    func makeNSView(context: Context) -> NSTextViewWrapper {
        let wrapper = NSTextViewWrapper()
        wrapper.imeStatus = imeStatus
        return wrapper
    }

    func updateNSView(_ nsView: NSTextViewWrapper, context: Context) {
        nsView.imeStatus = imeStatus
    }

    class NSTextViewWrapper: NSTextView {
        var imeStatus: Bool = false

        override func keyDown(with event: NSEvent) {
            if imeStatus{
                super.keyDown(with: event)
            } else {
                insertText(event.characters ?? "" ,replacementRange: self.selectedRange())
            }

            // Log to observe key events (can be replaced with more actions if needed)
            print("Key down: \(event.characters ?? "") | IME Status: \(imeStatus)")
        }
    }
}
#Preview {
    ContentView()
}
```

ashie:
@ghostflyby Thanks for your comment, sorry for my late response.

It seems that your example is bypassing IME.
As I described before, force bypassing IME isn't intended in the current implementation.
On the other hand, I know that bypassing IME is also required in some contexts like yours.

To resolve both requirements, it might be better to extend 3rd argument of `glfwSetInputMode(win, GLFW_IME, value)` like this:

* `GLFW_IME_DISABLE`: Force bypass IME. Should be supported on all platforms.
* `GLFW_IME_ENABLE`: Don't bypass IME. Should be supported on all platforms
* `GLFW_IME_OFF`: Turn off IME by platform's fashion. Might not be supported on some platforms. 
* `GLFW_IME_ON`: Turn on IME by platform's fashion. Might not be supported on some platforms.

We need to consider the detailed specifications a bit more though.

We are discussing about this topic also in https://github.com/clear-code/glfw/pull/5.

ghostflyby:
> @ghostflyby Thanks for your comment, sorry for my late response.
> 
> It seems that your example is bypassing IME. As I described before, force bypassing IME isn't intended in the current implementation. On the other hand, I know that bypassing IME is also required in some contexts like yours.
> 
> To resolve both requirements, it might be better to extend 3rd argument of `glfwSetInputMode(win, GLFW_IME, value)` like this:
> 
> * `GLFW_IME_DISABLE`: Force bypass IME. Should be supported on all platforms.
> * `GLFW_IME_ENABLE`: Don't bypass IME. Should be supported on all platforms
> * `GLFW_IME_OFF`: Turn off IME by platform's fashion. Might not be supported on some platforms.
> * `GLFW_IME_ON`: Turn on IME by platform's fashion. Might not be supported on some platforms.
> 
> We need to consider the detailed specifications a bit more though.
> 
> We are discussing about this topic also in [clear-code#5](https://github.com/clear-code/glfw/pull/5).


The cocoa implementation behavior is not aligning with other platforms. Text events and IME are different. On non-cocoa platforms there are such basic char events on IME disabled as fallback, where current implementations are actually bypassing IME and falling back to basic char events. However cocoa is currently switching between IMEs as it has no such basic char events.
My example is just imitating such a basic char event, which preserves ASCII text input with definitely no composition window, aligning with the behavior of other platforms.

VVD:
friendly ping

Houfeng:
ping

VVD:
Celebrating 3 years! 🎉🥳🎊

travelliu:
ping

Houfeng (approved):

jellychen (approved):

jellychen:
This question is currently stuck at what point? Isn’t this quite an important issue?

GADDQ:
it's really hard to imagine this important pr didn't get the enough attention......

GADDQ:
@dougbinks could u pls help to check this pr? thanks!

dougbinks:
I understand that this issue is important to you, but please do not ping the maintainers as it just uses up our time responding to you. We would rather spend the time we have on the project.

Merging PRs like this takes a huge amount of time. We've been discussing this PR and I have done some testing and code review but we are very far off being able to merge it.

GADDQ:
oh, sorry, my bad. I just want help this pr to be attention. thanks for your time to explain to me, won't be the next time🙏



---Original---
From: "Doug ***@***.***&gt;
Date: Wed, Dec 3, 2025 01:40 AM
To: ***@***.***&gt;;
Cc: "Earth ***@***.******@***.***&gt;;
Subject: Re: [glfw/glfw] Add IME support for each platform (PR #2130)


dougbinks left a comment (glfw/glfw#2130)
 
I understand that this issue is important to you, but please do not ping the maintainers as it just uses up our time responding to you. We would rather spend the time we have on the project.
 
Merging PRs like this takes a huge amount of time. We've been discussing this PR and I have done some testing and code review but we are very far off being able to merge it.

jellychen:
I’m no longer using GLFW — not because GLFW is bad, but because I have some personal requirements that differ from what it provides. I implemented my own abstraction layer for both Win32 and macOS. After revisiting this project, I realized that GLFW’s IME implementation actually isn’t very good.

dougbinks:
@jellychen 
> I realized that GLFW’s IME implementation actually isn’t very good.

I presume you mean the IME implementation in this PR? If you have useful feedback on the PR we'd love to hear it, but I totally understand if you don't have time given you're no longer using GLFW.

Tnze:
I think we should use the [TSF](https://learn.microsoft.com/en-us/windows/win32/tsf/text-services-framework) rather than IMM on Windows.

ashie:
> I think we should use the [TSF](https://learn.microsoft.com/en-us/windows/win32/tsf/text-services-framework) rather than IMM on Windows.

We agree that TSF is the modern and recommended IME framework on Windows, and we do not rule out adding TSF support in the future. In fact we did consider using TSF at an early stage of this work.
However, for this PR and the current stage of GLFW IME support, we intentionally choose IMM for the following technical and architectural reasons.

### 1. Dependency and architectural impact

TSF is not a drop-in replacement for IMM.
A correct TSF implementation requires following things, they introduce substantial new architecture compared to IMM, which integrates directly with the Win32 message loop already used by GLFW.

* COM initialization and lifetime management
* Interaction with ITfThreadMgr, document managers, and text stores
* Careful handling of apartment models and message pumping

### 2. Language and implementation mismatch (C vs C++)

GLFW is written in portable C. On the other hand introducing TSF would require force C++ into the Windows backend, or complex C-based COM wrappers that significantly increase code complexity. Neither is desirable for an initial IME implementation of GLFW.


### 3. Stability and ecosystem experience

Although TSF is the modern API, real-world experience shows that it is not trivial to get right:

* Large projects such as Chromium or Firefox have historically faced long-standing issues related to TSF behavior, focus handling, and IME edge cases
* Correct TSF integration requires extensive testing across different IMEs (MS IME, ATOK, Google IME, etc.), multiple Windows versions,  various focus and window activation scenarios and so on. In contrast, IMM is mature, stable, and well-understood, and still widely used for legacy and lightweight applications. For GLFW’s initial IME support, predictability and robustness are more important than API modernity.

### 4. Incremental approach and maintainability

This PR aims to introduce minimal, conservative IME support that:

* Integrates cleanly with existing Win32 message handling
* Has a small and auditable code footprint

Once this baseline is merged and proven stable, it becomes much easier to:

* Evaluate real-world IME usage feedback
* Define a clean abstraction boundary
* Consider adding TSF as an optional or alternative backend

Jumping directly to TSF would significantly raise the bar for review, testing, and long-term maintenance.

ashie:
@dougbinks

> Merging PRs like this takes a huge amount of time. We've been discussing this PR and I have done some testing and code review but we are very far off being able to merge it.

Thanks a lot for taking the time to review and test this PR — we appreciate your effort.
We understand that IME support can be difficult to evaluate, especially for contributors outside CJK environments, so we’re happy to be patient.
If there’s anything we can do to help move this forward (for example, splitting the PR or addressing specific concerns), please let us know.

Tnze:
@ashie That makes sense. Thanks for the explanation.

ghostflyby:
@ashie any chance for switching from TIS on macOS?
 TIS is neither intended for this purpose nor intuitive on actual behavior to users. The current implementation based on it is complex.
https://github.com/clear-code/glfw/pull/7

bczhc:
Tested this PR under Wayland. I'm able to input Chinese text using Fcitx5 into multiple glfw apps, although they're displayed as `?`s but I suppose it's just a font issue. However, in the game Minecraft, Fcitx5 candidate view is able to show (when Minecraft stays non-fullscreen) when I press the IME switch key (super+space for me), but I can't commit any text. Input something, hit space, the candidate view closes, and nothing is in Minecraft chat box.

Interestingly without this PR (just using glfw/glfw) I'm able to input text in Minecraft and other glfw apps under Xwayland (use `-DGLFW_BUILD_WAYLAND=OFF` compile option).

lilydjwg:
> Interestingly without this PR (just using glfw/glfw) I'm able to input text in Minecraft and other glfw apps under Xwayland (use `-DGLFW_BUILD_WAYLAND=OFF` compile option).

It is not interesting. You can unset `WAYLAND_DISPLAY` env variable to force use of Xwayland and XIM. This is why I still use Xwayland for Minecraft.

bczhc:
> You can unset `WAYLAND_DISPLAY` env variable to force use of Xwayland and XIM.

XD yeah that's a quicker way. All is good when i'm using X11 previously, and Xwayland is the way to go for Minecraft for now, plus patch #1972 for fixing the unexpected deletion issue - everything's perfect.

codebykevin:
Following this with great interest. I am looking at using GLFW as the windowing host for a Wayland-native Tcl/Tk port, which would be the fourth supported platform.

bczhc:
@lilydjwg I just managed to fix the IME input issue of Minecraft.

https://github.com/glfw/glfw/blob/3d89365fdb7449487993515156cf81011dc6121c/src/input.c#L321-L328

`window->callbacks.character` in `_glfwInputChar` is always null on my end but `charmods` is not, from some debugging; no matter whether I launch Minecraft under Wayland or through Xwayland.

Then, first, thanks to @ashie who implemented the Wayland v3 input protocol in this PR. In [function `textInputV3CommitString`](https://github.com/clear-code/glfw/blob/c0d34f3634157370fd63089122e4e732233d9d86/src/wl_window.c#L2273-L2277), Fcitx5 committed text is ready, however `window->callbacks.character` is still null and it basically prevents the text go into Minecraft. So the solution/workaround is quite simple: substitute `callbacks.charmods` for it.

Commit https://github.com/bczhc/glfw/commit/21b37a508f415628cb31f8bde463164e4fcd0462 does this fix and https://github.com/bczhc/glfw is the custom glfw I'm using now. One remaining issue is, fcitx5 is not positioned properly or just invisible when Minecraft is in fullscreen but it just doesn't matter too much for me.

Edit:

> fcitx5 is not positioned properly or just invisible when Minecraft is in fullscreen but it just doesn't matter too much for me.

I switched to niri later. On niri, fcitx5 will show up (at middle-left of the screen) when Minecraft is in fullscreen. Just perfect!

lilydjwg:
Thanks, your patch works for me. The window position is not good for windowed mode however (barely visible at the bottom) in my case, but in full screen it is actually better that the window covers the chat input area.

```
[1592955.139] {Default Queue} zwp_text_input_v3#57.preedit_string("啊", 0, 0)
[1592955.143] {Default Queue} zwp_text_input_v3#57.done(8)
[1592955.145] {Default Queue}  -> zwp_text_input_v3#57.set_cursor_rectangle(0, 1080, 0, 0)
[1592955.148] {Default Queue}  -> zwp_text_input_v3#57.commit()
```

My screen is 3840x2160@2x, so in windowed mode which has a lower height than my screen, the point at surface coordinate (0, 1080) is outside the surface.

lilydjwg:
> The window position is not good for windowed mode however (barely visible at the bottom) in my case

Well, I patched my Wayland compositor to make sure the window is fully visible no matter what the client wants.

rrtt217:
> @lilydjwg I just managed to fix the IME input issue of Minecraft.
> 
> https://github.com/glfw/glfw/blob/3d89365fdb7449487993515156cf81011dc6121c/src/input.c#L321-L328
> 
> `window->callbacks.character` in `_glfwInputChar` is always null on my end but `charmods` is not, from some debugging; no matter whether I launch Minecraft under Wayland or through Xwayland.
> 
> Then, first, thanks to @ashie who implemented the Wayland v3 input protocol in this PR. In [function `textInputV3CommitString`](https://github.com/clear-code/glfw/blob/c0d34f3634157370fd63089122e4e732233d9d86/src/wl_window.c#L2273-L2277), Fcitx5 committed text is ready, however `window->callbacks.character` is still null and it basically prevents the text go into Minecraft. So the solution/workaround is quite simple: substitute `callbacks.charmods` for it.
> 
> Commit [bczhc@21b37a5](https://github.com/bczhc/glfw/commit/21b37a508f415628cb31f8bde463164e4fcd0462) does this fix and https://github.com/bczhc/glfw is the custom glfw I'm using now. One remaining issue is, fcitx5 is not positioned properly or just invisible when Minecraft is in fullscreen but it just doesn't matter too much for me.

Minecraft itself used `callbacks.charmods` instead of `callbacks.character` before 26.1, which is incorrect and broke usage of keys w/ modifiers in text input like Ctrl+A/C/V. The proper fix is to make a Minecraft mod.

bczhc:
> Minecraft itself used `callbacks.charmods` instead of `callbacks.character` before 26.1, which is incorrect and broke usage of keys w/ modifiers in text input like Ctrl+A/C/V. The proper fix is to make a Minecraft mod.

Oki thanks for the info. So finally - Mojang fixed that.

rrtt217:
I should mention that 26.1 does not fix [IMBlocker#104](https://github.com/reserveword/IMBlocker/issues/104) or [MC-122477](https://bugs.mojang.com/browse/MC/issues/MC-122477) on X11/XWayland. Keys without modifiers still invoke both a character and a key event.
On Wayland 26.1 using `callbacks.character` can actually fix them.

bczhc:
> I should mention that 26.1 does not fix [IMBlocker#104](https://github.com/reserveword/IMBlocker/issues/104) or [MC-122477](https://bugs.mojang.com/browse/MC/issues/MC-122477) on X11/XWayland. Keys without modifiers still invoke both a character and a key event. On Wayland 26.1 using `callbacks.character` can actually fix them.

I mean this is an already-known old-old bug and it's an X11-specific bug: https://github.com/fcitx/fcitx5/issues/653#issuecomment-1943823050. On wayland IME in Minecraft is implemented using [`text-input-v3`](https://wayland.app/protocols/text-input-unstable-v3) and such "preedit-deletion causes commited text deleted" bug doesn't happen more.

I'm always using my custom patched glfw and this is the cherry-picked commit which fixes it: https://github.com/bczhc/glfw/commit/219c8e391caedac82f54628cfe63db1dcf9ee480. And for me - I prefer using patched libs rather than Minecraft mods.

Anyway I think the Minecraft stuff here we're talking about is a bit offtopic here because it has nothing to do with this PR per se and upstream glfw. We could move elsewhere and feel free to cc or ping me from other discussions if there's more MC stuff information!

ashie:
I've rebased this PR.
Past versions are preserved here:
https://github.com/clear-code/glfw/branches/all?query=im-support

Ryoya15:
Is it a bad idea to apply the changes from #1972, which haven't been merged even after so many years, to this branch?

If not, I'd like you to do it.

ashie:
I'd prefer not to include #1972 in this PR.

#1972 is really a standalone X11 event-filtering fix rather than an IME-specific change. It is certainly relevant to IME behavior on X11, and IME support works more correctly with that fix applied, but it also affects general key event processing and addresses its own issue (#1794).

Although the patch itself is small, the behavioral change is not entirely trivial. It changes when _glfwInputKey is generated relative to the filtered X11 events, which is exactly why additional investigation and event sequence analysis were added to the discussion on #1972.

Because of that, #1972 deserves to be reviewed and evaluated on its own merits. Combining it into this already large IME PR would broaden the review scope and mix two separate discussions together.

Therefore, I'd prefer to keep this PR focused on IME support and leave #1972 as an independent change.

ashie:
> @ghostflyby Thanks for your comment, sorry for my late response.
> 
> It seems that your example is bypassing IME. As I described before, force bypassing IME isn't intended in the current implementation. On the other hand, I know that bypassing IME is also required in some contexts like yours.
> 
> To resolve both requirements, it might be better to extend 3rd argument of `glfwSetInputMode(win, GLFW_IME, value)` like this:
> 
>     * `GLFW_IME_DISABLE`: Force bypass IME. Should be supported on all platforms.
> 
>     * `GLFW_IME_ENABLE`: Don't bypass IME. Should be supported on all platforms
> 
>     * `GLFW_IME_OFF`: Turn off IME by platform's fashion. Might not be supported on some platforms.
> 
>     * `GLFW_IME_ON`: Turn on IME by platform's fashion. Might not be supported on some platforms.
> 
> 
> We need to consider the detailed specifications a bit more though.
> 
> We are discussing about this topic also in [clear-code#5](https://github.com/clear-code/glfw/pull/5).

One additional point regarding the proposed API design:

My intention is to keep `GLFW_IME_DISABLE` as the default behavior in order to preserve existing application behavior and avoid introducing unintended changes to input handling for existing users. This reflects the current GLFW behavior and minimizes the risk of breaking applications that rely on it.

At the same time, `GLFW_IME_ENABLE` is intended as an explicit opt-in for applications that require IME-aware input handling.

In other words, IME-aware behavior is being added as an optional capability rather than changing the default semantics of GLFW.

This is still a design proposal and has not been implemented yet, so feedback on the API shape is welcome.

codebykevin:
I am implementing IME for Wayland outside of GLFW. I am unable to wait for this slowly-ripening feature to be ready.

ghostflyby:
@ashie 
I have not been talking about any behavioral changes, but an implementation detail for the current API surface and bug fixes.

The example and PR I made just tried to align the platform behaviors to the current one while make it simple and correct.

Using TIS to switch IME is causing several bugs, which are unfixable if still using the current approach.
1. switching between US keyboard and the first locale related IME, but expected to be between a ASCII input and last IME the user chose.
2. IME switch cursor indicator is showing itself automatically, which should never happen without manual toggle.

ashie:
Thanks, I understand the issues you are describing with the current TIS-based approach on Cocoa.

I agree that there are real UX problems in the current implementation, especially around IME switching behavior and system UI indications.

However, I would like to separate these implementation-level fixes from the current API design discussion in this PR. The goal here is to first settle the API semantics, and then revisit platform-specific implementation improvements accordingly.

ashie:
FYI: Relevant background discussions can also be found in:
https://github.com/clear-code/glfw/pull/5
https://github.com/clear-code/glfw/pull/7

ghostflyby:
The description **Force bypass IME.** is misleading.
```
* `GLFW_IME_DISABLE`: Force bypass IME. Should be supported on all platforms.
```

It seems to actually be `pre-IME GLFW behaviour` aka `forced ASCII text input`, so API description and documentation might need changes.

ashie:
Thanks, I see the point. I'll revise the API description to make it clearer.

ashie:
After further consideration, I have changed my mind about the API design.

I no longer think that `GLFW_IME_ENABLE` / `GLFW_IME_DISABLE` is the right abstraction for this problem.

Instead, I think GLFW needs an explicit distinction between **window focus** and **text input focus**.

Therefore, my current plan is to introduce an API as following instead of `GLFW_IME_ENABLE` / `GLFW_IME_DISABLE`:

```c
glfwSetTextInputFocus(window, GLFW_TRUE);
glfwSetTextInputFocus(window, GLFW_FALSE);
```

The purpose of this API is to tell GLFW when the application enters or leaves a text input context.

The reason is that many modern applications, including games and browsers, render their own UI inside a single native window. In such applications, native window focus and text input focus are not the same thing.

For example, a game window may remain focused while the user is controlling the game world, navigating menus, chatting, renaming items, or editing text. Only some of those situations actually require text input.

After revisiting the feedback received in this discussion, I now think that most complaints are not really about IME state management itself.  The common problem is that the platform input method still considers the application to be in a text input context when the application has already switched back to gameplay or some other non-text-input mode.

Traditional GUI applications usually get this distinction "for free" because text widgets have their own native focus handling. Games and other single-window applications generally do not.

I still think IME state control (`ON` / `OFF`) may be useful as a separate, platform-dependent capability, but I no longer think it should be the primary mechanism used to solve this problem.

To me, explicit text input focus appears to be a more fundamental abstraction that maps naturally to all platforms and application types.
One consequence of this approach is that applications will need to explicitly notify GLFW when entering or leaving a text input context.

This means that existing applications may need to add a few calls at places such as chat boxes, text editors, search fields, or rename dialogs. However, I think this is unavoidable because only the application knows when text input is actually expected.

ashie:
As a side experiment, I built an alternative X11 IME backend using IBus on top of the API proposed here:

https://github.com/clear-code/glfw/pull/10

Not part of #2130, just a design validation exercise.

ashie:
I've pushed a prototype implementation based on the text-input-focus approach mentioned above:

https://github.com/clear-code/glfw/tree/im-support-text-input-focus

The main change is that I no longer think IME enable/disable is the right abstraction for this problem. Instead, the branch introduces explicit text input focus management via `glfwSetTextInputFocus()`.

The goal is to distinguish between native window focus and application text input focus, which I now believe is the more fundamental issue for games and other applications that render their own UI inside a single native window.

After some testing and cleanup, I plan to merge this into the current `im-support` branch.

ashie:
The text-input-focus work mentioned above is now complete.
It has been split into reviewable commits, tested on the available platforms and documented.

I've add following public API:

```c
glfwSetTextInputFocus(window, GLFW_TRUE);
glfwSetTextInputFocus(window, GLFW_FALSE);
```

This API represents application text input focus, not IME on/off state.
It is meant for applications such as games and browsers that render their own UI inside a single native window, where native window focus and text input focus are separate concepts.

- `GLFW_TRUE` means the application has entered a text input context, such as a chat box, text field, search box, rename dialog or editor.
- `GLFW_FALSE` means the application has left that text input context and returned to gameplay, navigation or another non-text-input mode.

Applications that never call `glfwSetTextInputFocus` keep the previous platform behavior. Once the function is called for a window, that window uses explicit text input focus management.

This is related to, but distinct from, `glfwSetInputMode(window, GLFW_IME, ...)`. `glfwSetTextInputFocus` describes application intent, while `GLFW_IME` remains platform-specific IME state control. In practice, many cases that look like "turn IME on/off" are better described as entering or leaving text input focus.

Current platform behavior:

- X11 maps text input focus to XIM input context focus with `XSetICFocus` and `XUnsetICFocus`.
- Win32 gates GLFW's own IMM message handling on text input focus and temporarily detaches the HIMC while focus is out, because gating alone did not prevent native candidate UI on Windows.
- Cocoa skips `interpretKeyEvents` while explicit text input focus is inactive, preserving normal key input and avoiding TIS input source switching.
- Wayland maps text input focus to text-input-v1/v3 activation where available and ignores stale callbacks while disabled.

On focus out, platforms also reset or cancel active preedit/composition state where there is a safe existing mechanism.

The `input_text` test now has a text input focus toggle with three states: compatible, Focus In and Focus Out.

Testing so far:

- Win32 native: works as expected
- Wine: candidate UI may still appear while focus is out; this appears to be a Wine's limitation
- Wayland text-input-v1: text input focus behavior works; candidate position tracking may still be unreliable, but this also happens on the existing IME support branch and does not appear to be caused by this change

This work follows the design discussions and prior work in https://github.com/clear-code/glfw/pull/5 and https://github.com/clear-code/glfw/pull/7. The Cocoa implementation also builds on the work provided by ghostflyby in clear-code/glfw#7.

ashie:
I've found a bug in preedit cursor handling in Wayland backend.
It has been fixed in 9af719e6073a317f2dfa17f6cf0c373619b1ec8c.

ashie:
GNU/Linux IME users may find this interesting: there is now an experimental follow-up branch/PR that makes this API usable with Minecraft via LWJGL.

https://github.com/clear-code/glfw/pull/10#issuecomment-4970040326

It includes an X11 IBus backend prototype and a compatibility mode for the text-input-focus model. This is not part of the upstream proposal here yet, but it is a real-world validation experiment for the API design.

In early testing, Japanese input works on Linux without depending only on XIM, including preedit, commit and candidate window positioning. The text-input-focus model also avoids the common problem where an active IME keeps intercepting gameplay keys after leaving a text field.

ashie:
> Is it a bad idea to apply the changes from #1972, which haven't been merged even after so many years, to this branch?
> 
> If not, I'd like you to do it.

@Ryoya15  Could you try this? https://github.com/clear-code/glfw/pull/10#issuecomment-4986343612
It's the all-in-one branch for improving various IM related issues.
Probably it will fix most of issues that you mentioned in [Minecraft Wiki](https://ja.minecraft.wiki/w/%E3%83%81%E3%83%A5%E3%83%BC%E3%83%88%E3%83%AA%E3%82%A2%E3%83%AB:%E6%97%A5%E6%9C%AC%E8%AA%9E%E5%85%A5%E5%8A%9B).

Ryoya15:
Unfortunately, it never worked on my system. Pressing the Hankaku/Zenkaku key doesn't activate the IME.
I thought it might not be compatible with Fcitx5, so I did a clean install of Ubuntu 26.04 to test it, but it still didn't work.

ashie:
> Unfortunately, it never worked on my system. Pressing the Hankaku/Zenkaku key doesn't activate the IME. I thought it might not be compatible with Fcitx5, so I did a clean install of Ubuntu 26.04 to test it, but it still didn't work.

@Ryoya15  Thanks for giving it a try!
Let's continue the discussion on this matter (IBus, Fcitx5, Minecraft on Linux) at https://github.com/clear-code/glfw/pull/10
I would like to focus the discussion here on basic IM support and offload other topics to a different place.

PR #2151, opened 2022-07-09
Fix MoltenVK Loading on macOS

GameParrot:
Loading Vulkan fails on macOS even if the user has libMoltenVK.dylib. This is because it tries to load libvulkan.1.dylib, not libMoltenVK.dylib.


charles-lunarg:
This PR definitely shouldn't be taken as is. Not all applications use moltenVK directly, and they sometimes change what they use based on whether its a development build (where loader + validation layers are useful) or a release build (where nothing else should be needed).

elmindreda:
@charles-lunarg Absolutely agree!

PR #2152, opened 2022-07-13
Resolved #987 - Add glfwDragWindow function to latest platform structure

envyen:
Resolved a nice work by @felselva #987 to work on latest platform structure

> Add a function `glfwDragWindow` that starts a drag operation for the specified window. The implementation is done for X11, Windows, and requires tests on Wayland. Patch is related with the issue #923.

[example_windows_drag.webm](https://github.com/glfw/glfw/assets/1267337/c44d9ce6-42d0-4f19-b152-248d1a596d86)

@elmindreda @felselva


Seneral:
Just tested it on X11 and wayland:

On X11, I first made an error (first time using GLFWs inputs directly and not through Dear ImGui), which resulted in a VERY bad experience for me. Essentially the window stuck to my cursor at random offsets, dragging it along with it erratically (with the mouse cursor being way out of the window area even), and I couldn't do anything to close the window or do anything really, so I had to go into a separate session to kill my window. That was because I did a simple polling implementation (not using a callback) so I called it repeatedly - bad idea. This WILL happen to some devs, and it's bad enough that I would consider some preventative measures.
After I fixed it, it worked wonderfully.

Now on wayland, it didn't work at all. Luckily a quick search lead me on the right path, it needed the serial used in the mouse button event, which is _glfw.wl.serial, not _glfw.wl.pointerEnterSerial. Changing that, it worked fine, HOWEVER, stopping the drag never cleared the mouse button state, so in contrast to X11, on wayland it still thought the mouse button was pressed (glfwGetMouseButton(window, GLFW_MOUSE_BUTTON_LEFT) == GLFW_PRESS) until the next left mouse button press (and needed that, no right click, no key press, nothing else clears that state).

Haven't tested on windows yet, will do in the future.
Thanks for your work on this, hope it gets fixed up and merged soon.

src/wl_window.c
```diff
@@ -2360,6 +2360,11 @@ void _glfwFocusWindowWayland(_GLFWwindow* window)
                     "Wayland: The platform does not support setting the input focus");
 }
 
+void _glfwDragWindowWayland(_GLFWwindow* window)
+{
+    xdg_toplevel_move(window->wl.xdg.toplevel, _glfw.wl.seat, _glfw.wl.pointerEnterSerial);
```
Seneral, line 2365: Needs `_glfw.wl.serial` instead of `_glfw.wl.pointerEnterSerial`, else it does nothing.

Seneral (changes requested):

Seneral:
Seems related to https://gitlab.gnome.org/GNOME/mutter/-/issues/2523
Except... I'm on KDE Plasma, so I'm not using mutter.

Seneral:
So unfortunately the suggestions that the mouse leave event should silently set the mouse state to "undefined" with no mouse button release event, is idealistic. E.g. I'm using Dear ImGui and it's input abstraction, which with the GLFW backend, uses the GLFW event subsystem exclusively, without polling the button state, but keeping it's own state with GLFW events.
So it would not notice that the mouse button state changed and internally still thinks it's a drag - disabling things like button hovering feedback, etc.
It needs a GLFW mouse button release event to not become inconsistent. Now we could change the Dear ImGui GLFW backend to support wayland, OR we could be realistic and realise that there are many other abstractions/user code doing the exact same, and simply generate a mouse button release event on mouse leave (and the same for key leave).
This goes against the intention of the wayland protocol, but it seems to work, though I would expect there to be edge cases where it doesn't work. Though looking at [above linked issue](https://gitlab.freedesktop.org/wayland/wayland/-/issues/280), the wayland recommended approach might cause edge case issues anyway.

So I suggest we apply diff_realistic and merge this as soon as maybe we get another confirmation that windows works (and I guess a different wayland composer than KWin, might do that myself).

[diff_intended.txt](https://github.com/glfw/glfw/files/13219719/diff_intended.txt)
[diff_realistic.txt](https://github.com/glfw/glfw/files/13219720/diff_realistic.txt)

Seneral:
I tested it on windows, and it worked fine, but had the same issue that event-based systems got stuck in a state - there never was a mouse button up event for the header drag. In contrast to wayland, I found no documentation or event that tells the app to discard the state either, so I just send a button down event right when the window starts being dragged. That is not optimal and prevents any application from knowing if it is still being dragged, but I have not found a better way. And it does work.
Unfortunately on windows the docking to monitor sides/corners doesn't work automatically, so the lack of easy resizing the window is very apparent, so I disabled the custom header on windows. On X11+wayland, although my windows can't be resized yet, they can be docked, and it's usable enough that for now, I am enabling the custom header.
One big downside to any custom header is that when the program is freezing / being debugged, the header buttons don't work anymore, which means you can't close the window that way.
[diff_win_realistic.txt](https://github.com/glfw/glfw/files/13338374/diff_win_realistic.txt)

Seneral:
Here's my program using GLFW+Dear ImGui for a WIP custom header (no custom icons yet). This branch made the custom header possible, so I suggest merging it.
image

envyen:
Hi @Seneral 

A huge thanks to your review this in detail. 

This PR was made sometime back. I was in need for drag feature on my linux x11 ubuntu Imgui app, and didnt have wayland to try then, Let me go through your comments and I will integrate the suggested changes and update.

This is a nice function i still use, I hope this gets merged soon

Seneral:
@envyen Most ist just reasoning why I made the changes for the maintainers, and most importantly, what potential side effects they have. But for the purposes of the feature, applying the diff like that should be fine.

Did you implement custom resizing as well? Any hints on existing cross-platform solutions like this for that in glfw?

Seneral:
@envyen any looking at the changes? Trying to get this merged soon (hopefully)

envyen:
@Seneral, The changes are sensible and I have added the diff
Regrettably, I couldn't update this in time, and couldn't be merged before the 3.3.9 release.

src/wl_window.c
envyen: Added the suggestion

Seneral:
All good, maybe next release. But thank you for keeping this updated, I aim to have a resize-focused PR as well.
Though, it's still missing one line for the button release behaviour on windows (added that later, in the last diff I posted).

Apart from that (and more platforms than "just" windows, X11, wayland), the only thing missing here is, afaik, support for snapping the window to the sides on windows.

envyen:
Updated

> Though, it's still missing one line for the button release behaviour on windows (added that later, in the last diff I posted).

p0ryae:
@envyen I gave your PR another test. It works perfectly across every platform I tried except native Wayland. It does behave correctly under XWayland, as expected. From reading through the discussion here, it looks like you addressed at least one Wayland-specific issue already, but the underlying problem still seems to persist on full Wayland sessions. 

I can provide more logs if necessary, but the crash seems to stem from window->wl.xdg.toplevel being null, which leads to segfault.

**Update:** I decided to figure out a patch by myself. The issue was that `glfwDragWindow()` on Wayland only checked `window->wl.xdg.toplevel` which is NULL when libdecor (server-side decorations) is used, so it caused it to always fail and segfault. My fix gets the `xdg_toplevel` from the libdecor frame using `libdecor_frame_get_xdg_toplevel()` when libdecor is active. While I was at it, I also decided to add some additional null checks and validation for the seat and serial to prevent potential edge cases. Here is the reference code: 

https://github.com/p0ryae/glfw/blob/5eacea0c273a551c3c6d4f1930707f95904600c4/src/wl_window.c#L2379C1-L2414C2

```cpp

void _glfwDragWindowWayland(_GLFWwindow* window)
{
    struct xdg_toplevel* toplevel = NULL;
    
    if (window->wl.libdecor.frame)
    {
        toplevel = libdecor_frame_get_xdg_toplevel(window->wl.libdecor.frame);
    }
    else
    {
        toplevel = window->wl.xdg.toplevel;
    }
    
    if (!toplevel)
    {
        _glfwInputError(GLFW_PLATFORM_ERROR,
                        "Wayland: Cannot drag window: xdg_toplevel not created yet");
        return;
    }
    
    if (!_glfw.wl.seat)
    {
        _glfwInputError(GLFW_PLATFORM_ERROR,
                        "Wayland: Cannot drag window: no valid seat");
        return;
    }
    
    if (_glfw.wl.serial == 0)
    {
        _glfwInputError(GLFW_PLATFORM_ERROR,
                        "Wayland: Cannot drag window: no valid serial");
        return;
    }

    xdg_toplevel_move(toplevel, _glfw.wl.seat, _glfw.wl.serial);
}
```

envyen:
@p0ryae 
Thanks a lot for testing this, and I really appreciate the debugging and the fix you made.
It’s great to have a fresh set of eyes on this after the PR has been idle for a long time.

I tested your fixed `_glfwDragWindowWayland` and it works fine on my native Wayland build 👍
If you’d like to contribute your fixes directly to this PR, please feel free to do so. Let me know if you’d like to be added as a collaborator on the PR branch [envyen:drag-window-dev ](https://github.com/envyen/glfw-dev/tree/drag-window-dev)

Many times the PR had been updated to keep up with the upstream, but missed the merge for releases, 
This PR has not been updated for a long time after the 3.3.9 release. I will try pull in the latest and resolve conflicts. 

May be this a good time with your fix on wayland as well.  
I’d love to update and keep improving this useful feature, and hopefully get this merged soon.

p0ryae:
Thanks again for getting back to me and for resolving the conflicts!

I’ve accepted your invitation to the repository and pushed my changes, so they should now be reflected in the PR.

I’m actively using this PR in my own cross-platform project, so I’ll do my best to help maintain it and continue improving the Wayland path and any other edge cases that come up. Looking forward to getting this feature into a shape that’s ready for upstream (albeit it may not be soon: https://github.com/glfw/glfw/issues/2266#issuecomment-3588547208).

Seneral:
Thanks, glad to see activity here!
I've been using this glfwDragWindow + glfwResizeWindow + Additional Cursors + A small memory leak fix for a long time now under native wayland, but without libdecor, nice it works with libdecor now as well.
There really is a lot to fix for proper Wayland support, and while I haven't rebased on these latest fixes here yet, perhaps the remaining commits will be useful to somebody:
https://github.com/Seneral/glfw
Namely proper cursor support since the overlap of cursors that are supported across X11, XWayland, and libwayland is honestly abysmal, with e.g. libwayland not supporting the more general Up/Down, but only Up or Down - so I added these specific variants for your application to use, and proper fallbacks for all platforms.

p0ryae:
@envyen this is awesome. I'll give it a try. 

Although for your impl of `glfwResizeWindow`, I think it's for the best if it's in a separate PR. 

For this PR, I’ll integrate the existing work for the rest and include you as a Co-Author (unless you'd prefer to be added as a contributor to the PR's repo).

PR #2191, opened 2022-09-22
Use wglGetPixelFormatAttribivARB to get pixel format count

taniwha:
DescribePixelFormat can return a very different number of pixel formats than does wglGetPixelFormatAttribivARB: 220 vs 670 on my system (Nvidia GTX 965 M). Being different in that direction is not a problem, but if DescribePixelFormat returns more formats than
wglGetPixelFormatAttribivARB, the loop will fail and terminate the context creation.

### DIFF
--- a/src/wgl_context.c
+++ b/src/wgl_context.c
@@ -72,10 +72,20 @@ static int choosePixelFormatWGL(_GLFWwindow* window,
     int attribs[40];
     int values[sizeof(attribs) / sizeof(attribs[0])];
 
-    nativeCount = DescribePixelFormat(window->context.wgl.dc,
-                                      1,
-                                      sizeof(PIXELFORMATDESCRIPTOR),
-                                      NULL);
+    if (_glfw.wgl.ARB_pixel_format)
+    {
+        int attrib[] = { WGL_NUMBER_PIXEL_FORMATS_ARB };
+
+        wglGetPixelFormatAttribivARB(window->context.wgl.dc, 1, 0,
+                                     1, attrib, &nativeCount);
+    }
+    else
+    {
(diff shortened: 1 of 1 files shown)


taniwha:
I haven't been able to verify it was the actual cause, but I used glfw for a little project for a client and the client got `WGL: Failed to retrieve pixel format attributes` on his machine (macbook running Windows on Parallela (so I guess something like Wine)).

As the call to wglGetPixelFormatAttribivARB failing is the only way that string can get printed, and it's in a loop controlled by nativeCount, the count being too large is the only thing that made sense since he could run other people's OpenGL apps.

NateIsStalling:
Hi @elmindreda, is there any recommended workaround for this issue or is any additional information or assistance needed?  
I'm also encountering the `WGL: Failed to retrieve pixel format attributes` error on my machine (macbook running Windows on Parallels VM) running the sample program from the [Getting started](https://www.glfw.org/docs/3.0/quick.html) guide using the 3.3.9 December 13, 2023 release, 64-bit Windows binaries, on a Windows 11 VM.

The 3.3.8 WIN64 binaries appear to work as expected with the same configuration, but testing a LWJGL project using LWJGL 3.3.2 build 13 binaries fails with the same `WGL: Failed to retrieve pixel format attributes` error - LWJGL 3.3.2 April 1, 2023, release notes include `glfw: Updated to latest 3.3.8 (up from 3.3.6)` - the previous LWJGL 3.3.1 version appears to work as expected and does not cause a glfw error.

elmindreda:
This issue should be fixed now with 00e86d4b733103a23278fe53ce58a0d14dd47d32. Thank you both! Please let me know if you would like your credit line changed.

PR #2192, opened 2022-09-29
Fix Joysticks On Linux

TheBrokenRail:
Joystick support isn't initialized unless ``_GLFW_LINUX_JOYSTICK`` is defined, and this is never defined. I replaced the checks with checks for ``GLFW_BUILD_LINUX_JOYSTICK``, which is actually defined.

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -394,7 +394,7 @@ GLFWbool _glfwConnectWayland(int platformID, _GLFWplatform* platform)
         _glfwGetKeyScancodeWayland,
         _glfwSetClipboardStringWayland,
         _glfwGetClipboardStringWayland,
-#if defined(_GLFW_LINUX_JOYSTICK)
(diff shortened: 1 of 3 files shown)


TheComputerGuy96:
@TheBrokenRail I actually noticed this issue before you when debugging the lack of controller detection on Minecraft with the MidnightControls mod, but didn't PR it soon enough :frog:

TheComputerGuy96 (approved):
This makes controllers appear in the joysticks test on GLFW, but there are unwanted things appearing as joysticks and mapping detection code is incomplete so my Xbox One S controller appears with no mappings set

I'll probably address these issues in future PRs, but anyway this change is good enough :frog:

TheBrokenRail:
Any update on this or my other PR?

elmindreda:
Thank you so much for this fix!

Issue #1904, opened 2021-05-24
- glfw refuses to go fullscreen with custom resolution refresh rate


Kerai:
I'm using GLFW through LWJGL on Windows 10, `glfwGetVersion` reports version 3.4.0

1920x1080@75hz is my current desktop resolution, it was set with NVidia control panel. It is so called "custom resolution"

`glfwGetVideoMode(glfwGetPrimaryMonitor())` correctly reports 1920x1080@75 as my current video mode

`glfwGetVideoModes(glfwGetPrimaryMonitor())` only lists "default" video modes reported by monitor, doesn't list my current custom desktop resolution.

`glfwSetWindowMonitor(windowHandle, glfwGetPrimaryMonitor(), 0, 0, 1920, 1080, 75)` - ignorez 75hz parameter, incorrectly sets resolution to 1920x1080@60

GLFW correctly recognizes this resolution as current: ````glfwGetVideoMode```` returns 

glfwGetVideoModes doesn't list my current resolution and glfwSetWindowMonitor

elmindreda:
It may be that your custom resolution is failing the display mode test that Windows provides (`CDS_TEST` in `win32_monitor.c`). You can, at your own peril, remove that test from your copy of GLFW and see if that fixes it.

PR #2209, opened 2022-10-20
Win32: Always return the current video mode

comp500:
Always returns the currently used video mode from _glfwGetVideoModesWin32. This handles the existing edge case (where there are no valid video modes) as well as custom resolutions added in the driver that might not necessarily be supported by the display; see https://github.com/glfw/glfw/issues/1904 and https://github.com/comp500/BorderlessMining/issues/26.

These custom resolutions don't show up in EnumDisplaySettings (unless you pass EDS_RAWMODE which can return invalid resolutions - could be an opt-in flag but shouldn't be default functionality) but retrieving them as the current video mode should allow a custom resolution to be used if it is already set; particularly for applications that only allow resolutions returned in the video mode list. There should be no need to use CDS_TEST as the mode is already current.

Fixes https://github.com/glfw/glfw/issues/1904.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -260,6 +260,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - comp500
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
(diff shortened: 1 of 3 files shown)


icedev-jp:
Why hasn't this been merged already? Any problems?

Issue #212, opened 2014-01-12
Mechanism for identifying a Rift


jherico:
The current GLFW monitors api, at least on Windows and OSX really identifies 'screens' rather than monitor (working from the terminology that a screen represents a virtual desktop area, while a monitor identifies a physical display device).  

This poses problems for developers attempting to target the Oculus Rift, particularly if they're using cloned displays, as the top level monitor only identifies itself by the name of "Generic PnP Monitor".  

Potential solutions could involve changing the behavior to include both virtual screens and actual monitors, or providing a parallel API for physical devices in addition to the current API for screens.  Alternatively, providing an API to access the EDID information for each monitor might prove useful.  

I'll look into the code and see if there's an easy way to adapt for the first approach, listing all monitors, active, inactive, cloned or otherwise.

elmindreda:
@jherico Have you had any success making OpenGL rendered direct HMD mode output independent of the default framebuffer size, with or without GLFW? The tip from the Oculus forums to call `ovrHmd_AttachToWindow` before context creation seems to make no difference.

jherico:
No I haven't.  It seems that while the Direct3D window size isn't related
to the Rift resolution, the OpenGL window you create must be the same
resolution as the Rift (or rather, if you scale it down, the resulting
image quality on the Rift suffers).  I've also found that unlike the
Direct3D window, which can be hidden, the OpenGL window has to, at least
initially, be visible on the screen.  It seems that after a bit you can
hide it and still have the Rift work, but I'm not clear on the delay
involved.

I suspect that because more of the OpenGL implementation is baked into the
vendor drivers with GL than with D3D that there are fewer touchpoints
available into which Oculus can inject their code.  Hopefully the big
vendors will get on board and start providing some decent VR extensions in
OpenGL soon.

PR #2211, opened 2022-10-29
Cocoa: Ignore monitors which are mirroring another

bear24rw:
Currently monitors which are mirroring another one are returned by `glfwGetMonitors` but their `ns.screen` is `NULL`. Some functions like `glfwGetMonitorContentScale` and `glfwGetMonitorWorkarea` assert that `ns.screen` is not `NULL`. Since mirrored monitors are logically the same we should only consider the primary monitor and ignore the non-primary ones.

The alternative to this would be to make the mirrored monitor have the same `ns.screen` as the primary.


bear24rw:
> The alternative to this would be to make the mirrored monitor have the same ns.screen as the primary.

My main concern with doing it this way is that it will confuse [imgui's multi viewport support](https://github.com/ocornut/imgui/docking) and any other software that assumes each `GLFWmonitor` has a unique screen.

Also, the caller has no way of knowing that the monitors `ns.screen` is `null`. If they try to call `glfwGetMonitorWorkarea` they will get the `_glfwInputError` but if they don't abort we continue down the function and use the null `ns.screen` object.

It seems simpler to consider mirrored displays as a special situation that glfw does not care about and only have glfw care about the primary displays.

ws909:
Are there any advantages to also reporting mirrored screens? In many cases, their properties will differ from the screen that's the source of the content. This would require two different `ns.screen`s (not that I read the source code). As long as the image of the source screen is being mirrored, is there ever any advantage to knowing the properties of the target? Or otherwise being able to tell that there is a screen being mirrored?

bear24rw:
> In many cases, their properties will differ from the screen that's the source of the content.

The monitor properties like physical relative position and size yet, but in macOS it seems like there is a single `NSScreen` object which is the same resolution and workarea between them.

I'm not sure about other platforms, maybe there are some where you can have different resolutions on each of the mirrored monitors?

> This would require two different ns.screens (not that I read the source code). 

`ns.screens` is the underlying`NSScreen` object provided by `macOS`, I'm not sure you can create a second one manually.

> As long as the image of the source screen is being mirrored, is there ever any advantage to knowing the properties of the target? Or otherwise being able to tell that there is a screen being mirrored?

I could see some applications wanting to know which monitors are being mirrored (I can't think of a specific application but I could imagine someone wanting it). I think this would require a new public api and I have no idea what the cross platform monitor mirroring situation looks like.

bear24rw:
Looking through the commit history this commit might be relevant: https://github.com/glfw/glfw/commit/1b6d8a6ed6375ec37ebc6f209fe1a75d689ec908

@ elmindreda do you know if there is any risk to ignore mirrored monitors?

bear24rw:
> > As long as the image of the source screen is being mirrored, is there ever any advantage to knowing the properties of the target? Or otherwise being able to tell that there is a screen being mirrored?
> 
> I could see some applications wanting to know which monitors are being mirrored (I can't think of a specific example but I could imagine someone wanting it). I think this would require a new public api and I have no idea what the cross platform monitor mirroring situation looks like.

@elmindreda suggests adding `glfwGetMirroredMonitor ` in https://github.com/glfw/glfw/issues/212#issuecomment-56364845

ws909:
> `ns.screens` is the underlying `NSScreen` object provided by `macOS`, I'm not sure you can create a second one manually.

Quartz Display Services can be used:
[CGGetOnlineDisplayList](https://developer.apple.com/documentation/coregraphics/1454964-cggetonlinedisplaylist)

The _Discussion_  section for the documentation of that function mentions an important use case for distinguishing between the source and target screen of a mirroring process, and managing both explicitly.

If the framebuffer hardware is connected, a display is considered connected or online.
When hardware mirroring is used, a display can be online but not active or drawable. Programs that manipulate display settings (such as gamma tables) need access to all displays, including hardware mirrors, which are not drawable.

ws909:
> Some functions like `glfwGetMonitorContentScale` and `glfwGetMonitorWorkarea` assert that `ns.screen` is not `NULL`. 

`glfwGetMonitorContentScale` can be rewritten to:

```objective-c
void _glfwGetMonitorContentScaleCocoa(_GLFWmonitor* monitor,
                                      float* xscale, float* yscale)
{
    @autoreleasepool {
        
    const CGDisplayModeRef mode = CGDisplayCopyDisplayMode(monitor->ns.displayID);

    if (xscale)
        *xscale = (float) (CGDisplayModeGetPixelWidth(mode) / CGDisplayModeGetWidth(mode));
    if (yscale)
        *yscale = (float) (CGDisplayModeGetPixelHeight(mode) / CGDisplayModeGetHeight(mode));
        
    CGDisplayModeRelease(mode);

    } // autoreleasepool
}
```


For `glfwGetMonitorWorkarea`, `CGDisplayIsInMirrorSet` could potentially be used in conjunction with other methods to discover the mirroring source screen, get its NSScreen object, and then use that for the work area.

PR #2215, opened 2022-11-06
implement support for wp-fractional-scale-v1

Zamundaaa:
For details see https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/149, but effectively the protocol allows compositor and client to talk in pixels instead of logical coordinates, allowing games to run at native resolution.


src/wl_window.c
```diff
@@ -407,6 +408,32 @@ static const struct wl_surface_listener surfaceListener =
...
+
+void handleScaleFactor(void *userData,
+                       struct wp_fractional_scale_v1 *wp_fractional_scale_v1,
+                       uint32_t scale_8_24)
+{
+    _GLFWwindow* window = userData;
+
+    const double oldScale = window->wl.compositorToClientScale;
+    const double newScale = scale_8_24 / (double)(1UL << 24);
+    printf("scale changed! %f", newScale);
```
vyivel, line 420: Left from debugging?

Zamundaaa: yep

vyivel:
Fallback decorations use pre-scaled `window->wl.{width,height}` for sizes, but don't send the factor itself.

image

Zamundaaa:
Fallback decorations should now be correct

theofficialgman:
any updates on this?

Zamundaaa:
Due to time constraints I've paused work on the protocol and implementations for now. It'll probably be a few months before I continue working on it.

theofficialgman:
@Zamundaaa `set_scale_factor` is NOT part of the v1 protocol that was merged and implemented

it is part of the v2 protocol that as of now is unmerged and not implemented anywhere:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/149

Zamundaaa:
I know, I'm the author of the now-v2 protocol. You're right though, the title of this needs updating

theofficialgman:
maybe I am missing something but isn't the fractional scaling protocol supposed to allow for rendering at the native monitor resolution?

with kwin wayland 5.27 (fractional scaling v1 implemented) and the decoration fallbacks from this PR removed (so it can buid and be used), my scaling factor is 1.5, and my native monitor resolution is 3840x2160, but I am rendering at 2560x1440 with this PR

theofficialgman:
I am also here to report that cursor is broken in minecraft with this PR applied to master. works correctly on wayland (building with wayland, no x11) without this PR

Zamundaaa:
This PR is not compatible with the v1 protocol, it can cause random memory accesses and weird stuff. It's really not ready for testing until I pick up work on the v2 protocol again and update this PR for it

Zamundaaa:
thanks. I would prefer to rebase it myself once I continue working on it, but I'll ask for help with that if needed

elmindreda:
Hello again, I'm about to implement support for `fractional-scale-v1`. Is `fractional-scale-v2` still under development and if so is there anything in particular I should be mindful of when implementing support for v1?

Zamundaaa:
Yes. After we found some fun problems coming down to the fractional-scale-v1 approach in recent KWin development, I'm picking up v2 again. For implementing v1, as long as you keep the v1 protocol optional there shouldn't be any additional hurdles for implementing v2.

To potentially save you some time with v1, if you're testing the implementation on Gnome, if you see gaps and/or overlaps between the window content and the decoration, that is sadly expected with some scale factor + window size combinations. v1 is incompatible with subsurfaces and thus libdecor's approach to client side decorations.

Issue #1699, opened 2020-05-24
Feature request: glfwGetCurrentMonitor


nornagon:
macOS and Windows both have some concept of which monitor is "current" when multiple displays are attached.

macOS: `[NSScreen main]` returns the monitor which contains the window which currently has keyboard focus.
Windows: `MonitorFromWindow(GetForegroundWindow())` does the same.

It would be great to be able to get this information through glfw to support the use case of opening a window on the "correct" display.

DerekCresswell:
Do you mean [glfwGetPrimaryMonitor](https://www.glfw.org/docs/3.3/group__monitor.html#ga721867d84c6d18d6790d64d2847ca0b1)?

nornagon:
That function returns the first monitor in the monitor list, which is not the same as the monitor which the user is currently using.

elmindreda:
> It would be great to be able to get this information through glfw to support the use case of opening a window on the "correct" display.

Do you mean windowed mode windows, full screen windows or both?

nornagon:
Once the "current" monitor is fetched, it should be possible to open either a fullscreen or a windowed mode window on that monitor. For my particular use case I'm interested in opening a windowed mode window on the current monitor. This API would be a sufficient addition to accomplish that using other monitor geometry querying methods that are already available.

DerekCresswell:
Do I understand correctly, you want to get the monitor a window is on, or "most" on?

nornagon:
@DerekCresswell both macOS and Windows have APIs to perform this function, as mentioned in the top post.

elmindreda:
This would be a useful addition. I propose `glfwGetClosestMonitor` to (hopefully) help tell it apart from `glfwGetWindowMonitor`.

nornagon:
@elmindreda thanks! IMO `Closest` implies geometrically closest, which isn't what the mentioned APIs do. Perhaps `glfwGetFocusedMonitor`?

slawrence2302:
Here are a couple functions I have been using to get these behaviors using the existing API. The first gets the monitor the cursor is currently on (get focused monitor?). This requires the window to be created first, but if the window is resized/repositioned immediately after creation it is not even visible in the wrong location before it is moved. The second gets the monitor the window is most on by area (get closest monitor). Note, these are actually translated versions because my original versions further functionalize some of this stuff, so may have some typo bugs.

```C
bool glfw_get_mouse_monitor(
    GLFWmonitor** monitor,
    GLFWwindow* window
[...]

PR #2220, opened 2022-11-20
New window API function glfwGetMonitorFromWindow (something like that was requested in issue #1699)

Shalindor:
Maybe useful, had to implement it to aid in handling irregular multi-monitor setups

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -954,6 +954,94 @@ GLFWAPI void glfwSetWindowAttrib(GLFWwindow* handle, int attrib, int value)
     _glfwInputError(GLFW_INVALID_ENUM, "Invalid window attribute 0x%08X", attrib);
 }
 
+typedef struct _rect { int x, y; int w, h; } _rect;
+
+#define RECT_INTERSECTS(ra, rb) \
+  ((ra)->x <= ((rb)->x + (rb)->w) && ((ra)->x + (ra)->w) >= (rb)->x && \
+  (ra)->y <= ((rb)->y + (rb)->h) && ((ra)->y + (ra)->h) >= (rb)->y)
+
+static _rect _get_intersection(_rect* ra, _rect* rb)
+{
+  _rect result = { 0, 0, 0, 0 };
+  if (RECT_INTERSECTS(ra, rb))
+  {
+    result.x = _glfw_max(ra->x, rb->x);
+    result.w = _glfw_min((ra->x + ra->w), (rb->x + rb->w)) - result.x;
+    result.y = _glfw_max(ra->y, rb->y);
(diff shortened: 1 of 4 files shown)


ws909:
#1699 

(For the purpose of linking these issues)

dougbinks:
Looking at this along with #2404 the questions I'd like to raise about the returned monitor are:

1. Should the monitor returned correspond to the monitor which the window will fill if maximised? This is potentially not the same as the monitor the window covers the most.
2. What should be returned for a iconified window?

src/window.c
```diff
@@ -954,6 +954,95 @@ GLFWAPI void glfwSetWindowAttrib(GLFWwindow* handle, int attrib, int value)
     _glfwInputError(GLFW_INVALID_ENUM, "Invalid window attribute 0x%08X", attrib);
 }
 
+typedef struct _rect { int x, y; int w, h; } _rect;
+
+#define RECT_INTERSECTS(ra, rb) \
+  ((ra)->x <= ((rb)->x + (rb)->w) && ((ra)->x + (ra)->w) >= (rb)->x && \
+  (ra)->y <= ((rb)->y + (rb)->h) && ((ra)->y + (ra)->h) >= (rb)->y)
```
dougbinks, line 961: Since this is only used once should just copy+paste inline into that function.

src/window.c
```diff
@@ -954,6 +954,95 @@ GLFWAPI void glfwSetWindowAttrib(GLFWwindow* handle, int attrib, int value)
...
+
+static _rect _get_intersection(_rect* ra, _rect* rb)
```
dougbinks, line 963: I think `_` prefix is only used on internal functions with `glfw` in their name. All helper functions do not appear to have this prefix.

src/window.c
```diff
@@ -954,6 +954,95 @@ GLFWAPI void glfwSetWindowAttrib(GLFWwindow* handle, int attrib, int value)
...
+      scratchRect = _get_intersection(&windowRect, &monitorRect);
+
+      currentDim = scratchRect.w * scratchRect.h;
+      overlapDim = overlapRect.w * overlapRect.h;
+
+      if (currentDim > 0 && currentDim > overlapDim)
+      {
+        overlapRect = scratchRect;
+        overlapMonitor = i;
+      }
```
dougbinks, line 1034: It might be simpler to compute and store the area.

SuperUserNameMan:
Hope you don't mind me interfering, but anything that relies on `glfwGetWindowPos()` would fail on Wayland platform, because it does not allow to know the position of a window.

dougbinks:
Yes, I'd temporarily forgotten about that. Wayland will need a separate solution, so using a platform function pointer with most of them returning the generic window area / monitor area calculation function would be required.

PR #2228, opened 2022-12-16
Win32: Add dark mode for title bar

floppyhammer:
Result:

dark_title_bar


ws909:
I believe this PR should not be merged into GLFW. I think GLFW benefits more from a unified API for managing themes across all platforms. Camilla has also given her support for at least the most basic theming capabilities, in the past. This PR will be very useful for such an API, though.

ws909:
@floppyhammer 

Any chance you could have a look at #2260, and help with the Win32 implementation?

floppyhammer:
@ws909 Yes, I'll see what I can do.

elmindreda:
Depending on how long it takes to get explicit support for themes done, I think a patch that applies the system theme to the GLFW window frame could still be useful for parity with the other platforms. Unless it breaks the intended future defaults, but I don't think it would?

ws909:
@floppyhammer Thanks. :)

@elmindreda Yeah, you're probably right. MacOS adapts automatically too, so this is in line with the intended behaviour of GLFW. My goal was to reduce the need for refactoring, but this PR is not very large, so it's not necessarily a lot of work, especially not if the same author helps with the Win32 theming implementation.

ws909:
I notice that you seem to be squashing commits rather frequently for a PR that seems to be complete, meaning, minor edits. I thought I'd just let you know that you should also add yourself to the CONTRIBUTORS.MD file, as well as a one-line update to the readme about the change. :) Else Camilla will make those changes and squash everything again, when merging.

elmindreda:
Yes, please do!

CONTRIBUTORS.md
```diff
@@ -260,6 +260,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Haoyun Chen
```
ws909, line 263: Should be in alphabetical order by last name.

README.md
```diff
@@ -121,6 +121,7 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+ - [Win32] Added support for dark title bar #2228
```
ws909, line 124: Should be inserted at the top of the other [win32] items in this last.

ws909 (approved):

README.md
floppyhammer: @ws909 Done.

ws909 (approved):

arsiliath:
Is this still being worked on? Would love this feature.

floppyhammer:
@arsiliath It's finished.

arsiliath:
> @arsiliath It's finished.

Thank you for your reply. To use it, would I have to check out this branch and build from source?

floppyhammer:
@arsiliath Yes.

PR #2229, opened 2022-12-17
Potential memory leak when terminating joystick support

mikejzx:
regfree was only being called in _glfwTerminateJoysticksLinux if the inotify instance was valid.  This seems to be erroneous as the regex is always compiled in _glfwInitJoysticksLinux even if the inotify instance initialisation fails.

### DIFF
--- a/src/linux_joystick.c
+++ b/src/linux_joystick.c
@@ -326,7 +326,9 @@ GLFWbool _glfwInitJoysticksLinux(void)
 
     // Continue without device connection notifications if inotify fails
 
-    if (regcomp(&_glfw.linjs.regex, "^event[0-9]\\+$", 0) != 0)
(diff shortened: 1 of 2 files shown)


elmindreda:
Thank you for the bug report and fix!

This fix introduces a different bug, however. If GLFW is terminated before the joystick module has been initialized, `regfree` will be called with invalid data on termination. There needs to be a boolean tracking whether the regex has been compiled.

mikejzx:
Please review my latest changes--I've addressed the issue you noted by adding a boolean which tracks whether the regex is compiled.

elmindreda:
Looks good! This was merged as c992226a9c63f80e157d6bd9bf2372acd2a1c795. Thank you!

PR #2235, opened 2022-12-22
glfwGetCocoaView() request to cocoa_window.m

mightgoyardstill:
is it worth adding a getCocoaView function to expose the NSView into our programs?
IMO saves having to write obj-c/obj-c++ files when it could be simply added to the bottom?

```
GLFWAPI id glfwGetCocoaView(GLFWwindow* handle)
{
    _GLFWwindow* window = (_GLFWwindow*) handle;
    _GLFW_REQUIRE_INIT_OR_RETURN(nil);

    if (_glfw.platform.platformID != GLFW_PLATFORM_COCOA)
    {
        _glfwInputError(GLFW_PLATFORM_UNAVAILABLE,
                        "Cocoa: Platform not initialized");
        return NULL;
    }

    return window->ns.view;
}
```


dougbinks:
I have an in-progress version of this available on my branch https://github.com/glfw/glfw/compare/master...dougbinks:glfw:get-cocoa-view

dougbinks:
Many thanks for the feature request and code snippet. If you look at the commit above you'll see you've been credited in the contributors under your github alias, let me know if you want to use your name instead.

PR #2238, opened 2022-12-28
Update logic for X11 RandR version check

### LINKED ISSUE #2236: X11, tiny logical operator bug
https://github.com/glfw/glfw/blob/57cbded0760a50b9039ee0cb3f3c14f60145567c/src/x11_init.c#L717-L718

the comment says:
**// The GLFW RandR path requires at least version 1.3**

the logical or should be replaced with logical and. (&& instead of ||)

bmitc:
Closes #2236

Just trying to tackle some low hanging fruit to help close out an issue.

Changed major to be greater than or equal to 1, rather than just greater than, and switched from OR to AND operator.

### DIFF
--- a/src/x11_init.c
+++ b/src/x11_init.c
@@ -715,7 +715,7 @@ static GLFWbool initExtensions(void)
                                 &_glfw.x11.randr.minor))
             {
                 // The GLFW RandR path requires at least version 1.3
-                if (_glfw.x11.randr.major > 1 || _glfw.x11.randr.minor >= 3)
+                if (_glfw.x11.randr.major >= 1 && _glfw.x11.randr.minor >= 3)
                     _glfw.x11.randr.available = GLFW_TRUE;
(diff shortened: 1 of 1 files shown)


src/x11_init.c
```diff
@@ -715,7 +715,7 @@ static GLFWbool initExtensions(void)
                                 &_glfw.x11.randr.minor))
             {
                 // The GLFW RandR path requires at least version 1.3
-                if (_glfw.x11.randr.major > 1 || _glfw.x11.randr.minor >= 3)
+                if (_glfw.x11.randr.major >= 1 && _glfw.x11.randr.minor >= 3)
```
bmitc, line 718: Going by the comment *only*, the major version check should be greater than or equal to 1 and not just greater than 1, and the logical operator should be AND (`&&`) and not OR (`||`) between the major and minor version checks.

PR #2244, opened 2023-01-04
Add `glfwSetApplicationIcon`

### LINKED ISSUE #2041: Set macOS Dock icon
This is a feature request to add an API to set the Dock icon in macOS. Currently, `glfwSetWindowIcons` emits a `GLFW_FEATURE_UNAVAILABLE` error on macOS. This makes sense, as window icons in macOS reflect a document. However, it is possible to change the application Dock icon at runtime. Activity Monitor (by Apple) animates the window icon to present a graph of resource usage. `glfwSetWindowIcons` could be implemented to set the Dock icon instead. Though, if I'm not mistaken, changing the task bar icon is also possible in Windows, though I'm unaware if that actually is the window icon. If not, a `glfwSetApplicationIcons` could be of interest.

ws909:
Resolves #2041 

**Status**: Replaced by #2293.

Emits `GLFW_FEATURE_UNIMPLEMENTED` on X11, Wayland and Win32.

The Cocoa implementation simply selects the largest provided `GLFWImage`.
Tested on MacOS 12.5.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -99,6 +99,7 @@ video tutorials.
  - Aaron Jacobs
  - JannikGM
  - Erik S. V. Jansson
+ - Andreas O. Jansen
  - jjYBdx4IL
  - Toni Jovanoski
  - Arseny Kapoulkine
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+ - Added `glfwSetApplicationIcon` to set the application's icon (#2041)
  - Added `GLFW_PLATFORM` init hint for runtime platform selection (#1958)
  - Added `GLFW_ANY_PLATFORM`, `GLFW_PLATFORM_WIN32`, `GLFW_PLATFORM_COCOA`,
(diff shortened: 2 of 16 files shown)


ws909:
Just noticed there's an issue with this PR. The icon is not reverted if the count is zero, and some asserts should be added to the implementation of `glfwSetApplicationIcon`. Will fix later today.

ws909:
The current API cannot be implemented on Windows and Wayland. Wayland does not have any protocol for this functionality. On Windows, taskbar icons and their labels are set per window, not per application. There are three solutions to this:

1. Keep `glfwSetApplicationIcon(count, icon)`, and provide another function: `glfwSetTaskbarIcon(window, count, icon)`
2. `glfwSetDockIcon(window, count, icon)`
  a. Sets the MacOS dock icon on each invocation
  b. Sets the MacOS dock icon when `window` is `NULL`.

Solution 1 and 2.b provide the simplest API. Solution 2.a requires more careful use by the client, and if their intent is different icons for different windows, they will run into issues on MacOS, where the Dock icon will be overridden multiple times. To solve this, clients must check the platform beforehand, and conditionally execute this method. 

I think solution 1 is the most clear and straightforward, however, 2.b limits the API surface.

ws909:
Replaced by #2293.

Issue #2161, opened 2022-07-29
Doesn't build on MacOSX Mountain Lion


chaosesqueteam:
glfw fails to build on lion and mountain lion, x86_64 mac.
glfw-devel also fails to build.

The error on 10.5 is:

{{{
:debug:main Starting logging for glfw @3.3.7_0
...
CMakeFiles/glfw.dir/nsgl_context.m.o -c /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_graphics_glfw/glfw/work/glfw-3.3.7/src/nsgl_context.m
:info:build /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_graphics_glfw/glfw/work/glfw-3.3.7/src/cocoa_window.m:328:28: warning: instance method '-occlusionState' not found (return type defaults to 'id') [-Wobjc-method-access]
:info:build     if ([window->ns.object occlusionState] & NSWindowOcclusionStateVisible)
:info:build                            ^~~~~~~~~~~~~~
:info:build /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_graphics_glfw/glfw/work/glfw-3.3.7/src/cocoa_window.m:328:46: error: use of undeclared identifier 'NSWindowOcclusionStateVisible'
:info:build     if ([window->ns.object occlusionState] & NSWindowOcclusionStateVisible)
:info:build                                              ^
:info:build /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_graphics_glfw/glfw/work/glfw-3.3.7/src/cocoa_window.m:328:28: warning: instance method '-occlusionState' not found (return type defaults to 'id') [-Wobjc-method-access]
:info:build     if ([window->ns.object occlusionState] & NSWindowOcclusionStateVisible)
:info:build                            ^~~~~~~~~~~~~~
:info:build 2 warnings and 1 error generated.
:info:build make[2]: *** [src/CMakeFiles/glfw.dir/cocoa_window.m.o] Error 1
}}}
 
This particular package is needed by many opengl apps.
Not all. But those who don't need it still like to build it as a reassurance that opengl actually works on their machine before building any other opengl project.

PR #2254, opened 2023-01-16
Cocoa: Patched use of macOS 10.9 window occlusion API for macOS 10.8 compatibility

NeRdTheNed:
This fixes building against macOS SDKs older than 10.9, noted in #2161.

I'm not super familiar with Objective-C and this is my first time contributing to GLFW, so hopefully this PR is OK! I've tested this fix by building GLFW against the macOS 10.8 SDK on macOS 10.14, which now works.


elmindreda:
I'm sorry, I lost track of this PR just before fixing #2161. I added you to the credits for the next release.

PR #2255, opened 2023-01-21
ci: Update to actions/checkout@v3

jpr42:
Fixes the following CI warning:

"Node.js 12 actions are deprecated. Please update the following
actions to use Node.js 16: actions/checkout@v2..."


elmindreda:
Thank you for the fix!

PR #2256, opened 2023-01-21
Minor CMake Cleanup

jpr42:
- No need to set `CMP0054` since the cmake min is 3.4
- No need to set `CMAKE_LEGACY_CYGWIN_WIN32` since the cmake min is 3.4
- Set `GLFW_STANDALONE` via `string(COMPARE EQUAL ...)`


CMakeLists.txt
```diff
@@ -2,12 +2,6 @@ cmake_minimum_required(VERSION 3.4...3.20 FATAL_ERROR)
 
 project(GLFW VERSION 3.4.0 LANGUAGES C)
 
-set(CMAKE_LEGACY_CYGWIN_WIN32 OFF)
```
juan-lunarg, line 5: "On CYGWIN, the undocumented “CMAKE_LEGACY_CYGWIN_WIN32” mode for compatibility with CMake versions older than 2.8.4 has been removed." - https://discourse.cmake.org/t/cmake-3-26-0-rc1-is-ready-for-testing/7394

PR #2259, opened 2023-01-25
Win32: Prevent white window background during initialization

W4RH4WK:
Some applications do not render frames immediately after spawning a window.
Currently, GLFW windows have a white background, which causes a white flash when an application is launched.

This white flash is quite annoying in the situations GLFW is used (assuming mainly games and similar applications that will start with a dark background).
A white flash before a dark background is far worse than displaying a black window before light content takes over.

This commit sets the window's background color to black and lets Windows handle painting the background.

Alternatively, if having a fixed, dark background is not desired, I suggest setting `hbrBackground` to `NULL` or `(HBRUSH)GetStockObject(BLACK_BRUSH)` depending on a hint.
`WM_ERASEBKGND` shouldn't be handled either way.

Here's an example where I've added a 1 second sleep between window creation and rendering the first frame.

Before:

https://user-images.githubusercontent.com/328798/214606517-1f2acd91-81c8-4955-8f1d-4ecb4c6fb90f.mp4

After:

https://user-images.githubusercontent.com/328798/214606559-8139aaac-68bd-487e-858c-0b1127934f92.mp4

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -1122,11 +1122,6 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM l
             break;
         }
 
-        case WM_ERASEBKGND:
-        {
-            return TRUE;
-        }
-
         case WM_NCACTIVATE:
         case WM_NCPAINT:
         {
@@ -1265,6 +1260,7 @@ static int createNativeWindow(_GLFWwindow* window,
         wc.lpfnWndProc   = windowProc;
         wc.hInstance     = _glfw.win32.instance;
         wc.hCursor       = LoadCursorW(NULL, IDC_ARROW);
+        wc.hbrBackground = (HBRUSH)GetStockObject(BLACK_BRUSH);
 #if defined(_GLFW_WNDCLASSNAME)
         wc.lpszClassName = _GLFW_WNDCLASSNAME;
 #else


ws909:
As a workaround, can you not postpone showing the window until after all setup is completed? Create a hidden window, then prepare everything, then show and render? Possibly render, then show?

W4RH4WK:
As a workaround, yes, I could do that. I also don't have a problem applying this patch to the version of GLFW I am using, I am building from source anyway.

I just thought, the behavior in my PR is more desirable. (And closer to SDL if that's relevant.)

ws909:
Does this apply to OpenGL or Vulkan? It doesn't seem to be an issue on MacOS in Vulkan, and the default OpenGL frame buffer is black in OpenGL. Is it possible to fix this by setting the `glClearColor` immediately after window creation?

W4RH4WK:
It's not related to any graphics API, but just how Windows handles windows drawing. At that moment, it's not the frame buffer that is displayed, but the raw background of the window; hence the change to the window class' `hbrBackground` field and the `WM_ERASEBKGND` signal.

ws909:
Thanks for the clarifications. I'm wondering if perhaps the right solution is to set the "raw" background of the window to be equal to the title bar and border colour. I think that's what UWP does. If not, perhaps a window hint, but that's potentially bloat.

W4RH4WK:
I definitely prefer black, but if that's too harsh, `COLOR_WINDOWFRAME` might be the way to go.

For me it looks like this:

image

ws909:
Yeah, that's the border colour, and I agree that it doesn't look right. Is there one for the title bar colour instead?

W4RH4WK:
> Perhaps [this](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.systemcolors.window?view=net-7.0)?

That link goes to the .Net documentation. Which is a lot different from the ancient win32 API we have to use here.
[This](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassa#members) is the documentation for the `hbrBackground` field. It lists available brushes (colors). I've tried all of them and the `COLOR_WINDOWFRAME` is the closest. There doesn't seem to be one for the title bar.

I am currently looking into other parts of the API.

ws909:
> That link goes to the .Net documentation. Which is a lot different from the ancient win32 API we have to use here. 

I know, but I made a quick search, didn't find the Win32 equivalent, but found that, and right below it is the WindowFrame (similar enough to what you used in Win32), so I figured it was sufficient. ;) 

> [This](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassa#members) is the documentation for the hbrBackground field. 

Thanks. ;P

> It lists available brushes (colors). I've tried all of them and the COLOR_WINDOWFRAME is the closest. There doesn't seem to be one for the title bar.

:/ I don't have experience with Win32, so I'm not aware of a solution. I did do some basic fiddling in C# with UWP many years ago, though, and I can remember that the default behaviour was to use the title bar colour (user selected theme colour). Just checked some basic apps on my Windows 10 computer, and that's the behaviour for those as well, so hopefully there is similar functionality in Win32, if this is behaviour that suits GLFW best. I think this looks best, and I personally find both black and white to be too harsh, but black is better than white, at least for those of us using dark modes. Don't weigh in my opinion too much.

ws909:
@W4RH4WK Is it possible that the `COLOR_WINDOWFRAME` looks different because this color has not been adjusted for the dark mode? So, if the mode is bright, the color will not be adjusted, and the the title bar and window background will look the same?

W4RH4WK:
Not sure if I understand your question correctly.

I am using Windows 10 and my color theme is set to dark.

image

The title bar color can be set independently from the rest.

ws909:
Couldn't find anything on Microsoft's sites about it, but Apple has [this page](https://developer.apple.com/design/human-interface-guidelines/foundations/color), with different colour values for the system colours, for the light and dark modes, [This Stackexchange Meta answer](https://meta.stackexchange.com/a/386167) provides some insight too.

W4RH4WK:
Since there doesn't seem to be a _definitive answer_ should we reconsider adding a hint here?

ws909:
Neh, let’s not keep this on hold. If this is a better default for most applications, it’s a fine change. We can always revisit this topic later. 

Dispute resolved; @elmindreda.

PR #2260, opened 2023-01-25
Add theme support to GLFW

### LINKED ISSUE #2257: Theme support
This is a feature request for adding basic theming support to GLFW.

Goals:
- Query the current (default) system theme (light/dark).
- Query for a system theme colour if the platform supports it.
- Get notified every time the system changes its theme (user edited it, MacOS's automatic day/night, etc).
- Set a window's theme to change automatically based on the system theme (default GLFW behaviour).
- Set a window's theme (light/dark).
- Set a specific theme colour for a window, if the platform supports it.

Similar issues and PRs:
- #1516
- #2228

This is a very rough idea of what this API might look like:

```c
typedef struct GLFWtheme
{
    int baseTheme; // light/dark
    int hasColor;
    float color[4];
} GLFWtheme;

#define GLFW_THEME_LIGHT 0
#define GLFW_THEME_DARK 1

typedef void (* GLFWthemefun)(GLFWtheme* theme);

/*! @brief Notifies the application when the system default theme changes.
 *
 */
GLFWAPI void glfwSetSystemThemeCallback(GLFWthemefun callback);

/*! @brief Sets the theme for a window.
 *
 *  @param[in] window The window to set the theme for.
 *  @param[in] theme The theme to set. Pass `NULL` to set it to the system default.
 */
GLFWAPI void glfwSetTheme(GLFWwindow* window, GLFWtheme* theme);
GLFWAPI GLFWtheme* glfwGetTheme(GLFWwindow* window);
GLFWAPI GLFWtheme* glfwGetSystemDefaultTheme();
```

Windows supports custom colours for its window borders and title bars. The libdecor decorations for Wayland likely supports this too. MacOS does not support custom colours, but does support light/dark modes. MacOS does however have an accent colour, which is used for text. The accent colour is both gettable and settable.

ws909:
Resolves #2257

This is a rough draft for theming support. You're welcome to make contributions. Win32, X11 and Wayland (libdecor (not yet merged)) platforms need a developer.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -5845,6 +5845,37 @@ GLFWAPI void glfwSetClipboardString(GLFWwindow* window, const char* string);
  */
 GLFWAPI const char* glfwGetClipboardString(GLFWwindow* window);
 
+typedef struct GLFWtheme
+{
+    int baseTheme; // light/dark
+    int flags;
+    unsigned char color[4];
+} GLFWtheme;
+
+#define GLFW_BASE_THEME_DEFAULT 0
+#define GLFW_BASE_THEME_LIGHT 1
+#define GLFW_BASE_THEME_DARK 2
+
+#define GLFW_THEME_FLAG_HAS_COLOR 1
+#define GLFW_THEME_FLAG_HIGH_CONTRAST 2
+#define GLFW_THEME_FLAG_VIBRANT 4
+
+typedef void (* GLFWthemefun)(GLFWtheme* theme);
+
+/*! @brief Notifies the application when the system default theme changes.
+ *
+ */
+GLFWAPI GLFWthemefun glfwSetSystemThemeCallback(GLFWthemefun callback);
+
+/*! @brief Sets the theme for a window.
+ *
+ *  @param[in] window The window to set the theme for.
+ *  @param[in] theme The theme to set. Pass `NULL` to set it to the system default.
+ */
+GLFWAPI void glfwSetTheme(GLFWwindow* handle, GLFWtheme* theme);
+GLFWAPI GLFWtheme* glfwGetTheme(GLFWwindow* handle);
+GLFWAPI GLFWtheme* glfwGetSystemDefaultTheme();
+
 /*! @brief Returns the GLFW time.
  *
  *  This function returns the current GLFW time, in seconds.  Unless the time
--- a/src/cocoa_init.m
+++ b/src/cocoa_init.m
@@ -704,7 +704,8 @@ void _glfwTerminateCocoa(void)
 
 GLFWtheme* _glfwGetSystemDefaultThemeCocoa(void)
 {
-    return NULL;
+    _glfwInputError(GLFW_FEATURE_UNIMPLEMENTED, NULL);
+    return NULL; // TODO: implement
 }
 
 #endif // _GLFW_COCOA
--- a/src/cocoa_init.m
+++ b/src/cocoa_init.m
@@ -496,8 +496,11 @@ GLFWbool _glfwConnectCocoa(int platformID, _GLFWplatform* platform)
     const _GLFWplatform cocoa =
     {
         GLFW_PLATFORM_COCOA,
+        // Init
         _glfwInitCocoa,
         _glfwTerminateCocoa,
+        _glfwGetSystemDefaultThemeCocoa,
+        // Input
         _glfwGetCursorPosCocoa,
         _glfwSetCursorPosCocoa,
         _glfwSetCursorModeCocoa,
@@ -516,6 +519,7 @@ GLFWbool _glfwConnectCocoa(int platformID, _GLFWplatform* platform)
         _glfwPollJoystickCocoa,
         _glfwGetMappingNameCocoa,
         _glfwUpdateGamepadGUIDCocoa,
+        // Monitor
         _glfwFreeMonitorCocoa,
         _glfwGetMonitorPosCocoa,
         _glfwGetMonitorContentScaleCocoa,
@@ -524,6 +528,7 @@ GLFWbool _glfwConnectCocoa(int platformID, _GLFWplatform* platform)
         _glfwGetVideoModeCocoa,
         _glfwGetGammaRampCocoa,
         _glfwSetGammaRampCocoa,
+        // Window
         _glfwCreateWindowCocoa,
         _glfwDestroyWindowCocoa,
         _glfwSetWindowTitleCocoa,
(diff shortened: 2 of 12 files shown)


elmindreda:
Having a struct may become an issue in the future as theming evolves. Functions for separate properties on the window or a separate opaque theme object may be more flexible.

ws909:
> Having a struct may become an issue in the future as theming evolves. Functions for separate properties on the window or a separate opaque theme object may be more flexible.

That's definitely true, but I wanted to keep it simple for now. There are workarounds for extending the current structures, but those are not necessarily as pretty. I was also thinking that by the time there's a need to extend it, GLFW 4 is upcoming.

Nonetheless, I agree, and I'll change this into an opaque object.

> as theming evolves

Do you think the theming API should include more goals?
Do you have anything specific in mind about future additions/enhancements/extensions to the theming API?
Is there anything in particular that the current design obstructs?

The most likely change I can imagine, is adding more base themes, if "base themes" should even be a thing. The most simple change, is to make all light variations positive numbers, and all negative variations dark. As I'm writing this, I'm starting to believe a better change is to turn the theme struct into an opaque object, remove the notion of a "base theme" (at least for now), and replace it with a light/dark variation, and then extend this object with more functions in the future.

### Gaussian blur

I also hope it's possible to support a gaussian blur of some sort, like `NSVisualEffectView` (I believe it is), like in this screenshot. It's not possible to implement this in Vulkan, or OpenGL (without hacks), or another API, as it requires reading the frame buffer content of everything behind the window. Whether or not it's the compositor or CSD GUI frameworks doing this on the different platforms is irrelevant, but GLFW clients cannot use this without tapping into the native APIs directly, circumventing GLFW, potentially even causing issues for the GLFW implementations. The fullscreen case is easier for clients, as it only requires the currently active desktop image, and its positioning and scaling, but GLFW doesn't provide facilities for this either. 

I don't think this should be part of any theming API, but it is worth noting that the "vibrancy" appearances on MacOS use this, and that MacOS switches to the non-vibrant appearances when "reduce transparencies" is selected in the accessibility menu in the preferences application.

Even if this is included in the theming, it should not be a goal for initial theming support, and so should be on hold for now.

Skjermbilde 2023-01-27 kl  21 30 54

ws909:
The `GLFW_THEME_FLAG_HIGH_CONTRAST` is about an accessibility feature more than a theming feature, and so may be a better fit for an accessibility API for GLFW, rather than theming, but I included it here, because the theming implementation for MacOS uses NSAppearance, which is where the high-contrast setting is reported, but it's possible to decouple these. The MacOS implementation can only report that the system is in high contrast mode or not, but not actually override that for the window.

ws909:
Because themes are set for each window, rather than the whole application, on MacOS, this can have some unintended side effects. An example of this, is the "about application" window that's included in the default menu bar of the application. The theming API in this PR does not allow for setting the appearance of this window. It will always use the system appearance.

ws909:
There's a few ways to add functionality for checking if a window's theme adapts automatically, or if the client has previously overridden this:

1. Simply rely on the client to be aware of their own actions.
2. Make `GLFW_DONT_CARE` valid input and output for `glfwThemeGetAttribute` and `glfwThemeSetAttribute`. Potentially replace `GLFW_THEME_DEFAULT` with `GLFW_DONT_CARE`. Potentially replace the theme variation with a new attribute: `GLFW_THEME_ATTRIBUTE_VARIATION` or `...BRIGHTNESS`. Add a new 2nd parameter to `glfwThemeGetColor` and `glfwThemeSetColor`, which specifies which colour to get/set. Add new constant: `GLFW_THEME_COLOR_MAIN`. Make this a valid theme attribute and color specifier. Remove `GLFW_THEME_ATTRIBUTE_HAS_COLOR`.
3. Add a new function, `glfwGetThemeState(GLFWwindow* handle)` (or a similar name), which reports `GLFW_TRUE` for all theme properties that are specified by the client, and `GLFW_FALSE` for all properties that are inherited from the system theme.
4. Add a new function, `glfwGetThemeState(GLFWtheme* theme, int attribute)` which checks if a given attribute adapts automatically for the given theme, or if the theme contains a user-defined value for that attribute. This function must also accept checking for the variation/brightness and colour. It makes no sense to call this function for a theme retrieved from `glfwGetSystemDefaultTheme`.

The PR currently uses solution 1.

The least simple solution is number 2. This requires clients to additionally call `glfwGetSystemDefaultTheme` if they need a theme property of a window, and that window adapts automatically to the system theme.

Implementing solution 3 risks causing problems in the future.

ws909:
I'm considering a redesign.

Add:
```c
typedef struct GLFWcolor
{
    int flags;
    union
    {
        int[4] integers;
        float[4] floats;
    }
} GLFWcolor;
```

And either add this:

```c
// Makes GLFW select the colors from the best match. For instance, if the parent theme 
// defaults to DEFAULT variation, and there's a child with DARK variation, GLFW will use 
// the colors in that child, when the system theme is dark. If the child lacks colors, the 
// parent's colors for these cases will be used instead. If the system theme is light, and 
// there is no child theme with LIGHT, GLFW will use the colors from the parent. 
GLFWAPI void glfwThemeAddChild(const GLFWtheme* child);

GLFWAPI const GLFWtheme* const * glfwThemeGetChildren(int* count);

// or (I don't like this):

GLFWAPI void glfwThemeSetColor(GLFWtheme* theme, int specifier, int variation, GLFWcolor* color);
GLFWAPI void glfwThemeGetColor(GLFWtheme* theme, int specifier, int variation, GLFWcolor* color);
```

... or change the API to this:
```c
#define GLFW_ACCESSIBILITY_HIGH_CONTRAST 1
#define GLFW_ACCESSIBILITY_REDUCE_TRANSPARENCY 2
#define GLFW_ACCESSIBILITY_REDUCE_MOTION 3

typedef void (* GLFWaccessibilityfun)(int attribute, int value);

GLFWAPI GLFWaccesibilityfun glfwSetAccessibilityCallback(GLFWaccesibilityfun* callback);
GLFWAPI int glfwGetAccessibilityAttribute(int attribute);

// GLFW_DONT_CARE
#define GLFW_THEME_LIGHT 0
#define GLFW_THEME_DARK 1

#define GLFW_THEME_COLOR_MAIN 1

typedef void (* GLFWthemefun)(int theme);
typedef void (* GLFWcolorfun)(int specifier, GLFWcolor* color);

GLFWAPI GLFWthemefun glfwSetThemeCallback(GLFWthemefun callback);
GLFWAPI GLFWcolorfun glfwSetColorCallback(GLFWcolorfun callback);

GLFWAPI int glfwGetTheme();
GLFWAPI void glfwGetColor(int specifier, GLFWcolor* color); // NULL if none available

GLFWAPI int glfwGetWindowTheme(GLFWwindow* window);
GLFWAPI void glfwSetWindowTheme(GLFWwindow* window, int theme);

GLFWAPI void glfwSetWindowColor(GLFWwindow* window, int specifier, GLFWcolor* color); // NULL to reset to default
GLFWAPI void glfwGetWindowColor(GLFWwindow* window, int specifier, GLFWcolor* color); // NULL if none available
```

The latter is easier to implement, easier to understand, but requires more work from the user for automation.

@elmindreda
Any chance of an opinion?

Issue #1834, opened 2021-01-14
Vsync broken in macOS 11/m1 again?


nevyn:
I'm noticing that `glfwSwapInterval` doesn't seem to be doing anything on my M1 Mac Mini running Big sur. Looking at the issue history (in particular https://github.com/glfw/glfw/issues/1337), it seems this is not the first time it's happening :/

My monitor is 50hz, but I get ~100fps no matter what swap interval I set -- 0, 1, 50, whatever, I always get ~100fps in my app.

LilaQ:
I have a similar problem, M1 with Big Sur, SDL2 project is locked to 60fps, no matter what Vsync settings I choose.

PR #2277, opened 2023-02-19
macOS: Use CVDisplayLink for vsync

### LINKED ISSUE #2249: Swap interval ignored on macOS 13
OpenGL swap interval seems to have broken again, as reported in #1990, after having been fixed in 12.1.

Philanatidae:
Fixes #2249, #1834.

Somewhere along the line, Apple broke NSOpenGLCPSwapInterval. I've been researching this for the past week or so from various sources, and I decided to go ahead with using CVDisplayLink instead of NSOpenGLCPSwapInterval. SDL ended up doing the same thing (see issue thread [here](https://github.com/libsdl-org/SDL/issues/4918)).

I fixed it in my fork, and I decided to make this PR back to upstream for discussion on whether GLFW wants to go down this route as well.

The approach is very similar to SDL; pause the main thread right before swapping buffers in NSGL (in `glfwSwapBuffers`), and wait for the CVDisplayLink callback to continue. `glfwSwapInterval` is respected by waiting that many callbacks.

Because `cocoa_platform.h` uses a mutex, I had to shuffle around the includes in `internal.h` to get it included before `cocoa_platform.h`. I also had to add conditional variables to `posix_thread.h` and `win32_thread.h`. Multi-threading isn't my strong suite so I wouldn't be surprised if this can be improved further.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -159,10 +159,11 @@ endif()
 if (GLFW_BUILD_COCOA)
     target_link_libraries(glfw PRIVATE "-framework Cocoa"
                                        "-framework IOKit"
-                                       "-framework CoreFoundation")
+                                       "-framework CoreFoundation"
+                                       "-framework CoreVideo")
 
     set(glfw_PKG_DEPS "")
-    set(glfw_PKG_LIBS "-framework Cocoa -framework IOKit -framework CoreFoundation")
+    set(glfw_PKG_LIBS "-framework Cocoa -framework IOKit -framework CoreFoundation -framework CoreVideo")
 endif()
 
 if (GLFW_BUILD_WAYLAND)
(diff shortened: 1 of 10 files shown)


ws909:
@krackers A runtime check is better, so that it’s not needed to recompile for each OS version. 

———————

This PR reintroduces CoreVideo as a dependency. There’s been concern from some people that fixes for the swap interval haven’t been merged yet. How much of an issue is this?

ws909:
> What's the concern with having CoreVideo as a dependency? Or by "how much of an issue is this" were you referring to the tearing on < 10.14? Since glfw seems to support down to 10.8 (per the faq), it would break a number of platforms.

No, I was referring to the problem of reintroducing CoreVideo (CV; CVDisplayLink) as a dependency. 
I was referring to [this](https://www.glfw.org/docs/3.4/news.html):

> CoreVideo dependency has been removed
GLFW no longer depends on the CoreVideo framework on macOS and it no longer needs to be specified during compilation or linking.

PR #2285, opened 2023-02-27
Initial support for libdecor with run-time loading

elmindreda:
This is my proposal for initial libdecor support, which is an off-shoot from #1693. It should do everything this PR does, while adding (optional) dynamic loading and implementing libdecor paths for a few more GLFW features. It's a re-implementation that came out of learning the libdecor API, so any bugs in it are my fault and not @christianrauch's.

libdecor is a really neat library and fit in easily once a few things in GLFW were cleaned up. I'm likely using it incorrectly in places, but it mostly Works on My Machine now. Tested with libdecor 0.1.0-3build1 on Ubuntu 22.04.

I kept the existing client-side decorations as a fallback of last resort, as in their current state I think they're better than no decorations at all, but with libdecor installed you should never see them.

Currently toggling `GLFW_DECORATED` isn't working well with Cairo decorations and isn't working at all with XDG decorations, and I haven't been able to figure out why.

I don't feel entirely confident about the workaround I added to make sure libdecor's globals are in place if some other part of `glfwInit` fails and the libdecor context is destroyed before the regular event pump starts. This is needed because the 0.1.0 Cairo plugin destroys its shm global unconditionally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,7 @@ information on what to include when reporting a bug.
  - [X11] Bugfix: The OSMesa libray was not unloaded on termination
  - [X11] Bugfix: A malformed response during selection transfer could cause a segfault
  - [X11] Bugfix: Some calls would reset Xlib to the default error handler (#2108)
+ - [Wayland] Added improved fallback window decorations via libdecor (#1639,#1693)
  - [Wayland] Added dynamic loading of all Wayland libraries
  - [Wayland] Added support for key names via xkbcommon
  - [Wayland] Added support for file path drop events (#2040)
--- a/docs/compat.dox
+++ b/docs/compat.dox
@@ -128,6 +128,14 @@ wayland-protocols 1.6, and mandatory at build time.  If the running compositor
 does not support this protocol, the screensaver may start even for full screen
 windows.
 
+GLFW uses the [libdecor library](https://gitlab.freedesktop.org/libdecor/libdecor)
+for window decorations, where available.  This in turn provides good quality
+client-side decorations (drawn by the application) on desktop systems that do
+not support server-side decorations (drawn by the window manager).  On systems
+that do not provide either libdecor or xdg-decoration, very basic window
+decorations are provided.  These do not include the window title or any caption
+buttons.
+
 GLFW uses the [xdg-decoration
 protocol](https://cgit.freedesktop.org/wayland/wayland-protocols/tree/unstable/xdg-decoration/xdg-decoration-unstable-v1.xml)
 to request decorations to be drawn around its windows.  This protocol is part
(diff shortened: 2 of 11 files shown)


elmindreda:
I kept the cushion of preparatory fixes inside this PR this time as they may need to be changed as a result of the pending review.

christianrauch:
Overall this works quite nicely. The interactive resizing of the `splitview` example is quite "gluey" compared to the X11 case. I think this is because the rendering takes longer than the framerate. When running via X11 you can see that the window content gets cut off or the window expands with no content. The other examples run well and are responsive and it is nice to see that the window size and content are synchronised when using Wayland.

One general Wayland thing I noticed is that, if X11 and Wayland are enabled at compile time, X11 will still take precedence. You must set `GLFW_BUILD_X11` to OFF to use the Wayland front end. In my view, this should be the other way around, first try to connect to a Wayland server and if unavailable, connect to an X11 display server.

gracicot:
> One general Wayland thing I noticed is that, if X11 and Wayland are enabled at compile time, X11 will still take precedence. You must set GLFW_BUILD_X11 to OFF to use the Wayland front end. In my view, this should be the other way around, first try to connect to a Wayland server and if unavailable, connect to an X11 display server.

I noticed that too. The environment variables are also not enough to change which protocol glfw will use. It will only use wayland if I disable X11 support in CMake.

TheBrokenRail:
It's a bit of a hacky workaround, but you can also just unset the `DISPLAY` environmental variable to force it to use Wayland.

elmindreda:
> Overall this works quite nicely. The interactive resizing of the `splitview` example is quite "gluey" compared to the X11 case. I think this is because the rendering takes longer than the framerate. When running via X11 you can see that the window content gets cut off or the window expands with no content. The other examples run well and are responsive and it is nice to see that the window size and content are synchronised when using Wayland.

Yep, there's a lot of work left to do specifically around surface updates, in all three paths (libdecor, XDG and fallback). Glad to hear the code didn't entirely fall over, though! Need to catch up on the discussions around this in both SDL and libdecor.

> One general Wayland thing I noticed is that, if X11 and Wayland are enabled at compile time, X11 will still take precedence. You must set `GLFW_BUILD_X11` to OFF to use the Wayland front end. In my view, this should be the other way around, first try to connect to a Wayland server and if unavailable, connect to an X11 display server.

I didn't want to make the Wayland backend the default until it was fully usable. This PR will be a huge step towards that and there is already #2035 for when it's ready.

Note that, in addition to what @TheBrokenRail said, you can select the Wayland backend (or another one) before initialization by [setting the GLFW_PLATFORM init hint](https://www.glfw.org/docs/3.4/intro#platform).

elmindreda:
On that topic, what are your opinions on having a `GLFW_PLATFORM` environment variable that influences the selection if the init hint is set to `GLFW_ANY_PLATFORM`?

elmindreda:
@christianrauch Are you reasonably happy with this being merged in its current state as initial libdecor support?

elmindreda:
Arghle, I forgot to tag this PR in the commit. This has been merged now, just slightly rebased.

PR #2286, opened 2023-03-01
Window Taskbar Progress (closes #1183)

### LINKED ISSUE #1183: Adding Windows Taskbar "Effects"
Would it be possible to add Effects to the Windows Taskbar like progressbar or icons like it is done here (https://www.codeproject.com/Articles/42345/Windows-Goodies-in-C-Taskbar-Progress-and-Status) but im no C programmer. Maby is helps you guys.

It would be a nice feature, I hope it will be added.

GamesTrap:
This is my implementation for window taskbar progress support, which would resolve #1183.

# What's new

- `glfwSetWindowTaskbarProgress()` function to display the progress of some action in the taskbar.
- `GLFW_TASKBAR_PROGRESS_NOPROGRESS` value to stop displaying the taskbar progress.
- `GLFW_TASKBAR_PROGRESS_INDETERMINATE` value to display the taskbar progress in an indeterminate state (on Windows this cycles the progress from 0 to 100 repeatedly).
- `GLFW_TASKBAR_PROGRESS_NORMAL` value to display the user given progress status (0-100%) on the taskbar.
- `GLFW_TASKBAR_PROGRESS_ERROR` value to display an _error_ taskbar progress.
- `GLFW_TASKBAR_PROGRESS_PAUSED` value to display a _paused_ taskbar progress.

# Platform differences

## Windows

On Windows the [ITaskbarList3 interface](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-itaskbarlist3) is used to set the taskbar progress which is available since Windows 7.

## Linux

On Linux the [Unity LauncherAPI](https://wiki.ubuntu.com/Unity/LauncherAPI) is used via DBus. A valid application .desktop file with the same name as the executable is required for the API to work. As far as I know only KDE and the Unity DE use this API.

For the DBus communication I added an implementation which loads the needed functions at runtime from the `libdbus-1.so.3` shared library, see `posix_dbus.h` and `posix_dbus.c`.

`GLFW_TASKBAR_PROGRESS_INDETERMINATE`, `GLFW_TASKBAR_PROGRESS_ERROR` and `GLFW_TASKBAR_PROGRESS_PAUSED` are not supported by the Unity LauncherAPI so they instead behave like `GLFW_TASKBAR_PROGRESS_NORMAL`.

# Implementation status

- [x] Windows (supported since Windows 7)
- [x] Linux (X11 & Wayland)
- [ ] macOS (help wanted)

I'm unable to work on macOS support because I don't have any Apple hardware.
I found [DockProgress](https://github.com/sindresorhus/DockProgress) which implements what we want but is using Swift instead of Objective-C. Pull requests into my branch are welcome!

# Tested Platforms

- [x] Kubuntu 22.10 (X11 & Wayland)
- [x] Kubuntu 22.04 LTS (X11 & Wayland)
- [x] Windows 11 Pro (Visual Studio 2022)
- [x] Windows 10 Pro (Visual Studio 2019)
- [ ] Windows 8.1
- [ ] Windows 7
- [x] Windows Vista Ultimate (Visual Studio 2010)
- [ ] Windows XP
- [ ] macOS


ws909:
I'll help with the MacOS implementation in the next few days. :)

GamesTrap:
@ws909 Reading the Microsoft docs also indicates this behavior. They prioritize based on the progress state (error, paused, normal and indeterminate in this order). When two windows in a group try to set the normal progress value then the least progress gets used.

include/GLFW/glfw3.h
```diff
@@ -1272,6 +1272,57 @@ extern "C" {
...
 
+/*! @addtogroup window
+ *  @{ */
+/*! @brief Disable the progress bar.
+ *
+ *  Disable the progress bar.
+ *
+ *  Used by @ref window_taskbar_progress.
+ */
+#define GLFW_TASKBAR_PROGRESS_NOPROGRESS    0
```
ws909, line 1283: I think this is better described as `GLFW_TASKBAR_PROGRESS_DISABLED`.

include/GLFW/glfw3.h
```diff
@@ -3304,6 +3356,43 @@ GLFWAPI void glfwSetWindowTitle(GLFWwindow* window, const char* title);
...
+ *
+ *  @thread_safety This function must only be called from the main thread.
+ *
+ *  @sa @ref window_taskbar_progress
+ *
+ *  @since Added in version 3.4.
+ *
+ *  @ingroup window
+ */
+GLFWAPI void glfwSetWindowTaskbarProgress(GLFWwindow* window, const int progressState, int completed);
```
ws909, line 3394: The `completed` parameter seems like a boolean parameter. I think it's better described simply with the generic name "value".

include/GLFW/glfw3.h
GamesTrap: Done 3e7b016.

include/GLFW/glfw3.h
GamesTrap: Done a480de9.

ws909:
So, I created a quick working implementation. I'll create a PR for [this branch](https://github.com/ws909/glfw/commits/taskbar-progress) hopefully tomorrow, when it's finished, but while I looked at it, I wondered why you chose to use integers for the progress bar? Using float or double provides a higher precision. This is also what `NSProgressIndicator` uses (double).

The default progress bar in Cocoa is very thin. Some applications such as IntelliJ, use a much thicker progress bar, but doing this requires subclassing `NSProgressIndicator` with custom drawing. So while an extra `thickness` parameter could be added to the function, providing an implementation is much more complicated. It is also possible to add a new separate function; `glfwSetTaskbarProgressBarThickness` later in the future, if this is needed. It's also possible to modify the vertical position of the progress bar, but this requires a new parameter (or separate function) for GLFW users to set. I also tried setting a tooltip and accessibility names for the indicator, but this seemed to have no effect, so my initial idea to add that as a parameter, serves no purpose.

I figured a better way of displaying the progress when multiple windows are using the progress bar, is to calculate the combined completion of all of them. This can be improved if each progress is given a weight, which would require a new parameter (defaults to `GLFW_DONT_CARE`), but this is not needed.

GamesTrap:
> I'll create a PR for [this branch](https://github.com/ws909/glfw/commits/taskbar-progress) hopefully tomorrow, when it's finished

:+1: 

> I wondered why you chose to use integers for the progress bar?

I did this because on Windows the [ITaskbarList3::SetProgressValue()](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-itaskbarlist3-setprogressvalue) function only allows setting the progress value as a `ULONGLONG`. Now that you mentioned this, I also think it would be better to use a `double` (also used in the [Unity LauncherAPI](https://wiki.ubuntu.com/Unity/LauncherAPI)) instead and round the value to the nearest integer in the Windows implementation.

> So while an extra `thickness` parameter could be added to the function, providing an implementation is much more complicated.

This feature would be macOS specific, the [Unity LauncherAPI](https://wiki.ubuntu.com/Unity/LauncherAPI) and the [ITaskbarList3 interface](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-itaskbarlist3) both do not provide such functionality.

> It's also possible to modify the vertical position of the progress bar, but this requires a new parameter (or separate function)

Same as above.

> I figured a better way of displaying the progress when multiple windows are using the progress bar, is to calculate the combined completion of all of them.

Something like this? `double progressValue = combinedWindowProgress / (windowCount * 100.0)` so we keep the 0% to 100% range.  
Should this value then be set for each window which doesnt use the `GLFW_TASKBAR_PROGRESS_DISABLED` state?  
Should we also set the same state for each window (prioritizing error, paused, normal and indeterminate in this order)?

_On Windows the windows can be ungrouped (so each can display its own progress data), this would result in each window displaying the same progress and state._ :/

ws909:
> Something like this? double progressValue = combinedWindowProgress / (windowCount * 100.0) so we keep the 0% to 100% range.
Should this value then be set for each window which doesnt use the GLFW_TASKBAR_PROGRESS_DISABLED state?
Should we also set the same state for each window (prioritizing error, paused, normal and indeterminate in this order)?

Well, it was a comment on the MacOS implementation I’ll write, so apply that formula to the Dock tile’s `NSProgressIndicator` (Which there only exists one of). And `windowCount` will not be the application’s total window count, but the number of windows with a progress state other than `DISABLED`. And it doesn’t make sense to have this order of priorities for the tile’s state on MacOS. If any window has normal/paused/error (those are equal on MacOS), use that for the indicator, else the indeterminate state, and then disable it if no window uses it. So, in short, that part of the comment was completelt unrelated to the API itself, and all the other platform implementations.

ws909:
>  macOS (supported since macOS 10.0 [...]

The oldest version of MacOS that GLFW supports, is 10.8, so there shouldn't be any problems with the Dock progress indicator on any versions of MacOS. Cocoa replaced Carbon in 10.0 (first version of OS X), which was used in classic Mac OS. GLFW hasn't had a Carbon implementation since GLFW 2.

ws909:
Would `GLFW_PROGRESS_INDICATOR_` be a better prefix than `GLFW_TASKBAR_PROGRESS_`, for the defines?

GamesTrap:
> Would `GLFW_PROGRESS_INDICATOR_` be a better prefix than `GLFW_TASKBAR_PROGRESS_`, for the defines?

I named it `GLFW_TASKBAR_PROGRESS_` to make it clear that it should be used with `glfwSetWindowTaskbarProgress()`.

I can rename it but then we maybe should also rename the `glfwSetWindowTaskbarProgress()` function itself?.

ws909:
A `GLFW_PROGRESS_INDICATOR_` prefix specifies that this enum value is applicable to _any_ progress indicator.
This means that if GLFW ever adds progress indicators in any spots other than the taskbar/dock, this enum can be used for those as well. The likelihood of GLFW ever providing any other progress indicators, though, is very small.

`glfwSetWindowTaskbarProgress()` specifies that the specific progress indicator in question, is the one on the taskbar icon.

Personally, I would prefer changing `Taskbar` to `Dock` (simpler, shorter), or just remove it entirely.

`glfwSetWindowProgressIndicator()` could work. 

These are really not big issues.

GamesTrap:
My choice would be:

> A `GLFW_PROGRESS_INDICATOR_` prefix specifies that this enum value is applicable to _any_ progress indicator.

:+1:

> `glfwSetWindowProgressIndicator()` could work.

:+1:

GamesTrap:
@elmindreda The PR should now be feature complete and ready to be reviewed and merged (see updated first comment).

elmindreda:
@GamesTrap Thank you, will review as soon as I'm able!

GamesTrap:
I've updated the PR and also fixed a bug which was found in the SDL implementation that lead to crashes on big-endian systems.

For reference:

- https://github.com/libsdl-org/SDL/issues/13953
- https://github.com/libsdl-org/SDL/pull/14029

PR #2287, opened 2023-03-01
Wayland: XDG Activation protocol (implements glfwRequestWindowAttention)

GamesTrap:
This PR implements the `_glfwRequestWindowAttentionWayland()` function using the Wayland [xdg-activation-v1](https://wayland.app/protocols/xdg-activation-v1) protocol.

Tested on Kubuntu 22.10

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -202,6 +202,7 @@ video tutorials.
  - Riku Salminen
  - Brandon Schaefer
  - Sebastian Schuberth
+ - Jan Schuerkamp
  - Christian Sdunek
  - Matt Sealey
  - Steve Sexton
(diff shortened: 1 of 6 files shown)


GamesTrap:
In theory this protocol also allows us to implement `_glfwFocusWindowWayland()` (see [SDL Wayland_RaiseWindow()](https://github1s.com/libsdl-org/SDL/blob/HEAD/src/video/wayland/SDL_waylandwindow.c#L1450)).

In practice this didn't work for me on KDE so I left this one out.

elmindreda:
This looks good and works well, but it relies on what seems to be a relatively new protocol. Before merging it's necessary to consider how to deal with build compatibility with older systems, even if that ends up being "do nothing".

I'm not saying you need to solve this, but it needs to be addressed and I would welcome your input.

GamesTrap:
The protocol is already in testing/staging phase so it shouldn't be long until it gets into stable.
As far as I know the protocol is already supported on GNOME (Mutter), KDE (KWin) and Sway.

elmindreda:
Thank you for the PR! This has been merged as 93179702436b060aee0c2e28e5a681d72bf0d9ec, with slight modifications to adapt it to the current solution for the problem discussed above.

PR #2291, opened 2023-03-08
Joystick hat: treat invalid combinations as nothing pressed

PeterJohnson:
Previously this condition asserted, but on win32 it's possible to get opposing directions set simultaneously with some xinput controllers (due to the logic in win32_joystick.c:730-737).


elmindreda:
Thank you for reporting this! Out of curiosity, do you know of any specific device that causes it?

PeterJohnson:
It appears to happen with just a regular xbox controller if you roll around the dpad fast enough.

elmindreda:
Thank you for the bug report and PR! I never managed to reproduce this but I don't doubt it happened. It should hopefully be fixed now with 52405a9d599412545efd3bc3ace8342ae571ccd0, a slight modification of your solution.

PR #2293, opened 2023-03-09
Set MacOS Dock icon and query optimal icon size

### LINKED ISSUE #2041: Set macOS Dock icon

ws909:
Resolves #2041 
Replaces #2244 

The changes in this PR update `glfwSetWindowIcon` to allow a `NULL` window handle. When `NULL` is passed, GLFW sets the application's icon rather than a window's icon. An implementation for this is provided on MacOS.

This PR also updates `glfwSetWindowIcon` to allow querying for an optimal icon size. This requires an internal const-cast for the `images` parameter. This is unfortunate, but is required for backwards-compatibility. This preserves the signature of the function. This is currently only supported on MacOS.

Some platforms consume icons of more than one size at once. If GLFW is updated to return the optimal size for these platforms, the largest size should be returned. This is a limitation that is better revisited in GLFW 4.0.

This PR is more in line with #2286 and #2312, and should be the preferred choice over #2244.

Tested on MacOS with the `icon` and a new `animate-icon` test programs.

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -1531,11 +1531,11 @@ void _glfwSetWindowIconWin32(_GLFWwindow* window, int count, const GLFWimage* im
     if (count)
     {
         const GLFWimage* bigImage = _glfwChooseImage(count, images,
-                                                GetSystemMetrics(SM_CXICON),
-                                                GetSystemMetrics(SM_CYICON));
+                                                     GetSystemMetrics(SM_CXICON),
+                                                     GetSystemMetrics(SM_CYICON));
         const GLFWimage* smallImage = _glfwChooseImage(count, images,
-                                                  GetSystemMetrics(SM_CXSMICON),
-                                                  GetSystemMetrics(SM_CYSMICON));
+                                                       GetSystemMetrics(SM_CXSMICON),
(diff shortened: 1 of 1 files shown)


ws909:
@elmindreda Considering that you have assigned yourself to this, is this something that could get included in 3.4?

Displee:
I waited for this a long time now. Can this please be merged?

ws909:
Strikes me that a future enhancement to this new feature could include also caching the application icon for all platforms and then automatically set that as the window icon when new windows are created, on Win32, X11 and WL.

PR #2306, opened 2023-03-28
Don't crash when setting icon or getting focus on Wayland

orowith2os:
Required for Minecraft on Wayland to work with modified GLFW. Upstreamed from https://github.com/Admicos/minecraft-wayland/pull/46.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -2112,8 +2112,7 @@ void _glfwSetWindowTitleWayland(_GLFWwindow* window, const char* title)
 void _glfwSetWindowIconWayland(_GLFWwindow* window,
                                int count, const GLFWimage* images)
 {
-    _glfwInputError(GLFW_FEATURE_UNAVAILABLE,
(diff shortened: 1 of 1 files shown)


ws909:
If this change is essential to make Minecraft run, you need to maintain a separate fork of GLFW, compile GLFW yourself, and swap LWJGL's-provided native GLFW library with the library you compiled.

orowith2os:
> If this change is essential to make Minecraft run, you need to maintain a separate fork of GLFW, compile GLFW yourself, and swap LWJGL's-provided native GLFW library with the library you compiled.

That's what was already being done with the minecraft-wayland repo.

> Despite not having any actual right to make decisions here, I'd like to say that this is not going to get merged. _glfwInputError does not in itself cause any crash; it only sets the GLFW error message, and calls the registered error handler for GLFW. If Minecraft's error handler crashes on errors, you will have to contact Mojang for this issue.

Good to know. Ig I'll close this then, I'll just fork glfw directly and maintain that rather than updating the patch files in the future.

PR #2307, opened 2023-03-30
Win32 - Fetch more accurate monitor name

Scr3amer:
Hello guys,

This is my first PR for GLFW so I'm pretty sure I did a lot of awful things.

**For now take it just as a draft.**

Purpose of this PR: **Find if possible the actual name of the monitor instead of Generic PnP** that we often have.
I got the code by shamelessly copying and modifying the one from SDL2. (Thank you SDL guys)

**Main breaking change: Requires Windows 7 (I had to modify WINVER)**

I only did the Windows MinGW version because that's what I use and where I felt the need. I have no idea how accurate are the names returned on other platforms.

Anyway I'm getting ahead of myself.
First lets get some feedback on that:

- Coding convention
- Naming of the function
- Better way to do it ?
- Obvious bug I may have missed ?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -282,7 +282,7 @@ endif()
 # win32_platform.h.  We define them here until a saner solution can be found
 # NOTE: MinGW-w64 and Visual C++ do /not/ need this hack.
 if (MINGW)
-    target_compile_definitions(glfw PRIVATE WINVER=0x0601)
+    target_compile_definitions(glfw PRIVATE WINVER=0x0501)
 endif()
 
 # Workaround for legacy MinGW not providing XInput and DirectInput
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -282,7 +282,7 @@ endif()
 # win32_platform.h.  We define them here until a saner solution can be found
 # NOTE: MinGW-w64 and Visual C++ do /not/ need this hack.
 if (MINGW)
-    target_compile_definitions(glfw PRIVATE WINVER=0x0501)
+    target_compile_definitions(glfw PRIVATE WINVER=0x0601)
 endif()
 
 # Workaround for legacy MinGW not providing XInput and DirectInput
--- a/src/win32_monitor.c
+++ b/src/win32_monitor.c
@@ -37,6 +37,83 @@
 #include <wchar.h>
 
 
+// If the returned pointer is valid (not NULL) the caller of this function is in charge of freeing the memory when he is done.
+static char * tryGetMoreAccurateDisplayName(const WCHAR *deviceName)
+{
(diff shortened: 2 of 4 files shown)


ws909:
> Main breaking change: Requires Windows 7 (I had to modify WINVER)

GLFW 3 needs to support XP (for, ehm, reasons). You’ll need to conditionally compile it, or, the better option: check for availability at runtime, and fall back to the old solution if it’s not supported.

Scr3amer:
Thanks for the quick answer @ws909. I don't see how I can do runtime dispatch as I need some structs and includes that won't exist in XP.

Am I missing something ? Or does that mean I will have to take option 1: Conditional compiling.

Also:
- How should I call the cmake variable ? GLFW_USE_ENHANCED_MONITOR_NAME ?
- And how do I make the appveyor succeed ? My guess is that it uses GCC 5.3.0 and probably some old MinGW hence the fact it doesn't find dwmapi.h.

ws909:
> I don't see how I can do runtime dispatch as I need some structs and includes that won't exist in XP.

If they're opaque, you don't need the structure implementation. If they're transparent, you'll need to define them if they're not defined in the API. See [here](https://github.com/glfw/glfw/pull/2286/files#diff-9bc0a18c0222efe01b05444cfb5130b55fc08d0812de6b5146d6a6b208c8920f). Macro if/ifdef might work? 

> How should I call the cmake variable ? GLFW_USE_ENHANCED_MONITOR_NAME?

Any particular reason why this has to be configurable by the user? Any reason why the old behaviour must be retained?

> And how do I make the appveyor succeed ? My guess is that it uses GCC 5.3.0 and probably some old MinGW hence the fact it doesn't find dwmapi.h.

Without reading into the error message at the moment, the tests are supposed to make sure that GLFW compiles on all platforms supported by GLFW. It might be failing simply because you changed the lower requirement from XP to 7.

Scr3amer:
Thank you for the various answers.

- They are transparent so I will update the MR with a dynamic check see how this goes.

> Any particular reason why this has to be configurable by the user? Any reason why the old behaviour must be retained?
- You proposed this option if I can't do dynamic check when you said "conditionally compile". So I started thinking how I would expose it in CMake. Old behaviour must be retained for Windows XP compatibility and with this flag user could opt in to have better monitor name display but would require at least Windows 7. But now that you explained to me how to do the dynamic check this becomes pointless.

> `getName`, `getMonitorName`, `getAccurateName`, `queryDirectName`, etc.

I like getAccurateName. As a reminder the current public API is **glfwGetMonitorName** so I would propose getAccurateMonitorName or tryGetAccurateMonitorName for this internal function.

Scr3amer:
I am going to redo a proper PR because I realised I was using GLFW 3.3 instead of latest. I am going to close this one.

PR #2312, opened 2023-04-05
Application Window badge (closes #2248)

### LINKED ISSUE #2248: Set application icon badge
This is a feature request to add support for application icon badges, such as a message label on a Dock icon in MacOS, on the taskbar icon in Windows, and application icon in Wayland.

- The implementation on MacOS is trivial. This is 1 call: `NSApp.dockTile.badgeLabel = (NSString) ...`
- The implementation on Windows is [more complicated](https://learn.microsoft.com/nb-no/windows/win32/api/shobjidl_core/nf-shobjidl_core-itaskbarlist3-setoverlayicon?redirectedfrom=MSDN). It's per window, and requires being drawn in advance.
- Wayland: unknown.
- X11: unknown.

If some platforms support strings (MacOS, Windows), while others only support numbers, GLFW should support both. Can emit `GLFW_FEATURE_UNAVAILABLE` for strings on platforms that don't support it. 

I can provide an implementation for MacOS, as well as empty stubs for all platforms, <strike>and likely also a Windows implementation sometime in the future.</strike>

MacOS Dock icon badge label set to `"5"` for the Messages application:
Skjermbilde 2023-01-08 kl  01 32 19

A possible implementation is similar to `glfwRequestUserAttention`, that takes a window parameter, and sets it for the whole application on MacOS, but I don't think this is necessarily a good idea, because if you want different labels on different windows, previous ones will be overriden by newer ones. Might not be an issue, though.

GamesTrap:
**Note**: This commit is based on the branch used in:
- https://github.com/glfw/glfw/pull/2286
----
This is my implementation for window badge support, which would resolve:
- #2248.

# What's new

- `glfwSetWindowBadge()` function to display the badge containing a number in the dock or taskbar.
- `glfwSetWindowBadgeString()` function to display the badge containing a string in the dock or taskbar.
- Extended the tests/window program.

# Platform differences

## Windows

On Windows the GDI API is used to generate the 16x16 badge icon at run time.
The [ITaskbarList3 interface](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-itaskbarlist3) is used to set the badge in the taskbar which is available since Windows 7.

Window hint `GLFW_WIN32_GENERIC_BADGE` to replace the badge icon used in `glfwSetWindowBadge()` with a generic one which doesn't contain the given count (this currently looks like the Discord badge icon).

`glfwSetWindowBadgeString()` is unsupported because you could realistically only display at most 4 characters before the badge becomes completely unreadable so instead `GLFW_FEATURE_UNAVAILABLE` will be emitted.

## Linux

On Linux the [Unity LauncherAPI](https://wiki.ubuntu.com/Unity/LauncherAPI) is used via DBus. A valid application .desktop file with the same name as the executable is required for the API to work. As far as I know only KDE and the Unity DE use this API.

For the DBus communication I added an implementation which loads the needed functions at runtime from the `libdbus-1.so.3` shared library, see `posix_dbus.h` and `posix_dbus.c`.

`glfwSetWindowBadgeString()` function is not supported by the Unity LauncherAPI so the functions emits an `GLFW_FEATURE_UNAVAILABLE` error instead.

The badge can only be set for the whole application and not per window, thus the window parameter of the `glfwSetWindowBadge()` and `glfwSetWindowBadgeString()` functions must be `NULL` otherwise `GLFW_FEATURE_UNAVAILABLE` error will be emitted.

## macOS

On macOS the [NSDockTile class](https://developer.apple.com/documentation/appkit/nsdocktile?language=objc) is used to set the application badge in the dock which is available since macOS 10.5.

The badge can only be set for the whole application and not per window, thus the window parameter of the `glfwSetWindowBadge()` and `glfwSetWindowBadgeString()` functions must be `NULL` otherwise `GLFW_FEATURE_UNAVAILABLE` error will be emitted.

# Implementation status

- [x] Windows (supported since Windows 7, Thanks to @GyrosGeier for helping with GDI Transparency)
- [x] Linux (X11 & Wayland)
- [x] macOS (supported since macOS 10.5, Thanks to @ws909 for the implementation)

# Tested Platforms

As of [7b9df90](https://github.com/GamesTrap/glfw/commit/7b9df90be28810f5dbf6cbf28391bbabd753757a)

- [x] Kubuntu 22.10 (X11 & Wayland)
- [x] Kubuntu 22.04 LTS (X11 & Wayland)
- [x] Windows 11 Pro 22H2 (Visual Studio 2022)
- [ ] Windows 10 Pro 22H2 (Visual Studio 2019)
- [ ] Windows 8.1 (Visual Studio 2013)
- [ ] Windows 7
- [ ] Windows Vista Ultimate (Visual Studio 2010)
- [ ] Windows XP
- [x] macOS (by @ws909)


src/posix_dbus.c
```diff
@@ -0,0 +1,326 @@
...
+    if(!dbus_connection_send(_glfw.dbus.connection, msg, &serial))
+        _glfwInputError(GLFW_PLATFORM_ERROR, "Failed to send DBus message");
+    else
+        dbus_connection_flush(_glfw.dbus.connection);
+
+    //Free the message
+    dbus_message_unref(msg);
+}
+
+void _glfwUpdateBadgeDBusPOSIX(dbus_bool_t badgeVisible, int badgeCount)
```
ws909, line 237: These are two long, almost identical functions. Shouldn't this be one function, with a few parameters for each use-case?

src/posix_dbus.c
```diff
@@ -0,0 +1,326 @@
...
+    unsigned int exeNameLength = (exeNameEnd - lastFound) - 1;
+
+    //Create our final desktop file uri
+    unsigned int desktopFileLength = strlen("application://") + exeNameLength + strlen(".desktop") + 1;
+    char desktopFile[desktopFileLength];
+    memset(desktopFile, 0, sizeof(char) * desktopFileLength);
+    strcpy(desktopFile, "application://");
+    memcpy(desktopFile + strlen("application://"), lastFound + 1, exeNameLength);
+    strcpy(desktopFile + strlen("application://") + (exeNameLength), ".desktop");
+    desktopFile[desktopFileLength - 1] = '\0';
```
ws909, line 183: Is it necessary to build the `desktopFile` on every invocation? Can this be cached in `_glfw` (lazy loading)?

GamesTrap: We can lazy load it.
Should we load it during `_glfwInitDBusPOSIX()` or check on each invocation of `_glfwUpdateTaskbarProgressDBusPOSIX()` and `_glfwUpdateBadgeDBusPOSIX()` if its already loaded?

ws909: `_glfwInitDBusPOSIX()` sounds fine.

GamesTrap:
I've updated the PR to the current master branch commit https://github.com/glfw/glfw/commit/b35641f4a3c62aa86a0b3c983d163bc0fe36026d

PR #2327, opened 2023-04-22
Fix null pointer dereference

szsam:
Add null check for return values of _glfwPlatformGetTls().

### DIFF
--- a/src/egl_context.c
+++ b/src/egl_context.c
@@ -312,7 +312,7 @@ static GLFWglproc getProcAddressEGL(const char* procname)
 {
     _GLFWwindow* window = _glfwPlatformGetTls(&_glfw.contextSlot);
 
-    if (window->context.egl.client)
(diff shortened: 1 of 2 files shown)


akulpillai:
Hi! I’m a student researcher working with @szsam. We have identified this patch as one for a security vulnerability. 

Will a CVE be issued for this?

elmindreda:
Thank you for reporting this! The shared code is supposed to guarantee that the TLS slot is set correctly before calling these functions but I found one case where it could technically fail. This should be fixed now with fa0ea5c41f53119c68748a0e8570ef120b7fa703.

I've also added assertions to document this requirement with 93d70cb6a8c9571b5b493936596633a1a8538cea and removed one case of duplicate checks.

@akulpillai I have no plans to issue a CVE for this.

szsam:
Close in favor of the above two commits.

PR #2354, opened 2023-06-12
Better joystick detection for Linux

iacore:
cocoa_joystick.m seem to have access to USB data as well. However, I don't know what language that is.


iacore:
About the CI Appveyor test... is it really C20?

iacore:
@elmindreda

PR #2356, opened 2023-06-16
cmake: Update minimum to 3.6

juan-lunarg:
As of CMake 3.27:

Compatibility with versions of CMake older than 3.5 is now deprecated and will be removed from a future version.

Calls to cmake_minimum_required() or cmake_policy() that set the policy version to an older value now issue a deprecation diagnostic.

https://cmake.org/cmake/help/latest/release/3.27.html#deprecated-and-removed-features

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4...3.20 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.6...3.20 FATAL_ERROR)
 
 project(GLFW VERSION 3.4.0 LANGUAGES C)
 


juan-lunarg:
Closing since this PR since it isn't needed to fix the policy issue.

GLFW's usage of `cmake_minimum_required(VERSION 3.6...3.20 FATAL_ERROR)` already handles the policy issue in question.

juan-lunarg:
Feel free to ignore.

While going through this exercise with google-test they raised their CMake minimum to 3.13.

https://github.com/google/googletest
https://github.com/google/googletest/commits/main/CMakeLists.txt

So if a library as popular as googletest can raise their min to 3.13 I think GLFW can as well. Again this is just food for thought.

PR #2358, opened 2023-06-20
Draw only

ChoHag:
nb. This change is against the 3.3.8 tag which was installed on my system not the latest.

This is the preliminary work to allow GLFW to be used for its excellent GL context creation facilities and all of the effort which has gone into the mind-bogglingly complex task of ... opening a window*, but without it insisting on taking over responsibility for anything else.

For the moment I have only considered X. This change works for me and is required to get the level of control over input events that I need. The drag and drop changes were put in because the flag was right there while I was searching for input stuff to change.

Before putting in the work to clean this up to a releasable state I'm presenting it warts and all to get input on whether it's wanted and if so what it should look like.

[*] It sounds like it but that's not sarcasm. It really is an insane labrynth of decades old temporary fixes and worse.


ChoHag:
I was expecting a fairly obvious "compare this change against revision X" button here since although I tried to create the PR against the version I'd changed that button took me out of the new PR submission page. C'est la vie. If people who have the patience to work out yet another GUI know how to change this PR then please do so.

PR #2361, opened 2023-06-29
Wayland: don't crash if no cursor theme is available

AndreRenaud:
On some embedded systems there is no cursor theme set, which results in a crash in this function when we call `wl_cursor_theme_get_cursor`.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -2630,6 +2630,12 @@ GLFWbool _glfwCreateStandardCursorWayland(_GLFWcursor* cursor, int shape)
 {
     const char* name = NULL;
 
+    if (!_glfw.wl.cursorTheme) {
+        _glfwInputError(GLFW_CURSOR_UNAVAILABLE,
+                        "Wayland: No cursor theme");
+        return GLFW_FALSE;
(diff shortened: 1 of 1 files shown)


dougbinks:
Thanks for this PR.

One slightly confusing aspect is that GLFW should have failed the initialization as `_glfwInitWayland` calls https://github.com/glfw/glfw/blob/master/src/wl_init.c#L806 which would fail due to https://github.com/glfw/glfw/blob/master/src/wl_init.c#L367-L373

Did you check the return from `glfwInit`?

We do think that if there are sensibly configured machines with no theme then this shouldn't be a failing issue, but then we also need to fix the `glfwInit` return.

AndreRenaud:
It looks like you're correct - I was using this indirectly via a 3rd party library which may have been doing initialisation incorrectly. I was debugging the direct crash, and possibly missed the earlier cause. I'll close this and double check the other library.

Sorry about that.

dougbinks:
Thanks. If initialization is failing because of a lack of theme, and if the machine is correctly configured, then we should potentially change to handling no theme without failing.

PR #2382, opened 2023-08-21
Allow to set initial surface swap interval

dendy:
After creating surface context in _glfwPlatformCreateWindow function proceeds with reading context parameters in _glfwRefreshContextAttribs, which ends with clearing the front buffer when double buffering is enabled.

Such buffer clearing implicitly calls swapBuffer, which may not behave properly with the default swap interval. And user has no control to override this swap interval value with the glfwSwapInterval, because latter may only be called after the glfwCreateWindow, when it is too late.

Either clearing buffer should be disablable, or user should have control on the initial swap interval value before implicit swapBuffer is invoked.

This patch implements latter by introducing GLFW_SWAP_INTERVAL window hint. Default value is -1, which matches current policy of avoiding to override swap interval.

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -939,6 +939,8 @@ extern "C" {
  */
 #define GLFW_POSITION_Y             0x0002000F
 
+#define GLFW_SWAP_INTERVAL          0x00020010
+
 /*! @brief Framebuffer bit depth hint.
  *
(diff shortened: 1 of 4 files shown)


dendy:
This commit contains functional part only. Will amend it with the proper documentation once the whole idea about it will be accepted.

dendy:
@elmindreda Could you please take a look whether this feature makes sense to you?

dendy:
Hello, may I receive a review for this patch, please?

PR #2386, opened 2023-08-24
Fix pkg-config on Windows

moritz-h:
If building on Windows as shared library GLFW adds the suffix "dll" to the library name. This suffix is missing in the pkg-config file.


dougbinks:
Many thanks for this PR. I'll update the changelog and credits in a bit.

dougbinks:
Note: I added you to the credits as `moritz-h` as I could not find your full name. Let me know if you want that changed.

moritz-h:
> Note: I added you to the credits as `moritz-h` as I could not find your full name. Let me know if you want that changed.

@dougbinks Thanks. My full name is `Moritz Heinemann`. Seems most people in that file are listed by real name, so maybe fits better :smiley:

dougbinks:
Thanks, I've updated the contributers with your full name.

PR #2387, opened 2023-09-03
Remove apparently spurious line in header

eesctc:
When compiling glfw on a windows MSVC based system I get a warning about this line:
glfw\deps\glad/vulkan.h(6250,12): warning C4551: function call missing argument list 

The warning seems to be valid and removing the line appears to me to be safe.


eesctc:
I've done a little more reading around and it looks like this line will prevent warnings in other compilers or situations so I'm closing it.

PR #2392, opened 2023-09-10
AmigaOS4 support

afxgroup:
AmigaOS4 platform specific implementation.
It supports everything and it has been tested in several projetcs. 
Hope you can accept the request


dougbinks:
Many thanks for this PR. Adding a new platform is an impressive amount of work. I was fond of the Amiga, along with the 68K processor.

Sadly we can't accept this PR since we are unable to maintain it.

PR #2398, opened 2023-09-12
X11:: Add native access to X11 visual and XCB connection

YaaZ:
Exposed two new native handles, which are necessary for manually creating a Vulkan surface on X11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,8 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+ - Added `glfwGetX11Visual` and `glfwGetXCBConnection`
+   native functions for accessing X11 visual and XCB connection handle
  - Added `GLFW_PLATFORM` init hint for runtime platform selection (#1958)
  - Added `GLFW_ANY_PLATFORM`, `GLFW_PLATFORM_WIN32`, `GLFW_PLATFORM_COCOA`,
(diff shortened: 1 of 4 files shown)


elmindreda:
Thank you for this PR.

This looks potentially worth adding but why create the surface manually in the first place and not via `glfwCreateWindowSurface`? Is there a feature or extension missing from it?

YaaZ:
In my case I just wanted to speed up the startup by initializing Vulkan and GLFW in parallel, so to avoid redundant synchronization and initialization overhead I just did all Vulkan-related stuff myself.

pknowles:
The problem in general is there's API overlap, plumbing different native handles through to vulkan. Either `glfwCreateWindowSurface()` needs the app's vulkan context to work or the app needs access to the native handles.

At first I found `glfwCreateWindowSurface()` tried to initialize vulkan itself when I already have my own loader. Now, looking harder, I see there's `glfwInitVulkanLoader()` that would allow it to fetch vulkan functions from the app's loader (though only once, no re-init support). This is certainly an option for most apps.

In my case I'd like to use `GLFW/glfw3native.h` and allow apps to manage their own loader table(s) without having to share them as a global with GLFW. Other handles exist already and XCB is the missing bit.

FYI, this is my current shim to provide XCB handles. It's messy and I'd love to have these same calls in GLFW.
https://github.com/pknowles/vulkan_objects/blob/main/src/glfw_xcb_hack.cpp

YaaZ:
@elmindreda do you want to take another look at this?

PR #2404, opened 2023-09-26
Add a function to get the monitor on which a window is currently being displayed

Scr3amer:
Hello there,

Me again with another PR.

This time it is to add a function to retrieve the monitor on which a window is currently being displayed.

**I only did null and win32**

PRs are welcome for Linux and Mac OS.

Also I need some feedback on the naming documentation and such.

Cheers,
Scr3am

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -3898,6 +3898,24 @@ GLFWAPI void glfwFocusWindow(GLFWwindow* window);
  */
 GLFWAPI void glfwRequestWindowAttention(GLFWwindow* window);
 
+/*! @brief Returns the monitor on which the window is being currently displayed.
+ *
+ *  The window should not be NULL.
+ *
+ *  @param[in] window The window to query.
+ *
+ *  @errors Possible errors include @ref GLFW_NOT_INITIALIZED.
+ *
+ *  @thread_safety This function must only be called from the main thread.
+ *
+ *  @sa @ref window_monitor
+ *
+ *  @since Added in version 3.4.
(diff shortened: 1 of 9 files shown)


dougbinks:
One consideration is that it is possible to get the monitor on which a window is currently being displayed by calculating the area a window covers and comparing this to the `glfwGetMonitorWorkarea`, returning the largest value if there is an overlap or either the nearest or primary if there is non.

I'll look into whether the cross platform approach would be best as a fall back implementation or if more 'platform correct' features exist.

One thing we may want to consider is the following:

- Should the monitor returned correspond to the monitor which the window will fill if maximised? This is potentially not the same as the monitor the window covers the most.
- What should be returned for a iconified window?

Scr3amer:
Hey good to see you :)) !

I think your proposal is actually the PR I linked in my description.
I thought using native API would always be better than custom logic by default. But sure it is always cool to have a fallback for OSes that don't provide equivalent feature.

I'll double check the iconified behaviour tomorrow if God wills. I would love the API to return the monitor it will be displayed on if maximised OR nullptr.

Also IIRC win32 api provides an ENUM for the way to handle edge cases : 

> MONITOR_DEFAULTTONEAREST | Returns a handle to the display monitor that is nearest to the window.
> MONITOR_DEFAULTTONULL | Returns NULL.
> MONITOR_DEFAULTTOPRIMARY | Returns a handle to the primary display monitor.

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-monitorfromwindow

Issue #2377, opened 2023-08-02
Fullscreen window minimizes when clicked on 10.15 Catalina


FuzzyQuils:
OS and version: MacOS Catalina 10.15.7
Commit ID: 3eaf1255b29fdf5c2895856c7be7d7185ef2b241
Error Messages: None

(NOTE: This also happens on 3.3 stable) 

In my game engine, when it creates a GLFW window, it does so in windowed mode to start with in case loading the main resource archive takes a long time. I then set the window fullscreen later if configured to do so, otherwise it stays windowed.

On MacOS Catalina however, even with GLFW_MOUSE_PASSTHROUGH set to false, (I don't set the hint in the first place but I tested this to check) the mouse "clicks-through" the window and minimizes it if the engine starts in fullscreen. This state remains even if I then bring the window up again from the dock. Sometimes if I use my game engine's fullscreen toggle a couple of times it correctly focuses the window.

Zooming the window with the green button instead of using glfwSetWindowMonitor is not affected. Windowed mode is not affected. 

Big Sur and Monterey to my knowledge are not affected by this bug.

Gegy:
Hello! We have been investigating this issue due to this report on the Minecraft bug tracker: [MC-265519](https://bugs.mojang.com/browse/MC-265519). Not totally sure what has gone wrong here, but wanted to share our findings 🙂 
* It does not seem to be specifically related to the version of MacOS, but rather based on monitor configuration / screen resolution - it is possible to reproduce or not reproduce the issue on the same device with different screen or resolution setups
* The issue was introduced at some point between version 3.3.6 and 3.3.8 (suspicion is [this commit](https://github.com/glfw/glfw/commit/0d599026d04ddcc2695ae938fc1864523dab7a34), but have not confirmed this)
* The [NSWindowStyleMaskResizable](https://developer.apple.com/documentation/appkit/nswindowstylemask/nswindowstylemaskresizable) style flag seems to be related - explicitly removing it when in fullscreen mode seems to restore to previous behaviour
  * The resizable flag does actually allow the window to be resized even when in fullscreen mode, which is a bit of a fun experience 😄

FuzzyQuills:
> Hello! We have been investigating this issue due to this report on the Minecraft bug tracker: [MC-265519](https://bugs.mojang.com/browse/MC-265519). Not totally sure what has gone wrong here, but wanted to share our findings 🙂
> 
[...]

PR #2405, opened 2023-09-30
Cocoa: Make sure NSWindowStyleMaskResizable is masked when entering fullscreen.

FuzzyQuills:
On macOS Catalina and earlier, not having the resizable bit cleared in NSWindowStyleMask in fullscreen leads to windows minimising when clicked anywhere in the content area. This affects both my game engine and reportedly Minecraft, as detailed in issue #2377

This fix doesn't (and shouldn't need to) apply to Windows created in a fullscreen state as createNativeWindow in cocoa_window.m defines a default styleMask before altering it according to desired GLFW window hints and settings.

This hasn't been tested on newer macOS versions for regressions but should otherwise work there.

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -1279,7 +1279,7 @@ void _glfwSetWindowMonitorCocoa(_GLFWwindow* window,
 
     if (window->monitor)
     {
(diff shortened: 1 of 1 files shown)


elmindreda (approved):
Looks good!

Issue #90, opened 2013-07-10
Need a way to do pinch-to-zoom


pcwalton:
For Servo we need a way to use pinch-to-zoom. The `NSView` method we need bindings to is `magnifyWithEvent:`.

nippoo:
Any update on this? (@elmindreda). I don't know how stable/ready the touch branch is? It would be wonderful to have some level of multi-touch support, or at least expose pinch-to-zoom (even better, two-finger rotate etc) on devices that support it...

elmindreda:
@nippoo Sorry, no progress here. I'm currently swamped by unrelated issues and non-GLFW standards work.

danielkeller:
The code to do this is actually pretty straightforward. I don't have an actual patch, but you can see my implementation in the first few files of [this diff](https://github.com/danielkeller/Violet/commit/e3ff39ff726fe1f594fc3f08848fcbcdc4a5bee2).

Arcnor:
Is there any way the `touch` branch can be merged or @danielkeller changes can be added? Some kind of multi touch support is always better than no support at all, I think.

jadahl:
FWIW, in libinput and Wayland, the touchpad pinch gesture API/protocol provides both the pinch amount (how far away the fingers are from each other), a x/y motion (how the center fo the pinch moved), and the pinch rotation. If the GLFW API is going to be able to support more than just the first, either all these values should be passed to the callback, or some more complex event structure passed to the callback is needed.

Arcnor:
Can we at least start by adding the pointer index to the cursor position callback and functions? That way, even without proper gesture support you can do your own gesture recognition if you want

elmindreda:
@Arcnor No, that would break the ABI.

Arcnor:
@elmindreda ok, can we add another callback then? I don't know what the proper procedure is in this case, but getting access to the different pointers is enough for my use case, and sort of the minimum thing to have...

elmindreda:
@Arcnor That is what the `touch` branch does, which I'm hoping there'll be time for in 3.2.

Arcnor:
@elmindreda Didn't realize that. I'll eagerly wait for 3.2 then :smile:

jadahl:
It is not expected to be able to recieve individual touch points from
touchpads right? Support for that is something we ruled out as unreasonable
in Wayland, with the except a "raw touchpad" mode for things such as
chinese character writing.

Arcnor:
[...]

PR #2419, opened 2023-10-22
Add macOS events for trackpad zoom and rotate gestures

pfgithub:
Implements #90 for macOS with two new events: TrackpadZoom and TrackpadRotate. These events can also be implemented for Windows and Linux X11/Wayland, but I do not have a way to test those platforms. Zoom will need to make sure to use a similar scale on all platforms and rotate should use degrees.

- [x]    Change log entry in README.md, listing all new symbols
- [x]    News page entry, briefly describing the feature
- [ ]    Guide documentation, with minimal examples, in the relevant guide
- [ ]    Reference documentation, with all applicable tags
- [ ]    Cross-references and mentions in appropriate places
- [ ]    Credits entries for all authors of the feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,8 @@ information on what to include when reporting a bug.
    application (#2110)
  - [Cocoa] Bugfix: The Vulkan loader was not loaded from the `Frameworks` bundle
    subdirectory (#2113,#2120)
+ - [Cocoa] Added `glfwSetTrackpadZoomCallback` and `glfwSetTrackpadRotateCallback`
+   for trackpad zoom and rotate events (#90)
  - [X11] Bugfix: The CMake files did not check for the XInput headers (#1480)
  - [X11] Bugfix: Key names were not updated when the keyboard layout changed
    (#1462,#1528)
--- a/docs/news.dox
+++ b/docs/news.dox
@@ -223,6 +223,14 @@ the XDG-Shell protocol.  If your Wayland compositor does not support XDG-Shell
 then GLFW will fail to initialize.
 
 
+@subsubsection macos_trackpad_34 Support for trackpad zoom and rotate on macOS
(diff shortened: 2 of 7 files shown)


cblc:
Please, merge this 🙏 🙏 🙏  I really want Magic Trackpad gestures in my apps...

pfgithub:
I added wayland support, but I'm not really sure where to call `_glfwAddPointerGesturesListeners` from. The current place works it seems.

pfgithub:
This is done. I can't add Windows or X11 support because I can't test them, but if someone wants to here's information:

<details><summary>Windows</summary>

Add in `src/win32_window` in the switch with all the other events like WM_MOUSEWHEEL.

Firefox has an example of using these events: https://searchfox.org/mozilla-central/rev/f2c181a7ab3bfea4d2266521e6eac713630479b3/widget/windows/nsWinGesture.cpp#142-204

You might need to set up gesture config, as described above in that file.

Make sure that the resulting value sent to `_glfwInputTrackpadZoom` is a scale factor, and `_glfwInputTrackpadRotate` receives this event's angle minus the previous events angle in degrees centered at 0 (-180 to 180 rather than 0 to 360)

Make sure that zooming doesn't emit ctrl + scroll events once implemented. I'm not sure if it does this currently, but there was a reference to that in the windows documentation.

</details>

<details><summary>X11</summary>

Example from Gdk: https://gitlab.gnome.org/GNOME/gtk/-/blob/58c49ce6b044c1b08e099cb232f06e71365c5232/gdk/x11/gdkdevicemanager-xi2.c?page=2#L1907-1963

</details>

zacharypepin:
Any updates on this?

mooch443:
*bump* Would be great to have this merged!

PR #2420, opened 2023-10-24
Add the suffix of dll

jimwang118:
The pkg-config files installed with GLFW refers to "-lglfw3", but the installed lib file is called lib/glfw3dll.lib. GLFW seems to use a suffix for building a shared lib, see https://github.com/glfw/glfw/blob/3.3.8/src/CMakeLists.txt#L169
Fix the inconsistency between the library name and the link library name in the pc file when dynamically compiling on the window platform.


moritz-h:
Already tried to fix the same in #2386

dougbinks:
I've merged #2386 which resolves this, but since you both submitted a PR I've added you both to the credits.

PR #2423, opened 2023-10-31
Allow disabling the mouse button limit

jedenastka:
It's possible for at least the Wayland and X11 platforms to report mouse button codes above the 8 named by GLFW. There is however an arbitrary limitation in place that causes them to never be reported to the API consumer. This change removes that limitation.

The button state array is handled the same way the key state array is handled, and the way it's defined in the documentation - it contains state only for named buttons. Other buttons need to be received via a callback.

It seems like this change is compatible with existing code. GLFW documentation doesn't define `GLFW_MOUSE_BUTTON_LAST` as a hard limit of the reported `button` value, but as the last named mouse button.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -284,6 +284,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Grzesiek11
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
--- a/docs/input.dox
+++ b/docs/input.dox
@@ -502,8 +502,8 @@ initialization hint to `GLFW_FALSE`.
 glfwInitHint(GLFW_MOUSE_BUTTON_LIMIT, GLFW_FALSE);
 @endcode
 
-The callback function receives the mouse button, button action and
-[modifier bits](@ref mods).
+The callback function receives the [mouse button](@ref buttons), button action
+and [modifier bits](@ref mods).
 
 @code
 void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
--- a/docs/input.dox
+++ b/docs/input.dox
@@ -494,8 +494,9 @@ a mouse button callback.
 glfwSetMouseButtonCallback(window, mouse_button_callback);
 @endcode
 
-To handle all mouse buttons, instead of only the [supported ones](@ref buttons),
-set the @ref GLFW_MOUSE_BUTTON_LIMIT initialization hint to `GLFW_FALSE`.
+To handle all mouse buttons, instead of only ones with associated
+[button tokens](@ref buttons), set the @ref GLFW_MOUSE_BUTTON_LIMIT
+initialization hint to `GLFW_FALSE`.
 
 @code
 glfwInitHint(GLFW_MOUSE_BUTTON_LIMIT, GLFW_FALSE);
@@ -513,12 +514,12 @@ void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
 @endcode
 
 The mouse button is an integer that can be one of the
-[supported mouse buttons](@ref buttons) or, if @ref GLFW_MOUSE_BUTTON_LIMIT is
+[mouse button tokens](@ref buttons) or, if @ref GLFW_MOUSE_BUTTON_LIMIT is
 set to `GLFW_FALSE`, any other value.
 
 The action is one of `GLFW_PRESS` or `GLFW_RELEASE`.
 
-The last reported state for every [supported mouse button](@ref buttons) is also
+The last reported state for every [mouse button token](@ref buttons) is also
 saved in per-window state arrays that can be polled with @ref
 glfwGetMouseButton.
 
@@ -553,7 +554,7 @@ had been processed in the meantime, the state will reset to `GLFW_RELEASE`,
 otherwise it will remain `GLFW_PRESS`.
 
 The `GLFW_MOUSE_BUTTON_LAST` constant holds the highest value of any
-[supported mouse button](@ref buttons).
+[mouse button token](@ref buttons).
 
 
 @subsection scrolling Scroll input
--- a/docs/input.dox
+++ b/docs/input.dox
@@ -494,8 +494,15 @@ a mouse button callback.
 glfwSetMouseButtonCallback(window, mouse_button_callback);
 @endcode
 
-The callback function receives the [mouse button](@ref buttons), button action
-and [modifier bits](@ref mods).
+To handle all mouse buttons, instead of only the [supported ones](@ref buttons),
+set the @ref GLFW_MOUSE_BUTTON_LIMIT initialization hint to `GLFW_FALSE`.
+
+@code
+glfwInitHint(GLFW_MOUSE_BUTTON_LIMIT, GLFW_FALSE);
+@endcode
+
+The callback function receives the mouse button, button action and
+[modifier bits](@ref mods).
 
 @code
 void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
@@ -504,6 +511,10 @@ void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
         popup_menu();
 }
 @endcode
+
+The mouse button is an integer that can be one of the
+[supported mouse buttons](@ref buttons) or, if @ref GLFW_MOUSE_BUTTON_LIMIT is
+set to `GLFW_FALSE`, any other value.
 
 The action is one of `GLFW_PRESS` or `GLFW_RELEASE`.
 
--- a/docs/input.dox
+++ b/docs/input.dox
@@ -494,20 +494,19 @@ a mouse button callback.
 glfwSetMouseButtonCallback(window, mouse_button_callback);
 @endcode
 
-@anchor GLFW_MOUSE_BUTTON_LIMIT
+@anchor GLFW_DISABLE_MOUSE_BUTTON_LIMIT
 To handle all mouse buttons, instead of only ones with associated
-[button tokens](@ref buttons), set the @ref GLFW_MOUSE_BUTTON_LIMIT
-input mode to `GLFW_FALSE`.
+[button tokens](@ref buttons), set the @ref GLFW_DISABLE_MOUSE_BUTTON_LIMIT
+input mode.
 
 @code
-glfwSetInputMode(window, GLFW_MOUSE_BUTTON_LIMIT, GLFW_FALSE);
+glfwSetInputMode(window, GLFW_DISABLE_MOUSE_BUTTON_LIMIT, GLFW_TRUE);
 @endcode
 
-This input mode specifies whether to limit reported mouse buttons to only those
-that have an associated button token, for compatibility with earlier versions of
-GLFW that would only report buttons with an associated button token: users may
-have assumed that reported buttons always are in the range of
-@ref GLFW_MOUSE_BUTTON_1 to @ref GLFW_MOUSE_BUTTON_LAST.
+When this input mode is enabled, GLFW doesn't limit the reported mouse buttons
+to only those that have an associated button token, for compatibility with
+earlier versions of GLFW, which never reported any buttons over
+@ref GLFW_MOUSE_BUTTON_LAST, on which users could have relied on.
 
 The callback function receives the [mouse button](@ref buttons), button action
 and [modifier bits](@ref mods).
@@ -521,8 +520,8 @@ void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
 @endcode
 
 The mouse button is an integer that can be one of the
-[mouse button tokens](@ref buttons) or, if @ref GLFW_MOUSE_BUTTON_LIMIT is
-set to `GLFW_FALSE`, any other value.
+[mouse button tokens](@ref buttons) or, if the
+@ref GLFW_DISABLE_MOUSE_BUTTON_LIMIT input mode is set, any other value.
 
 The action is one of `GLFW_PRESS` or `GLFW_RELEASE`.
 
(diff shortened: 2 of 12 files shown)


dougbinks:
Many thanks for this PR. The code for this looks very reasonable, however I think we need some extra documentation to go along with this.

In particular the following comment on `glfwSetMouseButtonCallback` is not true for buttons beyond `GLFW_MOUSE_BUTTON_LAST`:

```
 *  When a window loses input focus, it will generate synthetic mouse button
 *  release events for all pressed mouse buttons.  You can tell these events
 *  from user-generated events by the fact that the synthetic ones are generated
 *  after the focus loss event has been processed, i.e. after the
 *  [window focus callback](@ref glfwSetWindowFocusCallback) has been called.
```

Additionally although the documentation suggests that events might have button presses for none named buttons it's not explicitly stated, which I think we should.

If you'd like to have a go at the documentation please do, otherwise I can handle it.

One potential concern I do have is that developers may have assumed that the largest mouse button is `GLFW_MOUSE_BUTTON_LAST` and implemented their own state arrays up to that value without checks. This change would break their code. If we want to avoid that then we would need an initialization hint to control permitting button events > `GLFW_MOUSE_BUTTON_LAST`. I am currently not sure if that is required - I think @elmindreda should make that call.

Alternatively to this solution, we could move the button array to a 32bit bitfield (without changing the external API), supporting up to 32 buttons and improving the performance of some button test checks. However I think that change could be done after this PR was integrated.

jedenastka:
Thanks for taking a look at this.

> In particular the following comment on `glfwSetMouseButtonCallback` is not true for buttons beyond `GLFW_MOUSE_BUTTON_LAST`: [...]

That's true, so I thought I would take a look at the keyboard input documentation for reference. After all, the synthetic key release events aren't generated for non-named keys either. Here's what I found:

> When a window loses input focus, it will generate synthetic key release events for all pressed keys.

It's not correct there either. Fixing it here and not there would be weird, so I'm going to make a separate MR for the keyboard part and then change it here as well.

> Additionally although the documentation suggests that events might have button presses for none named buttons it's not explicitly stated, which I think we should.

Good idea.

> One potential concern I do have is that developers may have assumed that the largest mouse button is `GLFW_MOUSE_BUTTON_LAST` and implemented their own state arrays up to that value without checks. This change would break their code.

Yeah, I have thought that might be the case. After all technically non-breaking changes can still break code.

jedenastka:
I opened #2433 which will fix the documentation for synthetic key and button releases.

dougbinks:
Given this effectively changes the API we need an init hint similar to the `GLFW_JOYSTICK_HAT_BUTTONS` initialization hint.

https://www.glfw.org/docs/3.3/intro_guide.html#init_hints

If you don't have time to add this let me know and I can do the work.

include/GLFW/glfw3.h
```diff
@@ -5188,11 +5193,14 @@ GLFWAPI GLFWcharmodsfun glfwSetCharModsCallback(GLFWwindow* window, GLFWcharmods
...
- *  release events for all pressed mouse buttons.  You can tell these events
- *  from user-generated events by the fact that the synthetic ones are generated
- *  after the focus loss event has been processed, i.e. after the
+ *  release events for all pressed named mouse buttons.  You can tell these
+ *  events from user-generated events by the fact that the synthetic ones are
+ *  generated after the focus loss event has been processed, i.e. after the
  *  [window focus callback](@ref glfwSetWindowFocusCallback) has been called.
  *
+ *  The reported `button` value can be higher than `GLFW_MOUSE_BUTTON_LAST` if
+ *  the button is not a [named button](@ref buttons).
```
dougbinks, line 5202: add "and GLFW_MOUSE_BUTTON_LIMIT init hint is set to GLFW_FALSE"

dougbinks:
Thanks for the changes - I've noticed one line I think needs to be added to the documentation.

jedenastka:
The commit message should have been "unset" but whatever.

dougbinks:
Many thanks for this work.

Apologies for not spotting this earlier but going over the input functionality it looks like another option for setting the `GLFW_MOUSE_BUTTON_LIMIT` might be via [glfwSetInputMode](https://www.glfw.org/docs/3.3/group__input.html#gaa92336e173da9c8834558b54ee80563b) as this is somewhat similar to how `GLFW_LOCK_KEY_MODS` operates. What do you think? If we want to use `glfwSetInputMode` I can do the work if needed.

Also, `named keys` were renamed to `key tokens` in [Improve documentation relating to key tokens](https://github.com/glfw/glfw/commit/9959dc69cadd1f7790060846ffbc44e7da56fc56) on 5th Dec. If we want to keep the same naming style we could move from `named mouse buttons` to `mouse button tokens`. However we can move to the required naming later but would appreciate your thoughts on the mouse button term.

In completing this PR we will also need the steps as [Contributing a feature](https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-feature), namely:

 - Change log entry in README.md, listing all new symbols. *We need to add the new `GLFW_MOUSE_BUTTON_LIMIT` symbol and brief explanation.*
 - [News page](https://github.com/glfw/glfw/blob/master/docs/news.dox) entry, briefly describing the feature. *TODO*
 - Guide documentation, with minimal examples, in the relevant guide. *Need to edit the Mouse button input section of docs/input.dox*. See [the current Mouse button input docs](https://www.glfw.org/docs/3.3/input_guide.html#input_mouse_button)
 - Reference documentation, with all applicable tags. *Done for the current status, but if we switch to glfwSetInputMode needs reworking*
 - Cross-references and mentions in appropriate places. *Done for the current status, but if we switch to glfwSetInputMode needs reworking*
 - Credits entries for all authors of the feature. *DONE*

I think we should also probably modify [events.c](https://github.com/glfw/glfw/blob/master/tests/events.c) to set `GLFW_MOUSE_BUTTON_LIMIT` off so it receives all events for mouse buttons.

Once again I can do the work if you don't have the time.

jedenastka:
> Apologies for not spotting this earlier but going over the input functionality it looks like another option for setting the `GLFW_MOUSE_BUTTON_LIMIT` might be via [glfwSetInputMode](https://www.glfw.org/docs/3.3/group__input.html#gaa92336e173da9c8834558b54ee80563b) as this is somewhat similar to how `GLFW_LOCK_KEY_MODS` operates. What do you think? If we want to use `glfwSetInputMode` I can do the work if needed.

As I'm not an active contributor, I don't know for sure when the one or the other is a better choice.

Trying to analyze it, the practical difference seems to be that init hints affect a particular window and not the whole program. The semantic difference on the other hand seems to be that input modes are used for changing *preferences* regarding input, where init hints are overall *changes to the behavior* of the library.

When taking this into account, I'd say making this an init hint makes more sense. It's not really a preference, more a difference between the old and new behaviour.

> Also, `named keys` were renamed to `key tokens` in [Improve documentation relating to key tokens](https://github.com/glfw/glfw/commit/9959dc69cadd1f7790060846ffbc44e7da56fc56) on 5th Dec. If we want to keep the same naming style we could move from `named mouse buttons` to `mouse button tokens`. However we can move to the required naming later but would appreciate your thoughts on the mouse button term.

I noticed that there were some discrepancies in the naming while working on #2417, so it's good that this has been taken care of. I indeed think that it'd be better for the naming to be consistent between keyboard and mouse input.

> In completing this PR we will also need the steps as [Contributing a feature](https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-feature), namely:

This was originally supposed to be classified as a bugfix, that's why I didn't do that before 😅. But yeah, now that it has documentation and new symbols it's pretty much a feature.

> * Change log entry in README.md, listing all new symbols. _We need to add the new `GLFW_MOUSE_BUTTON_LIMIT` symbol and brief explanation._
> 
> * [News page](https://github.com/glfw/glfw/blob/master/docs/news.dox) entry, briefly describing the feature. _TODO_
> 
> * Guide documentation, with minimal examples, in the relevant guide. _Need to edit the Mouse button input section of docs/input.dox_. See [the current Mouse button input docs](https://www.glfw.org/docs/3.3/input_guide.html#input_mouse_button)

Gonna work on these now, assuming the current usage of init hints. If it's decided that input modes are better, the change can be a single commit with both code and documentation changes.

jedenastka:
Looking at the documentation for mouse buttons to try improving it make me notice the version on my branch is outdated, so I took a second look at 9959dc6:

> The associated term 'named mouse button' was also replaced with
> 'supported mouse button'.

Seems like a change is already in effect. I still think using the same term for mouse buttons and keyboard keys would be a good idea though. For now I'll rebase and update to use the term "supported" though.

jedenastka:
I'd like some feedback on my wording in c025d54c. I'm not a native speaker and sometimes have trouble deciding if what I wrote is understandable.

While adding information to the input guide I also wanted to add that the mouse button values that are not supported (or named, or don't have tokens) are platform specific, like scancode values in case of keyboard input. But then I noticed that different platforms could also return different values for the buttons `GLFW_MOUSE_BUTTON_4` and over as well, unless this was tested (I tested X11, Wayland and Windows (by testing the Win32 API calls GLFW uses, my VM can't run GLFW programs), but I don't have access to a macOS enviroment). Not sure how to handle that properly...

dougbinks:
Some feedback:

## Mouse button naming

The 'supported mouse buttons' naming is misleading, we would prefer 'mouse buttons tokens' so the sentence `release events for all pressed supported mouse buttons` would become `release events for all pressed mouse buttons with mouse buttons tokens` Optionally we could be precise and add 

```
/*
* ...
* release events for all pressed mouse buttons with mouse buttons tokens (`GLFW_MOUSE_BUTTON_1` to
* `GLFW_MOUSE_BUTTON_LAST`).
* ...
*/
```

## GLFW_MOUSE_BUTTON_LIMIT
After communicating with elmindreda we think that *it may be worth considering making it an input mode instead of an init hint. It feels a little bit like GLFW_LOCK_KEY_MODS.*

So this would be part of [glfwSetInputMode](https://www.glfw.org/docs/latest/group__input.html#gaa92336e173da9c8834558b54ee80563b).
```
glfwSetInputMode(window,GLFW_MOUSE_BUTTON_LIMIT, GLFW_FALSE);
```

jedenastka:
Changed the term to 'mouse button tokens', it's indeed less confusing and I was able to word the docs better.

Where are initializations of the input modes located? They seemingly aren't in the `glfwCreateWindow` function, where I'd expect them to be (as they are part of the `_GLFWwindow` struct), and I can't find them by grepping the source or jumping around references with a language server.

jedenastka:
Done. I still think it feels out of place as an input mode though.

It might be more intuitive if it's purpose was inverted - instead of disabling the old handling, it'd enable the new one, and it'd be called `GLFW_DISABLE_MOUSE_BUTTON_LIMIT`.

dougbinks:
I concur. Do you want to change to using `GLFW_DISABLE_MOUSE_BUTTON_LIMIT`?

jedenastka:
Did that just now.

docs/input.dox
```diff
@@ -505,9 +519,13 @@ void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
 }
 @endcode
 
+The mouse button is an integer that can be one of the
+[mouse button tokens](@ref buttons) or, if the
+@ref GLFW_DISABLE_MOUSE_BUTTON_LIMIT input mode is set, any other value.
```
dougbinks, line 524: I think we need to change `any other value` to `any other positive value`.

CONTRIBUTORS.md
```diff
@@ -284,6 +284,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Grzesiek11
```
dougbinks, line 287: The contributors are normally alphabetic by last or only name, but I can fix that up after merge if needed.

docs/input.dox
```diff
@@ -494,6 +494,20 @@ a mouse button callback.
 glfwSetMouseButtonCallback(window, mouse_button_callback);
 @endcode
 
+@anchor GLFW_DISABLE_MOUSE_BUTTON_LIMIT
+To handle all mouse buttons, instead of only ones with associated
+[button tokens](@ref buttons), set the @ref GLFW_DISABLE_MOUSE_BUTTON_LIMIT
+input mode.
```
dougbinks, line 500: We need to make clear that this only works for the callback, and not for `glfwGetMouseButton`, with a sentence here, also in `glfw3.h` in the doc comments above `glfwSetInputMode` and `glfwGetMouseButton`.

dougbinks:
Thanks!

I've not done a full review but have added some comments.

I also note that we need to add `GLFW_DISABLE_MOUSE_BUTTON_LIMIT` to the comments above `glfwSetInputMode` in `glfw3.h`

CONTRIBUTORS.md
jedenastka: Looks like I'm already listed in the proper position, and this is just a duplicate coming from my error during merging back master.

docs/input.dox
jedenastka: I assumed this would be redundant, as this section of the documentation talks about the callback specifically, and when `glfwGetMouseButton` is mentioned later, it's explicitly said that it only works for named mouse buttons.

If you think clarifying it would be good regardless, I can do that.

dougbinks:
I have squash merged your commits into my branch:

https://github.com/glfw/glfw/compare/master...dougbinks:glfw:unlimited-mouse-buttons-merge

This way I can make any required minor changes on this branch and merge it to GLFW once complete.

jedenastka:
Sorry I didn't manage to find the time to do it. I was planning to do it this Friday...

dougbinks:
No problem - due to the release of GLFW 3.4 there was a merge which wasn't obvious, so I thought it best to handle this way. Hope that's OK!

dougbinks:
After some discussions we decided that the term `GLFW_UNLIMITED_MOUSE_BUTTONS` was preferable and also followed the pattern of the rest of the GLFW API, so I've changed over to using that and also altered any documentation to align with this.

https://github.com/dougbinks/glfw/compare/master...unlimited-mouse-buttons-merge

dougbinks:
This has now been merged into main.

Many thanks for your work on the PR, and your patience with my reviewing!

Issue #2427, opened 2023-11-05
CursorMode DISABLED on Windows gives wrong coordinates


Hilderin:
On Windows 11 i get wrong coordinates when using Disabled CursorMode. In RawMotion the coordinates increment at each mouse move and when RawMotion is disable, there's a lag before a get the mouse direction change.

I don't known if it could be related to https://github.com/glfw/glfw/issues/1790

I included some screenshots debugging the problem with the cursor.exe test project...

**CursorMode Disabled + RawMotion:**

Moving the mouse left:
![image](https://github.com/glfw/glfw/assets/81109165/a88e4fcb-db2f-4612-8fb2-4e65f01e45cf)

The problem seams caused by a recenter of the cursor in rawmotion.
I added a debug printf in _glfwPollEventsWin32:
![image](https://github.com/glfw/glfw/assets/81109165/11891e2d-cfb9-4d64-9a59-cf83c49cb060)

Add now when i move the mouse in Cursor Disabled mode and RawMotion, we can see the cursor is recentered at each mouse movement:
![image](https://github.com/glfw/glfw/assets/81109165/93cfdfb8-8510-41c3-903e-41d2091959fa)

If i comment the recentering in _glfwPollEventsWin32, it's seems to work perfectly:
![image](https://github.com/glfw/glfw/assets/81109165/b797a360-8449-4b89-9cbb-3c69e413e067)

**CursorMode Disabled + NoRawMotion:**
With the recentering commented, the direction changes are instant:
![image](https://github.com/glfw/glfw/assets/81109165/5bd792ad-acf2-47c3-99e1-b9d206ad6a05)

Hilderin:
Finally, it's the same issue as: https://github.com/glfw/glfw/issues/1276
I was working over Remote Desktop.

PR #2428, opened 2023-11-05
CursorMode DISABLED on Windows gives wrong coordinates

Hilderin:
A suggestion to fix the issuse: https://github.com/glfw/glfw/issues/2427

I just commented the recentering of the mouse position in _glfwPollEventsWin32 in cursor disabled mode. It's seems to work on my computer but i don't known if it could create a side effect for others.

**CursorMode Disabled + RawMotion:**

Moving the mouse left:
image

The problem seams caused by a recenter of the cursor in rawmotion.
I added a debug printf in _glfwPollEventsWin32:
image

Add now when i move the mouse in Cursor Disabled mode and RawMotion, we can see the cursor is recentered at each mouse movement:
image

If i comment the recentering in _glfwPollEventsWin32, it's seems to work perfectly:
image

**CursorMode Disabled + NoRawMotion:**
With the recentering commented, the direction changes are instant:
image

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -2096,20 +2096,22 @@ void _glfwPollEventsWin32(void)
         }
     }
 
-    window = _glfw.win32.disabledCursorWindow;
-    if (window)
-    {
-        int width, height;
-        _glfwGetWindowSizeWin32(window, &width, &height);
+    // Fixing the problem where recentering the cursor when the mouse move breaks the lastCursorPosX and lastCursorPosY
+    // and the coordinates sent in _glfwInputCursorPos are completly wrong.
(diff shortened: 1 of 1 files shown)


Hilderin:
I was wrong about the recenter in _glfwPollEventsWin32, it's super important because if the cursor is not recentered, the mouse will stop at the edge of the window.
After more testing the problem seems to be that even if the mouse is recentered, le raw positions are not and when the WIN_INPUT event is raised with MOUSE_MOVE_ABSOLUTE flag, we cannot calculate the offsets using lastCursorPosX and lastCursorPosY because they contains the center of the window but the new raw position have not moved with the recenter.

So, the real problem seems only using RawMotion.

I did not find a solution yet, working on it.

PR #2433, opened 2023-11-18
Specify that synthetic key releases are only generated for named keys

jedenastka:
The current wording states that all keys have synthetic key release events generated after focus is lost, but keys that aren't named don't have any state held, so no such events are generated for them.

This change clarifies that only named keys have the events generated for them.

Not sure if docs changes deserve a changelog entry, if so, it can be added.


dougbinks:
Good find, thanks for this.

PR #2449, opened 2023-12-21
Recommiting this submodule path?

Ice-Citron:


elmindreda:
@Ice-Citron It looks like some GLFW source files were not included in the build, specifically `platform.c` and `win32_module.c`. Those were both added since 3.3 so perhaps this is a case of an outdated build file?

PR #2453, opened 2023-12-24
Update boing.c

fatihaktas845:
Updating boing.c


elmindreda:
Thank you for this PR. I'm not sure if it's necessary to check for `_WIN32` first. Is the original test failing for you in any environment?

fatihaktas845:
The original test does not fail in any environment, but I did this to
(albeit slightly) speed up the compilation process. It's also my first
PR experience in software.


15 Ocak 2024 Pazartesi tarihinde Camilla Löwy ***@***.***>
yazdı:

elmindreda:
I'm sorry, I won't merge this PR. I don't think this change speeds up the compilation.

With this change the same number of checks are done on non-Win32 platforms, and an additional check is done on Win32. These checks are very fast because modern preprocessors are very efficient, so both version will perform the same even if this PR adds a few more lines to process.

If the condition controlled whether to include a huge header file and it was included unnecessarily on some platforms then that would be be worth doing, but this is about whether to define a single macro.

PR #2455, opened 2023-12-29
Fix joystick EV_KEY handling indexing below 0 on keyboard input

ptrToLiam:
This fixes an issue I was running into where input from my keyboard running ZMK firmware would be handled as a joystick, and would index the associated keymap with a value less than 0 (`code - BTN_MISC`, where `code` was smaller than `BTN_MISC`).

This issue was causing `SIGILL` crashes on my laptop every time I used that particular keyboard within my programs using glfw.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -278,6 +278,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Liam Malone
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
--- a/README.md
+++ b/README.md
@@ -121,6 +121,8 @@ information on what to include when reporting a bug.
 
 ## Changelog
 
+ - Bugfix: Linux joystick handling would crash with some keyboards, indexing the 
+   associated KeyMap with a value less than 0
  - Added `GLFW_PLATFORM` init hint for runtime platform selection (#1958)
  - Added `GLFW_ANY_PLATFORM`, `GLFW_PLATFORM_WIN32`, `GLFW_PLATFORM_COCOA`,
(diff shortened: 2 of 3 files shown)


elmindreda:
Thank you for the bug report! It seems the core issue is that GLFW is treating a keyboard like a joystick in the first place?

Liam-Malone:
Hi, yes the core issue is my keyboard being treated both as a keyboard and as a joystick.

Liam-Malone:
I will note: 
I only did the band-aid fix of preventing indexing below 0, my keyboard still registers as both gamepad and keyboard. This was enough of a fix to permit me to keep working on my own project, but doesn't resolve the core issue.

I have also since tested with another keyboard running QMK firmware and found NO issues with that keyboard. So I can say again that this issue is only occurring on the keyboard that has ZMK firmware, when connected to my laptop.

kangzixiang (approved):

tristanbatchler:
I can confirm this PR fixes my repro when applied to raylib 6.0's bundled GLFW.

Setup:

```text
Linux 6.18.26-gentoo-dist x86_64
Session: Wayland/labwc, with Xwayland running
Zig: 0.16.0
raylib-zig: devel#58f3c0fa328fc9ff48d17c2c3771ab9c5114aaa6
raylib runtime backend: DESKTOP (GLFW - X11)
Keyboard: Keychron Q3 Max over Bluetooth
```

Before the patch, temporary logging in `src/linux_joystick.c` showed:

```text
GLFW joystick EV_KEY: code=28 BTN_MISC=256 index=-228 value=0 name=Keychron Q3 Max Keyboard
```

and the program crashed at:

```c
js->linjs.keyMap[code - BTN_MISC]
```

I applied this PR's patch to raylib 6.0's bundled GLFW. The `src/linux_joystick.c` hunk applied cleanly, and after rebuilding, `InitWindow` completed and the program exited cleanly.

The README hunk no longer applied, but the code fix did.

PR #2457, opened 2023-12-31
Wayland: add possibility to disable scaling

manio:
Currently wayland displays can have some scalling applied (it is widely used with HiDPI displays).
It is then handled via a `viewporter` or `fractional-scale` protocols.
However it is sometimes desirable to explicitly disable scaling for some GLFW application to have a clear and perfect 1:1 pixel window content (ie. no scaling applied).

This PR adds such a possiblity.
To use it one can eg. set/export a variable like this:
`export GLFW_SCALE_TO_MONITOR=1`
and then start the application.

### DIFF
--- a/src/wl_init.c
+++ b/src/wl_init.c
@@ -366,6 +366,15 @@ static void createKeyTables(void)
     }
 }
 
+static GLFWbool getScaleSetting(void)
+{
+    const char* scaleEnvString = getenv("GLFW_SCALE_TO_MONITOR");
+    if (scaleEnvString)
+        return GLFW_TRUE;
+    else
+        return GLFW_FALSE;
+}
+
 static GLFWbool loadCursorTheme(void)
 {
     int cursorSize = 16;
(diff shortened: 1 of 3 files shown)


elmindreda:
With a9cc7c7260c32068a5374ce9d59515df540de970 there is now the `GLFW_SCALE_FRAMEBUFFER` window hint that lets you control this scaling per-window on Wayland. Is that sufficient for your needs?

manio:
I'll try to test this ASAP...

elmindreda:
> However it is sometimes desirable to explicitly disable scaling for some GLFW application to have a clear and perfect 1:1 pixel window content (ie. no scaling applied).

I think this is the issue.

You seem to be referring to the scaling the compositor performs on the window, over which we have no control.

The `GLFW_SCALE_FRAMEBUFFER` refers to scaling the framebuffer relative to the window size, so that the window contents (typically) maps 1:1 to output pixels again even if window sizes no longer do.

I think you want to set the hint to true.

elmindreda:
This PR essentially adds an option that switches the behaviors of `glfwGetWindowSize` and `glfwGetFramebufferSize`. It's not changing the actual size of the window or framebuffer, just what values are retrieved via these functions.

If the changes in this PR makes your program draw correctly, then it is likely using the wrong one of these functions somewhere.

manio:
@AlexandreRouma
Could you please comment on this? Are you able to adjust `sdrpp` scalling in the app itself instead of glfw ?

AlexandreRouma:
SDR++ bases the scaling on glfwGetWindowSize()

PR #2462, opened 2024-01-06
Add glfw3impl.h, a single header implementation of GLFW for C/C++

garettbass:
- assigned `_glfw`-qualified private names to all `static` symbols to avoid naming conflicts when including `glfw3impl.h`

- added casts as needed to compile `glfw3impl.h` as C++

- replaced `sprintf` with `snprintf` to avoid warnings from `clang -Wall`, (except on win32, where this provokes an appveyor failure `'snprintf' undefined`) 

- Added a brief remark on `glfw3impl.h` to the `Compiling GLFW` section in `README.md`

- Added `Garett Bass` to `CONTRIBUTORS.md`

I understand I should additionally provide updates to `docs/news.dox`, an example, and some means of testing that header-only compilation for C or C++ is not broken by future changes, which I will add in an additional commit if this change is deemed technically acceptable.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -14,6 +14,7 @@ video tutorials.
  - Luca Bacci
  - Keith Bauer
  - John Bartholomew
+ - Garett Bass
  - Coşku Baş
  - Niklas Behrens
  - Andrew Belt
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ available for all supported compilers.
 See the [compilation guide](https://www.glfw.org/docs/latest/compile.html) for
 more information about how to compile GLFW yourself.
 
+Alternatively, you can `#include <GLFW/glfw3impl.h>` in exactly one compilation
+unit of a C or C++ project to compile the entire implementation directly into
+your application, avoiding the need to separately compile and link with GLFW.
+
 
 ## Using GLFW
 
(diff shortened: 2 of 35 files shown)


Clownacy:
I thought this project was written in C89? snprintf is C99. Also, identifiers with a leading underscore are [reserved](https://stackoverflow.com/questions/69084726/what-are-the-rules-about-using-an-underscore-in-a-c-identifier) and should not be used.

dougbinks:
Can I ask why this PR was closed?

garettbass:
I was discouraged by the response and lost interest in this project.

dougbinks:
I'm sorry you were discouraged - I haven't had time to check in depth but the basic idea of this PR is potentially useful.

LocalSpook:
> I thought this project was written in C89?

Only those parts that have to be compatible with Visual Studio 2010 (C99 is allowed elsewhere).

M374LX:
Rebase: #2890

As identifiers with a leading underscore are reserved, a possible better idea is to use the ``glfw_`` prefix followed by a lowercase letter. This would make it easy to distinguish public (e.g., ``glfwInit()``) from private (e.g., ``glfw_acquireMonitorX11()``).

As GLFW already uses the ``_glfw`` prefix for many private names, this should be a topic for another issue or PR.

PR #2466, opened 2024-01-11
Liz3/colored titlebar

liz3:


liz3:
Apologies! this was never meant as a pull request to upstream

PR #2469, opened 2024-01-17
Fix compilation of Null-only library on Linux

DziubanMaciej:
In file news.dox it is mentioned that it should be possible to disable both X11 and Wayland backends to compile GLFW with only null platform.

This actually does not compile due to missing definintion of clockid_t and CLOCK_REALTIME. This commit fixes compilation errors.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -316,10 +316,8 @@ if (MSVC90)
 endif()
 
 # Workaround for -std=c99 on Linux disabling _DEFAULT_SOURCE (POSIX 2008 and more)
-if (GLFW_BUILD_X11 OR GLFW_BUILD_WAYLAND)
-    if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
-        target_compile_definitions(glfw PRIVATE _DEFAULT_SOURCE)
-    endif()
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    target_compile_definitions(glfw PRIVATE _DEFAULT_SOURCE)
 endif()
 
 if (GLFW_BUILD_SHARED_LIBRARY)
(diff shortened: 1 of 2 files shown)


elmindreda:
Thank you for the bug report. On which version of which Linux distribution does the build fail without this change?

DziubanMaciej:
Arch Linux. I haven't tried on other distros.

DziubanMaciej:
Hey, what is the status of this? I see it "Needs investigation". Does that mean that you guys want it fixed another way? Or do you just need to find time to verify that it really works?

elmindreda:
I don't doubt that the fix works. It means I have to set up an Arch VM to find out why it works and why it's needed. It may point to more work being needed elsewhere, or to some other solution, or to some additional thing to keep in mind for future code.

elmindreda:
The CMake half of this PR was already done a few months back by d9709904fec956524d4e9032a7fa8d0550d5d49b.

The other half is odd, since [the manpages](https://man.archlinux.org/man/core/man-pages/clockid_t.3type.en) for Arch [say that](https://man.archlinux.org/man/clock_gettime.2) including `time.h` should work. That is included [here](https://github.com/glfw/glfw/blob/1fe98a0d5388435720d380439497630ac0477455/src/posix_time.h#L31), and so is included for every compilation unit in GLFW.

DziubanMaciej:
To be honest, I'd have to jump in and re-debug this, because I don't remember the specifics about why that was needed. But latest glfw commit can be built fine with `-DGLFW_BUILD_X11=OFF -DGLFW_BUILD_WAYLAND=OFF`, so I guess that's isn't needed anymore.

Thanks for your time. I'll close this PR.

elmindreda:
Huh. Glad to hear it's working and thank you for checking again!

PR #2482, opened 2024-02-09
Ability to get a window's title

dougbinks:
This PR implements the feature requested by issue #1448, partially based on the implementation by #1909.

A design choice has been made to return the same title which was set, rather than the title which could be obtained from the platform. This reduces surface area of the code and is easier to reason about when making dynamic titles.


elmindreda (approved):
Looks good.

elmindreda:
This was merged as 95d464bb4b00c4a11683587591e3c90b3db9b0ed. Thank you!

PR #2491, opened 2024-02-20
Added `GLFW_ACCELERATION` window hint for hardware acceleration

### LINKED ISSUE #589: windows, support for non-accelerated pixel formats
I'm using glfw for simple render with OpenGL 1.1 support only
I am targeting for windows vm with no video-card installed (Microsoft basic display adapter)

I able to use my app if I create context manually, however it is not possible to use glfw because of check for acceleration:
https://github.com/glfw/glfw/blob/master/src/wgl_context.c#L210

I know it is very specific issue, could we ifdef + add cmake property for non-accelerated pf support?

thank you :)

SuperSodaSea:
- Add `GLFW_ACCELERATION` window hint to specify whether hardware acceleration is preferred or not. Default to `GLFW_TRUE`, only available on WGL currently.
- Add `GLFW_ACCELERATION` window attribute to retrieve whether hardware acceleration is enabled.
- Supports creating window when the Microsoft GDI software OpenGL ICD is the only available implementation by default.

Resolves #589.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -291,6 +291,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Tianlan Zhou
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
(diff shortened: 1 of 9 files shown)


HinTak:
Most of this already existed on https://github.com/glfw/glfw/compare/context-renderer - there should be a pull based on that? Would have saved me some time for mine too (#2571 )

HinTak:
Linked from #589

HinTak:
This seems to be equivalent to the context-renderer branch (which uses `--renderer=sw` instead)? Anyway, that branch is now inclusive (after some adjustment of 7 years of bitrot) in my #2571 .

Henrryxdjr (approved):

PR #2496, opened 2024-02-25
Wayland: Partially implement `glfwSetCursorPos`

Friz64:
Wayland _does_ support setting the cursor position with [`set_cursor_position_hint`](https://wayland.app/protocols/pointer-constraints-unstable-v1#zwp_locked_pointer_v1:request:set_cursor_position_hint), but in order to call that function we need a [`zwp_locked_pointer_v1`](https://wayland.app/protocols/pointer-constraints-unstable-v1#zwp_locked_pointer_v1) object.

By "partially implement" I mean that calling `glfwSetCursorPos` while the cursor mode is not DISABLED will do nothing immediately. I implemented a mechanism that defers the call to `set_cursor_position_hint` until we have locked the pointer, if the pointer is not currently locked. This has an effect in these two situations:

1. The application has already set the cursor mode to DISABLED, but the cursor is currently outside of the window.
2. The application has not yet set the cursor mode to DISABLED, but will (hopefully) do so soon.

In situation two we input an error that notifies the application about this.

This is everything required to make the use of `glfwSetCursorPos` in Minecraft work. When closing an GUI in Minecraft, it first sets the cursor position to the center, then sets the cursor mode to DISABLED immediately after. Because of this, Minecraft inadvertently makes use of this "defer" mechanism. Situation two applies. This does mean that the log is hit with an GLFW error every time an GUI is closed, but I don't think there's much that can be done to avoid this.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -2667,8 +2667,34 @@ void _glfwGetCursorPosWayland(_GLFWwindow* window, double* xpos, double* ypos)
 
 void _glfwSetCursorPosWayland(_GLFWwindow* window, double x, double y)
 {
-    _glfwInputError(GLFW_FEATURE_UNAVAILABLE,
-                    "Wayland: The platform does not support setting the cursor position");
+    if (!_glfw.wl.pointerConstraints)
+    {
+        _glfwInputError(GLFW_FEATURE_UNAVAILABLE,
+                        "Wayland: The compositor does not support setting the cursor position");
+        return;
+    }
+
+    if (window->wl.lockedPointer) {
+        zwp_locked_pointer_v1_set_cursor_position_hint(window->wl.lockedPointer,
+                                                       wl_fixed_from_double(x),
+                                                       wl_fixed_from_double(y));
+    } else {
+        if (window->cursorMode != GLFW_CURSOR_DISABLED) {
+            _glfwInputError(GLFW_PLATFORM_ERROR,
+                            "Wayland: Delaying the cursor position update until "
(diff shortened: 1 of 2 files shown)


src/wl_window.c
```diff
@@ -2667,8 +2667,34 @@ void _glfwGetCursorPosWayland(_GLFWwindow* window, double* xpos, double* ypos)
 
 void _glfwSetCursorPosWayland(_GLFWwindow* window, double x, double y)
 {
-    _glfwInputError(GLFW_FEATURE_UNAVAILABLE,
-                    "Wayland: The platform does not support setting the cursor position");
+    if (!_glfw.wl.pointerConstraints)
```
Friz64, line 2670: Side note: The `lockPointer` function makes also makes use of `pointer-constraints-unstable-v1` functions [here](https://github.com/Friz64/glfw/blob/2307384a83ab87ba454458db0579f0ca318151f8/src/wl_window.c#L2968), but only checks for `relative-pointer-unstable-v1`. Is that a bug?

Friz64:
This would need to be explained in the [documentation of `glfwSetCursorPos`](https://www.glfw.org/docs/latest/group__input.html#ga04b03af936d906ca123c8f4ee08b39e7).

I find it very curios that the documentation makes it seem like this has already been implemented, when that is not the case. Has this been implemented somewhere? If so, I can't find the code.

SzczurekYT:
A protocol for this was just recently merged I guess
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/337

PR #2509, opened 2024-03-10
Update null_monitor.c

eltociear:
comment minor fix


dougbinks:
This appears to be automated or semi-automated Github stat padding, with a large number of similar spelling correction PRs from the same user, so we will not be accepting this.

PR #2525, opened 2024-03-30
fix: disable dinput8 to fix startup delay

zao:
Some HID devices block DirectInput8 device enumeration for many seconds which adversely affects software that uses GLFW without needing joysticks at all.

This change vendors the build of GLFW3 with a patch that skips loading of the `dinput8` library and thus disables all functionality that checks for its availability.


zao:
Sorry, meant to PR a different repo indirectly using this patch, please disregard.

PR #2568, opened 2024-05-30
Wayland: Fix key repeat

Cydox:
Key repeat shoud only be interrupted when the repeating key is released, not when another key is released.

This matches the behavior on X11 (and also on win32 if my memory serves me right).

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -1796,10 +1796,11 @@ static void keyboardHandleKey(void* userData,
 
             timer.it_value.tv_sec = _glfw.wl.keyRepeatDelay / 1000;
             timer.it_value.tv_nsec = (_glfw.wl.keyRepeatDelay % 1000) * 1000000;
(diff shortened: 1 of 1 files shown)


dougbinks:
I can confirm that both Windows and X11 will repeat the last key pressed even after you release a previously pressed key.

Currently GLFW Wayland will stop all repeat events when any key is released, this PR aligns the key repeats with the behaviour of GLFW Wayland with that on Windows and X11 (and probably MacOS but I haven't tested that yet).

I've reviewed and tested the code, it looks good so I will shortly merge it and add you to the contributers.

dougbinks:
I have merged this with https://github.com/glfw/glfw/commit/768e81a0eb3ae411d108168fdff7cd3335f2a34a

I added a readme and contributors entry as per our [contributing a bug fix guidelines](https://github.com/Cydox/glfw/blob/fix-wayland-keyrepeat/docs/CONTRIBUTING.md#contributing-a-bug-fix). Please let me know if you want me to change the name details (I used your name from your Github profile).

Many thanks for finding the issue and creating this PR!

Issue #2570, opened 2024-06-01
Compatibility of glfw with the Apple software render in github CI


HinTak:
Github CI has a mac os runner which is headless AFAIK, but it  has a software renderer. However, glfw can't get to it - this can be illustrates by this pull (do not merge) which runs glfwinfo after build:
https://github.com/glfw/glfw/pull/2569

Both OSX OpenGL (the GLUT part) and SDL2 can get at the apple software renderer within github CI, so this is a glfw specific problem. It would be useful to nake this work, to allow glfw to be used within github's own CI workflow for testing.

I came to be here after going down the rabbit hole of trying to look at a mac os specific problem in skia-python (which has CI testing for all three platforms, and been using glut for testing on mac os x for the above reasons, instead of glfw as in the case of linux with Xvfb):
https://github.com/kyamagu/skia-python/issues/214 
and downstream https://github.com/FlorianRhiem/pyGLFW/issues/80

HinTak:
That sdl2 and opengl on osx works happily headless can been seen in the CI in https://github.com/FlorianRhiem/pyGLFW/issues/80#issuecomment-2143191281

dougbinks:
#2080 is related

HinTak:
For a successful run (with either SDL2 or GLUT) within github CI's macos 14 runner, the GL context reports its vendor/renderer, version, shader version as:

```
b'Apple Inc.'
b'Apple Software Renderer'
b'4.1 APPLE-21.0.19'
b'4.10'
```

HinTak:
Glfw itself has an offscreen rendering example : https://github.com/glfw/glfw/blob/master/examples/offscreen.c . It probably should be run post-build to check that it can run!

HinTak:
In my fork's ci branch I have added the two lines from #2080, as well as running both glfwinfo and the offscreen example. Here is the result from github action:


```
Run ./build-cocoa-static/examples/offscreen
  ./build-cocoa-static/examples/offscreen
  shell: /bin/bash -e {0}
  env:
    CFLAGS: -Werror
    MACOSX_DEPLOYMENT_TARGET: 10.11
    CMAKE_OSX_ARCHITECTURES: x86_64;arm64
0s
Run ./build-cocoa-static/tests/glfwinfo
GLFW header version: 3.5.0
GLFW library version: 3.5.0
GLFW library version string: "3.5.0 Cocoa NSGL Null EGL OSMesa monotonic"
GLFW platform: Cocoa
[...]

Issue #1592, opened 2019-10-31
Support for EGLDevice (headless/no X server, off-screen rendering)


MadcowD:
Hey all!

I am currently working on a down-stream project in reinforcement learning and we're trying to modify an LWJGL 3 application linking against GLFW to run headless (no X server, using EGLDevice). After looking around quite a bit it looks like others have attempted to develop such an extension for GLFW, but development appears to be largely abandoned (https://github.com/glfw/glfw/pull/786, https://github.com/glfw/glfw/issues/648).

Our team does not have the background and is not well suited to implement this feature for GLFW, but we are deeply interested in seeing it finished. I am wondering if there was any plan for this feature going forward, and if not, if anyone would be willing to finish it for a bounty (i.e. payment for its completion!). 

If this is a possibility, please let me know!

William Guss
wguss@cs.cmu.edu
Carnegie Mellon University

cshesse:
You may want to check out https://github.com/glfw/glfw/pull/1608

MadcowD:
I am in a situation where an X server cannot be run!

This looks great by the way. I am happy to pay to get this piece polished and merged if that is something you're interested in.

cshesse:
Based on the other outstanding PRs, I don't have a lot of hope that the maintainers will be interested.  Feel free to try out that branch for your use case though.

6r1d:
I'd like to see this feature in GLFW, too.

princec:
Seems I'm stuck without this as well :(

HinTak:
I might be interested in finishing this. My interests is that glfw should just work within github's headless CI system: #2570 and gotten the mac os x's part to automatically fall back to offscreen/software renderer #2571 . I am surprised that glfw doesn't automatically run its own offscreen and glfwinfo examples after automated build...

HinTak:
#2573 I had a bit of investigation of the osmesa api. Glfwinfo doesn't work headless either , even with all three: xvfb-run, --renderer=sw, and --context-api=osmesa. So it is quite disappointing.

It fails with vulkan - but silently - i have to add a print statement to know that's what it is.- this makes it rather unsatisfactory as a diagnostic tool: it should work against xvfb-run without any special switches, and ideally, without needing that either, auto-drop back to headless mode.

[...]

PR #2571, opened 2024-06-04
Apple software renderer retry

### LINKED ISSUE #2080: allow MacOS OpenGL software  renderer?
Seems to work by simply replacing:

> ADD_ATTRIB(NSOpenGLPFAAccelerated);

with

> ADD_ATTRIB(NSOpenGLPFARendererID);ADD_ATTRIB(kCGLRendererGenericFloatID);

in src/nsgl_context.m.

An automatic fallback would be the best option.

HinTak:
Implement automatic software renderer re-try, and run glfwinfo post-build in the headless environment.

Fixes #2080 and #2570

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -1108,6 +1108,8 @@ extern "C" {
  *  [GLFW_SCALE_FRAMEBUFFER](@ref GLFW_SCALE_FRAMEBUFFER_hint) window hint for
  *  compatibility with earlier versions.
  */
+#define GLFW_CONTEXT_RENDERER       0x0002200E
+
 #define GLFW_COCOA_RETINA_FRAMEBUFFER 0x00023001
 /*! @brief macOS specific
  *  [window hint](@ref GLFW_COCOA_FRAME_NAME_hint).
@@ -1179,6 +1181,9 @@ extern "C" {
 
 #define GLFW_WAYLAND_PREFER_LIBDECOR    0x00038001
 #define GLFW_WAYLAND_DISABLE_LIBDECOR   0x00038002
+
+#define GLFW_HARDWARE_RENDERER      0x00039001
+#define GLFW_SOFTWARE_RENDERER      0x00039002
 
 #define GLFW_ANY_POSITION           0x80000000
 
--- a/src/glx_context.c
+++ b/src/glx_context.c
@@ -467,6 +467,9 @@ GLFWbool _glfwCreateContextGLX(_GLFWwindow* window,
         return GLFW_FALSE;
     }
 
+    if (ctxconfig->renderer == GLFW_SOFTWARE_RENDERER)
+        setenv("LIBGL_ALWAYS_SOFTWARE", "1", 1);
+
     if (ctxconfig->client == GLFW_OPENGL_ES_API)
     {
         if (!_glfw.glx.ARB_create_context ||
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,8 @@ jobs:
               run: cmake -B build-cocoa-static
             - name: Build Cocoa static library
               run: cmake --build build-cocoa-static --parallel
+            - run: ./build-cocoa-static/tests/glfwinfo
+            - run: ./build-cocoa-static/examples/offscreen
 
             - name: Configure Cocoa shared library
               run: cmake -B build-cocoa-shared -D BUILD_SHARED_LIBS=ON
--- a/README.md
+++ b/README.md
@@ -132,6 +132,8 @@ information on what to include when reporting a bug.
  - [Null] Added EGL context creation on Mesa via `EGL_MESA_platform_surfaceless`
  - [EGL] Allowed native access on Wayland with `GLFW_CONTEXT_CREATION_API` set to
    `GLFW_NATIVE_CONTEXT_API` (#2518)
+ - Added `GLFW_CONTEXT_RENDERER` window hint and `GLFW_HARDWARE_RENDERER` and
+   `GLFW_SOFTWARE_RENDERER` hint values (#589)
 
 
 ## Contact
--- a/docs/window.md
+++ b/docs/window.md
@@ -375,6 +375,11 @@ does not update the window contents when its buffers are swapped.  Use OpenGL
 functions or the OSMesa native access functions @ref glfwGetOSMesaColorBuffer
 and @ref glfwGetOSMesaDepthBuffer to retrieve the framebuffer contents.
 
+@anchor GLFW_CONTEXT_RENDERER_hint
+__GLFW_CONTEXT_RENDERER__ specifies whether to create the context using
+a hardware or software renderer, if that is possible to control on the current
+platform.
+
 @anchor GLFW_CONTEXT_VERSION_MAJOR_hint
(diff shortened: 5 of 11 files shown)


jonathanhogg:
If it's still useful to know, I can confirm that this works for me on macOS Apple Silicon.

It is *painfully* slow as the software renderer appears to use only a single core. Sadly it also seems to have a mesh render bug with a few of my examples.

WerWolv:
Correctly uses accelerated graphics for me on my M2 Macbook Air 👍

iTrooz:
I can confirm the exact same build as Werwolv runs in a virtual machine with the "Apple Software Renderer" renderer

src/glx_context.c
```diff
@@ -467,6 +467,9 @@ GLFWbool _glfwCreateContextGLX(_GLFWwindow* window,
         return GLFW_FALSE;
     }
 
+    if (ctxconfig->renderer == GLFW_SOFTWARE_RENDERER)
+        setenv("LIBGL_ALWAYS_SOFTWARE", "1", 1);
```
sleeptightAnsiC, line 471: There are multiple problems here:
1. Does not actually work (at leas not for me) as the environment variable is being set too late. If I move this to the beginning of `_glfwCreateWindowX11()` or set the variable at command line, it will pick the `llvmpipe` (AKA software renderer) just fine.
2. This is a hack that works only with MESA. `LIBGL_ALWAYS_SOFTWARE` is specific to MESA's implementation, see https://docs.mesa3d.org/envvars.html Although, these days a lot of linux/bsd distros should ship with MESA by default, if someone uses something like proprietary NVIDIA drivers, MESA might not be available (many wikies suggest uninstalling MESA packages in such scenarios).
3. If you're setting this to `"1"` when `renderer == GLFW_SOFTWARE_RENDERER`, you may also set it to `"0"` when `renderer == GLFW_HARDWARE_RENDERER` (this does work, MESA's documentation does not mention it explicitly, but I've tested it).
4. Calling `setenv( , , 1);` with `1` as the last argument means that it overwrites the enviroment variable. Generally, an application should not overwrite environment variables; if someone sets `LIBGL_ALWAYS_SOFTWARE=0` at command line (or through script, etc...) it would be unexpected that application suddenly ignores it (by overtiring it under the hood). This should probably be `setenv( , , 0);` (with `0` as last argument).
5. This hack is not specific to GLX, you can also use it for EGL and probably for any other platform/context supported by MESA (e.g. it should work for WGL that runs under WINE).

Sorry for quite big and scary list. I just tested this and wanted to give some feedback.

include/GLFW/glfw3.h
```diff
@@ -1180,6 +1182,9 @@ extern "C" {
 #define GLFW_WAYLAND_PREFER_LIBDECOR    0x00038001
 #define GLFW_WAYLAND_DISABLE_LIBDECOR   0x00038002
 
+#define GLFW_HARDWARE_RENDERER      0x00039001
+#define GLFW_SOFTWARE_RENDERER      0x00039002
+
```
sleeptightAnsiC, line 1187: I'm not sure, if choice of names "CONTEXT_RENDERER" and "HARDWARE/SOFTWARE_RENDER" is a good idea. "RENDERER" can really mean anything. Most of OpenGL/GLX/ELG/etc extensions refer to "Acceleration" when talking about using GPU for rendering (e.g. [GLX_MESA_query_renderer](https://github.com/KhronosGroup/OpenGL-Registry/blob/200cea4030cb49d3e40677379e6368a5f0e8c27b/extensions/MESA/GLX_MESA_query_renderer.txt#L120-L122) has `GLX_RENDERER_ACCELERATED_MESA`; there are also few other extensions that use similar name). Not sure about Vulkan though.

src/glx_context.c
HinTak: This is for Apple software renderer (headless Apple boxes) . Why do you keep referring to Mesa? Apple's own renderer called itself 'software' in the headless case. I did not invent that naming.

sleeptightAnsiC:
Thanks for PR @HinTak 

I'm using it for running my tiny GLFW project inside of VM with macOS Ventura 13.6.9 x86_64 and it works quite well there. I've recently started messing with setting `GLFW_CONTEXT_RENDERER` explicitly (and also on Linux) and found some issues which made me look into the code more carefully. I've left few notes here and there. Hope you don't mind.

HinTak:
I explained at the beginning that this is mainly about Apple github CI https://github.com/glfw/glfw/issues/2570 . I can't remember why I have WGL or EGL code in there. As for headless Linux, I have already got an alternative: https://github.com/kyamagu/skia-python/blob/17a97a0e547c42acac6d9d6e0ec91960685d89a9/.github/workflows/ci.yml#L215 .

There is a EGL issue (and even a stale EGL branch) here, I believe. Linux / Mesa related comments should go there instead.

src/glx_context.c
HinTak: Mesa related comments please go elsewhere.

HinTak:
There is a (stale) EGL branch https://github.com/glfw/glfw/tree/egldevice .

sleeptightAnsiC:
> this is mainly about Apple github CI https://github.com/glfw/glfw/issues/2570 . I can't remember why I have WGL or EGL code in there.

This is quite confusing... Not gonna lie 👀

Sorry, I simply looked at the diffs and since they show that PR touches way more code and adds more hints I though it's not only about what mentioned in the description (and  not only about Apple platform itself).

I believe I know what is the source our of confusion, see commit history:
Screenshot_20250608_050647
Screenshot_20250608_050727
I think you did merge `glfw/context-renderer` into this PR by mistake, right (?) It seems like this brings all the code that I was refering to in my comments.

Sorry once again for not noticing this in the first place. I'm not near my workstation at the moment (writing this from phone). I'll try to move my comments to proper PRs per your request, as soon as I can.

In the meantime could you take a look at the commits? It really seems like the old commit was merged into this by mistake.

HinTak:
Look, can you comment on the other issue instead? https://github.com/glfw/glfw/issues/1592 . I am going to block you if you keep making unrelated comments.

ColleagueRiley:
@HinTak, Your PR includes changes for more than just Apple Software Rendering. Maybe this was unintentional and a result of merging upstream, or maybe it accidentally pushes changes you didn't intend to be pushed?

Either way, it is important to review the changes before they are pulled. I suggest, if you can, simplifying the changes down to only the ones relevant to the PR (e.g., Apple Software Rendering). I hope that explains that confusion.

HinTak:
@ColleagueRiley it was intentional. Explained in #1592 . It is offensive to click on "review" when not qualified to do so. Please stop now.

ws909:
@sleeptightAnsiC 
Thanks for the feedback and reviews. It's quite helpful and is appreciated by the core maintainers/developers (not me) and the rest of the community (me, I guess?). All help offloading tasks is welcome. 

You're right that this PR touches much more code than that responsible for the issue it seems to be concerned with. That's a problem, and in its current state, it doesn't seem take the full set of use-cases for GLFW into consideration. In other words, it's not ready for integration into GLFW. 

Your feedback and comments are appropriate _here_. Everyone commenting here is qualified to do so.

CC @ColleagueRiley

HinTak:
Not interested in flame wars. Please stop now.

PR #2583, opened 2024-06-24
Wayland: let the compositor decide which screen to use for fullscreen operations

OlegAckbar:
Wayland doesn't have a concept of primary/preferred display. By default, GLFW assuming that first reported wl_output is the primary screen, but that's [not](https://bugs.kde.org/show_bug.cgi?id=489135) [true](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/179). I've made a simple workaround that unsets wl_output from `xdg_toplevel_set_fullscreen` requests. That'll make the compositor decide on which screen fullscreen application must go.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -503,12 +503,12 @@ static void acquireMonitor(_GLFWwindow* window)
     if (window->wl.libdecor.frame)
     {
         libdecor_frame_set_fullscreen(window->wl.libdecor.frame,
-                                      window->monitor->wl.output);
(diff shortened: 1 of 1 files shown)


ninja-:
isn't window->monitor->wl.output supposed to be the window's monitor, aka the monitor holding >50% of the app?

it's weird that with latest glfw and wayland, Minecraft would always maximize on the first display, ignoring display of the window; maximizing on the display where the game window is would be be closest to user expectations

PR #2590, opened 2024-07-17
EGL: GLFW_ANGLE_PLATFORM_TYPE_VULKAN on macOS

nee-wom:
ANGLE vulkan backend is available for macOS and supports OpenGL ES 3.1 (both metal and OpenGL backends are limited to ES 3.0)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -294,6 +294,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - nee-wom
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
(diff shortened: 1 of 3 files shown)


edwinRNDR:
I would like this to be merged in. Currently I have to set the `ANGLE_DEFAULT_PLATFORM` environment variable to `vulkan` to be able to use the ANGLE Vulkan backend.

planetmarshall:
Thanks for this - I had to do something similar in my local fork.

It's worth noting that future versions of ANGLE (after chromium/7632) will require [MoltenVk](https://github.com/KhronosGroup/MoltenVK) 1.4.1 to provide a GLESv3 context due to some stricter requirements on Vulkan.

See https://issues.angleproject.org/issues/482620896 for more details.

PR #2614, opened 2024-09-04
Clarify that only shareable objects are shared between contexts

TobiSchluter:
The docs say that glfw asks that "all objects" be shared. This is misleading. Qualify this by making explicit that there are shareable (and thus non-shareable) objects.


Henrryxdjr (approved):

TobiSchluter:
Any chance this will be merged?

Issue #2601, opened 2024-08-06
_glfwGrabErrorHandlerX11 assert with no window manager running


joserfg2:
If there is no window manager running, detectEWMH returns without releasing the X11 error handler. The assert can be visible with test tools such as triangle-vulkan, window...

I tested it with an Xfce4 desktop and killing xfwm4 (with no restart).

OS and version: Red Hat Enterprise Linux release 8.6
Release or commit: glfw-3.4

Error messages:

build/tests $ triangle-vulkan
triangle-vulkan: /home/test/Downloads/glfw-3.4/glfw-3.4/src/x11_init.c:1099: _glfwGrabErrorHandlerX11: Assertion `_glfw.x11.errorHandler == NULL' failed.
Aborted (core dumped)

Call stack:

#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff63c2db5 in __GI_abort () at abort.c:79
#2  0x00007ffff63c2c89 in __assert_fail_base (fmt=0x7ffff652b898 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x2023b50 "_glfw.x11.errorHandler == NULL", 
    file=0x2023b18 "/home/test/Downloads/glfw-3.4/glfw-3.4/src/x11_init.c", line=1099, function=<optimized out>)
    at assert.c:92
#3  0x00007ffff63e83a6 in __GI___assert_fail (assertion=0x2023b50 "_glfw.x11.errorHandler == NULL", 
    file=0x2023b18 "/home/test/Downloads/glfw-3.4/glfw-3.4/src/x11_init.c", line=1099, 
    function=0x2024240 <__PRETTY_FUNCTION__.17056> "_glfwGrabErrorHandlerX11") at assert.c:101
#4  0x00000000013fca06 in _glfwGrabErrorHandlerX11 ()
#5  0x00000000014013c9 in createNativeWindow ()
#6  0x0000000001404801 in _glfwCreateWindowX11 ()
#7  0x00000000013f47f7 in glfwCreateWindow ()

In x11_init.c around line 512:
```c
// Check whether the running window manager is EWMH-compliant
//
static void detectEWMH(void)
{
    // First we read the _NET_SUPPORTING_WM_CHECK property on the root window

    Window* windowFromRoot = NULL;
    if (!_glfwGetWindowPropertyX11(_glfw.x11.root,
[...]

PR #2631, opened 2024-09-29
Fix resource leak (error handler)

collinalexbell:
See #2601. The early return in detectEWMH is leaking the error handler resource. This is a problem for window manager developers.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -294,6 +294,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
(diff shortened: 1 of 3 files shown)


collinalexbell:
I've confirmed this fix works on my machine. I can reload my WM without restarting X server if I compile it with the patched version of GLFW

er-azh:
This is a duplicate of #2593, but that is also not yet merged.

Henrryxdjr (approved):

collinalexbell:
When will this get merged?

elmindreda:
This should be fixed now with #2593.

PR #2636, opened 2024-10-13
Pluggable clipboard functionality on X11

HazardousPeach:
GLFW itself only supports plain text (UTF-8) clipboard content. While this is enough for many applications, some applications need to allow copy and pasting other types of content between applications, like files and images. 

When applications need extra clipboard functionality and target a small number of platforms, they can use platform-specific code on many platforms to add the extra clipboard functionality they need. However, because of the way the X11 clipboard functionality interacts with the X11 event queue, it's not currently possible to fully implement this extra functionality while running on GLFW.

This pull request exposes a the X11 connection information that GLFW uses for clipboard functionality through the GLFW X11 native header/flag. It also adds a hook where clients can handle all X11 selection events that arrive to GLFW. This is about the minimum needed to allow clients to write their own clipboard handling code with extra functionality within a GLFW application. Since it is already possible to implement this functionality using the Windows native APIs, this allows an application which runs on Windows and Linux+X11 (and maybe others)  to add this functionality.

This functionality was developed as part of my work on JuniorsToolbox, so you can find commented client code that uses this new API [here](https://github.com/JoshuaMKW/JuniorsToolbox/blob/5c771af12e46a6574ed7892e29ff3cf67f1bbe3b/src/gui/clipboard.cpp#L385-L614); that should provide a decent example of how it can be used in other clients.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -294,6 +294,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Alex Sanchez-Stern
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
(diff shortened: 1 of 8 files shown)


Henrryxdjr (approved):

HazardousPeach:
Bumping this, looks like we just need one more review for merge?

PR #2661, opened 2024-12-18
Add gamepad level support for misc1 and touchpad buttons

TheSenPie:
According to the [input documentation on Gamepad mappings in GLFW](https://www.glfw.org/docs/latest/input_guide.html#gamepad_mapping) the misc1 and touchpad buttons are not currently supported, however they are part of the [SDL_GameControllerDB], which GLFW uses to handle logic for Xbox like controllers. 

This PR adds support for missing buttons misc1 and touchpad, which are present on DualShock and DualSense controllers.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -294,6 +294,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Smbat Senpie Voskanyan
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
--- a/README.md
+++ b/README.md
@@ -133,7 +133,8 @@ information on what to include when reporting a bug.
  - [Null] Added EGL context creation on Mesa via `EGL_MESA_platform_surfaceless`
  - [EGL] Allowed native access on Wayland with `GLFW_CONTEXT_CREATION_API` set to
    `GLFW_NATIVE_CONTEXT_API` (#2518)
-
+ - Added `GLFW_GAMEPAD_BUTTON_MISC1` and `GLFW_GAMEPAD_BUTTON_TOUCHPAD` to the
+   gamepad_buttons button group
 
 ## Contact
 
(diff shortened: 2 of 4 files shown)


dougbinks:
You've changed some of the pre-existing defines for the DPAD, which could cause issues (for example if someone serializes out their input assignment and updates GLFW). I'd recommend moving the new defines to after the end of the existing ones.

Extending the number of buttons effectively changes the API (see #2423 for a similar discussion) and so may need a new input mode with `glfwSetInputMode`, but I'll have to check with @elmindreda. This would mean that the defines would be something like:

```C
#define GLFW_GAMEPAD_BUTTON_DPAD_UP         11
#define GLFW_GAMEPAD_BUTTON_DPAD_RIGHT      12
#define GLFW_GAMEPAD_BUTTON_DPAD_DOWN       13
#define GLFW_GAMEPAD_BUTTON_DPAD_LEFT       14
#define GLFW_GAMEPAD_BUTTON_LAST            GLFW_GAMEPAD_BUTTON_DPAD_LEFT
#define GLFW_GAMEPAD_BUTTON_MISC1           15
#define GLFW_GAMEPAD_BUTTON_TOUCHPAD        16
#define GLFW_GAMEPAD_BUTTON_LAST_EXT1       GLFW_GAMEPAD_BUTTON_TOUCHPAD        
```

TheSenPie:
I get your point. The reason why I added the misc1 and touchpad before dpad is exactly because I saw that some of the code parts used the assumption `GLFW_GAMEPAD_BUTTON_LAST - 4` for dpad hatml impl and didn't want to mess with it. I'm not familiar with `glfwSetInputMode` I'll check it out.

TheSenPie:
@elmindreda any update on this?

hackbar:
any chance this can get merged?

TheSenPie:
@dougbinks what does it take to merge this PR? It's been a while, so I lost a bit of context, but if reordering the defines as you specified is enough I can do that.

TheSenPie:
> any chance this can get merged?

Hopefully before GTA VI

PR #2667, opened 2024-12-31
RFC: add Nintendo Wii&GameCube support

mardy:
This PR is a draft, because the code needs some serious clean-up, but I'm submitting it already just to collect some early feedback to know if this is something you'd accept in principle, or if you rather wouldn't like to support obsolete platforms (in that case, we can maintain this as a big patch in devkitPro).

If you accept this changes upstream, I do commit on maintaining the Ogc platform. I did a similar PR to FreeGlut some months ago, and I've been maintaining the Ogc backend since then: https://github.com/freeglut/freeglut/pulls?q=is%3Apr+author%3Amardy+is%3Aclosed

If you'd like to get this code in, but you see some major issues with the code itself, please don't hesitate to comment even if this is a draft.


Fancy2209:
Shouldn't _glfwInitOgc call L2Enhance on the wii ifdef?

mardy:
> Shouldn't _glfwInitOgc call L2Enhance on the wii ifdef?

I'm not familiar with this call, and any drawbacks it might have. I see that we are calling it in the SDL2 `main()` function, whereas it's not being called in the freeglut port. If it's completely safe, I could certainly call it from the glfw initialization function, but the documentation for it makes me wonder if there are reasons to avoid it. Maybe @DacoTaco knows more about it?

DacoTaco:
> > Shouldn't _glfwInitOgc call L2Enhance on the wii ifdef?
> 
> I'm not familiar with this call, and any drawbacks it might have. I see that we are calling it in the SDL2 `main()` function, whereas it's not being called in the freeglut port. If it's completely safe, I could certainly call it from the glfw initialization function, but the documentation for it makes me wonder if there are reasons to avoid it. Maybe @DacoTaco knows more about it?

Y0!

its the first time i hear of that function, and by the looks of it it reconfigures the L2 cache to have 64k cache lines instead of 32k.
However, the comment suggests this comes with some stability issues.
Ive never seen it used, nor used it myself so i have no idea if its true. 

if the performance is worth it, i'd consider calling it as soon as you can in the app so that as many things can be cached correctly instead of having to do it mid runtime :)

dougbinks:
Hi! Many thanks for submitting this PR. Adding Nintendo Wii & GameCube is impressive.

We don't have either the equipment or bandwidth to maintain support for these platforms, and we won't add platforms to core GLFW which we can't support. We are considering adding a Ports section to the readme to which we could add links to this and to the AmigaOS4 port from #2392 .

PR #2678, opened 2025-01-15
Fix uninitialized variable rgba

rex4539:
Fixes

```
examples/particles.c:504:26: error: Uninitialized variable: rgba [uninitvar]
            vptr->rgba = rgba;
                         ^
```


dougbinks:
This looks like a false warning. The 4 byte (GLuint) variable rgba is fully written to prior to being read, see:

https://github.com/glfw/glfw/blob/master/examples/particles.c#L490-L495

PR #2679, opened 2025-01-28
Wayland: Implement cursor-shape-v1 protocol

### LINKED ISSUE #2668: Support for the cursor-shape-v1 protocol
Currently GLFW gets the cursor theme and size from `XCURSOR_THEME/SIZE`, but KDE Plasma 6 (maybe other desktops/compositors too) no longer have these variables set by default so the cursor theme/size is wrong:
Screenshot of boing example from master
Instead the cursor should look like this:
image
(Note: ClassiCube is using SDL3 here, it does not support GLFW)

The [cursor-shape-v1 protocol](https://wayland.app/protocols/cursor-shape-v1) allows you to set the cursor shape without jumping through hoops like getting the cursor theme/size etc, though not all compositors support it yet, so you could keep the old "solution" around for now along with the new protocol.

JakobDev:
This PR adds implements the [Cursor shape protocol](https://wayland.app/protocols/cursor-shape-v1) on Wayland. This protocol let's the Compositor render the cursor, so the cursor looks the same on every program.

The protocol is used with the default cursor and cursors created by `glfwCreateStandardCursor`.

`tablet-unstable-v2` is needed to compile `cursor-shape-v1`.

Fixes #2668


Zamundaaa:
looks good to me!

ghost:
Tested it with https://github.com/WerWolv/ImHex, the cursor size is correct now. I would say lgtm

src/CMakeLists.txt
```diff
@@ -105,6 +105,8 @@ if (GLFW_BUILD_WAYLAND)
     generate_wayland_protocol("fractional-scale-v1.xml")
     generate_wayland_protocol("xdg-activation-v1.xml")
     generate_wayland_protocol("xdg-decoration-unstable-v1.xml")
+    generate_wayland_protocol("tablet-unstable-v2.xml")
```
llyyr, line 108: You could just declare a dummy `zwp_tablet_tool_v2_interface` object somewhere to avoid bringing in tablet-unstable-v2 for this, since we won't even be using it.

PrakarshPanwar:
Hi @JakobDev, @llyyr. I managed to make it work without `tablet-unstable-v2` by adding dummy `zwp_tablet_tool_v2_interface` struct in `wl_window.c`. How to add changes in this PR upstream?

cfnptr:
Hello! Any news on this? It would be awesome to have the same cursor inside the window on KDE Wayland :)

jdkeke142:
Tested this patch on Minecraft 1.21.11 + KDE Plasma 6.3.6 + NVIDIA. Cursor theme and size are now consistent between the game window and the rest of the desktop.

Applied as part of a patchset: https://github.com/jdkeke142/glfw-wayland-minecraft

JakobDev:
I included some changes from @PrakarshPanwar

PR #2681, opened 2025-02-03
fix: unable to make window transparent in later win10 versions

std-microblock:
related issues/prs:
https://github.com/rust-windowing/winit/pull/675
https://github.com/std-microblock/b-shell/issues/1

this only fixes it if GLFW_DECORATED is off, else it's still broken.

before fix:
image

after fix:
image
(still broken without GLFW_DECORATED off)

image

thanks @well2333 for providing his PC for me to fix this issue.

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -68,6 +68,9 @@ static DWORD getWindowExStyle(const _GLFWwindow* window)
 {
     DWORD style = WS_EX_APPWINDOW;
 
+    if (_glfw.hints.framebuffer.transparent)
+        style |= WS_EX_LAYERED;
+
     if (window->monitor || window->floating)
         style |= WS_EX_TOPMOST;
 
@@ -380,15 +383,17 @@ static void updateFramebufferTransparency(const _GLFWwindow* window)
     if (FAILED(DwmIsCompositionEnabled(&composition)) || !composition)
        return;
 
+
     if (IsWindows8OrGreater() ||
         (SUCCEEDED(DwmGetColorizationColor(&color, &opaque)) && !opaque))
     {
         HRGN region = CreateRectRgn(0, 0, -1, -1);
         DWM_BLURBEHIND bb = {0};
-        bb.dwFlags = DWM_BB_ENABLE | DWM_BB_BLURREGION;
+        bb.dwFlags = DWM_BB_ENABLE;
         bb.hRgnBlur = region;
         bb.fEnable = TRUE;
 
+        SetLayeredWindowAttributes(window->win32.handle, 0x0030c100, 254, LWA_ALPHA);
         DwmEnableBlurBehindWindow(window->win32.handle, &bb);
         DeleteObject(region);
     }
@@ -1294,6 +1299,7 @@ static int createNativeWindow(_GLFWwindow* window,
 #else
         wc.lpszClassName = L"GLFW30";
 #endif
+        wc.hbrBackground = (HBRUSH)CreateSolidBrush(0x00000000);
         // Load user-provided icon if available
         wc.hIcon = LoadImageW(GetModuleHandleW(NULL),
                               L"GLFW_ICON", IMAGE_ICON,


dougbinks:
I can't replicate this issue, so currently can't test it other than to confirm that the changes do not seem to break `GLFW_TRANSPARENT_FRAMEBUFFER` on my system.

From the screenshots I can't tell if this implements framebuffer transparency with alpha. Can you test this with the GLFW `gears.c` example by making it non decorated?

The `gears.c` example should look like this (with decoration here):
image

dougbinks:
Following information from https://github.com/glfw/glfw/issues/2731 I have been able to replicate this using an AMD GPU and setting AMD Adrenaline and go to `Gaming -> Graphics -> OpenGL Triple Buffering` to `On`.


Without this PR:
image

With this PR:
image

So it appears to work even with decorations.

I'll review this PR shortly.

dougbinks:
From the [`WS_EX_LAYERED` documentation](https://learn.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles): *This style cannot be used if the window has a [class style](https://learn.microsoft.com/en-us/windows/win32/winmsg/about-window-classes) of either CS_OWNDC or CS_CLASSDC.*

The GLFW class style is: `wc.style         = CS_HREDRAW | CS_VREDRAW | CS_OWNDC;`

So this shouldn't be able to be set.

This also cannot be set on child windows in Windows 7 which we still support. Although GLFW itself doesn't create child windows I know some people do alter it to do so.

I note that the following also works on my machine for `SetLayeredWindowAttributes`:
```C
        SetLayeredWindowAttributes(window->win32.handle, 0, 255, LWA_ALPHA);
```

I think the alpha value should be 255 since the doc says: *Alpha value used to describe the opacity of the layered window. Similar to the SourceConstantAlpha member of the [BLENDFUNCTION](https://learn.microsoft.com/en-us/windows/desktop/api/wingdi/ns-wingdi-blendfunction) structure*, and the BLENDFUNCTION documentation says: *Set the **SourceConstantAlpha** value to 255 (opaque) when you only want to use per-pixel alpha values.*

I did a quick check to see if this change affects performance, which it did not.

I need to investigate this a bit more.

Nifty255 (commented):
Confirmed, the additions in this PR resolve my transparency issue #2731, across various configs which suffered the same problem including AMD OpenGL Triple Buffering, AMD integrated GPUs, and others (All Win 10/11). After testing it myself, I've found that all 3 additions are required.

There's a random extra semicolon on line 392. Otherwise, good.

dougbinks:
Camilla and I have done some more testing of this, using the `SetLayeredWindowAttributes` attribute change I propose and there is a problem with one configuration we tested:

- Windows 7, dedicated GTX 660M, with both DWM enabled and disabled: window is fully invisible with only the taskbar button visible.

Otherwise the change works on a variety of systems in addition to those mentioned in earlier comments:

- Windows 11 system with AMD CPU and AMD discrete GPU () with triple buffering on or off (RX 7900 XTX)
- Windows 10 system with all AMD system, with and without triple-buffering ( RX 7600).
- Windows 10 system with Intel and Nvidia (Optimus)  (Intel HD 630 and GTX 1050M).


I would propose adding the `IsWindows8OrGreater()` checks to the changes, i.e.

```C
    if (_glfw.hints.framebuffer.transparent && IsWindows8OrGreater())
        style |= WS_EX_LAYERED;
```

and
```C
    if (IsWindows8OrGreater() ||
        (SUCCEEDED(DwmGetColorizationColor(&color, &opaque)) && !opaque))
    {
        HRGN region = CreateRectRgn(0, 0, -1, -1);
        DWM_BLURBEHIND bb = {0};
        bb.dwFlags = DWM_BB_ENABLE | DWM_BB_BLURREGION;
        bb.hRgnBlur = region;
        bb.fEnable = TRUE;

        if (IsWindows8OrGreater())
          SetLayeredWindowAttributes(window->win32.handle, 0, 255, LWA_ALPHA);
        DwmEnableBlurBehindWindow(window->win32.handle, &bb);
        DeleteObject(region);
    }
```

I would add the check to the brush creation and use a stock brush:

```C
    if (IsWindows8OrGreater()))
        wc.hbrBackground = (HBRUSH)GetStockObject(BLACK_BRUSH);
```

adambasis:
I don't use glfw but in my opengl code when I removed `DWM_BB_BLURREGION` like in the issue, transparency was slightly off. Fully transparent pixels were darkened. When I enabled `DWM_BB_BLURREGION` again, it worked. So just fyi, might want to test that out. Again I don't use glfw so no idea.

PR #2682, opened 2025-02-03
Add GLFW_FULLSCREEN to supported window attributes (for X11)

Minterl:
Only implements the functionality for X11.
Falls back to the null platform implementation (returns 0) for all other platforms.


dougbinks:
The GFLW API currently has [glfwGetWindowMonitor](https://www.glfw.org/docs/latest/group__window.html#ga4d766499ac02c60f02221a9dfab87299) which returns the handle of the monitor that the specified window is in full screen on. So if this returns `NULL` the window is not fullscreen, and if it is a pointer to a `GLFWmonitor` the window is fullscreen.

Does this PR provide something extra which the above function doesn't?

PR #2686, opened 2025-02-12
Wayland: Implement `glfwSetWindowIcon`

### LINKED ISSUE #2597: Wayland window icon support
This protocol is quite recent: https://wayland.app/protocols/xdg-toplevel-icon-v1

It would allow to set a window icon and could enable to implement `_glfwSetWindowIconWayland`
If the Wayland compositor doesn't implement this protocol, just printing a warning like before is good enough.

Note: I got surprised when I saw that there was no issues opened for this, this is why I opened this one.

JakobDev:
This implements `glfwSetWindowIcon` on Wayland. This requires the Compositor to support the [XDG toplevel icon protocol](https://wayland.app/protocols/xdg-toplevel-icon-v1). At the moment this is only supported on the latest KDE version 6.3.

<details>
<summary>Screenshot</summary>

The following Screenshot shows `tests/icon` running under Wayland.

grafik

</details>

Note for those who want to test this with Minecraft:
`glfwSetWindowIcon` is for some reason only called under X11 but not under Wayland. You can test this by adding `printf` to the top of the function. So using this patch alone won't give you a Icon under Wayland.

Fixes #2597

### DIFF
--- a/deps/wayland/xdg-toplevel-icon-v1.xml
+++ b/deps/wayland/xdg-toplevel-icon-v1.xml
@@ -1,0 +1,205 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="xdg_toplevel_icon_v1">
+
+  <copyright>
+    Copyright © 2023-2024 Matthias Klumpp
+    Copyright ©      2024 David Edmundson
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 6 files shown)


TuxTheAstronaut:
I tested this with Minecraft and it does work Although in 1.21 versions of MC you have to use a mod that sets a window icon for it to work properly.

JakobDev:
As I had feared it looks like there is a condition inside Minecraft that checks if it's not running under Wayland before setting the Window icon. So when this is merged (what I don't expect in the next time given that many PRs here are just ignored by the Maintainers) I need to open a Bug in the Minecraft Bugtracker and wait to Mojang to resolve this Issue. Thanks for confirming this.

Fox2Code:
@JakobDev I opened an issue about the state of the project https://github.com/glfw/glfw/issues/2692

I think if no maintainer is left, getting the LWJGL folks to try to maintain a glfw fork might not be a bad idea.

orowith2os:
Would it be helpful to have an additional event on toplevel-icon-v1 to communicate to the compositor that the set icon is a child of another app, and that it would be nice if it linked the two in the compositor's UI somehow (if possible)? 

I can draft a protocol addition to w-p if so. I'm thinking GLFW could have an environment variable to configure whether this event is set or not, and game launchers could set it if they want their games to have this behavior (the main use case is Minecraft launchers).

DanielGibson:
Thank you very much, this works well on my machine (with latest KDE)!

Should be merged, IMHO

dougbinks:
> what I don't expect in the next time given that many PRs here are just ignored by the Maintainers

We're not ignoring PRs, we just are not able to merge them as quickly as some people would like.

ws909:
Same enhancement but for MacOS: 

- https://github.com/glfw/glfw/pull/2293

edvn0:
Do you need any assistance on implementation here? Is this mergable?

JakobDev:
> Is this mergable?

Should be mergable. There are currently no concerns.

PR #2688, opened 2025-02-16
resolve build problem in cygwin toolchain

padmadevd:
when building using cmake. it defines the macro _GLFW_WIN32 for win32. but the glfw don't recognise the macro. so that the definitions in the win_thread.c .... etc are not defined. so undefined error arise when building a application. This error arises only when using cygwin toolchain. when using mingw, it defines __MINGW32__ so the glfw recognise it.


padmadevd:
is there any problem in the change? or is this fix not important?

Issue #2687, opened 2025-02-12
Handling GLFW termination on library side.


abhi-k9:
Hi,

I have a library that uses GLFW to setup the context for OpenGL offscreen rendering. At the moment, I initialize and terminate the framework around my API calls.

However, the client code could also be depending on GLFW and need it in a initialized state after they use my API, i.e. my library should not terminate GLFW.

I was thinking of checking if it was already initialized upon entry to my API and if so do not terminate once the API is finished, but #2036 mentions this is not possible and attempts to add reference counting have been hanging for a long time #2070.

Are there any workarounds for this use case?

israfiel-a:
Have you tried using flags? Perhaps a function parameter to one of your library methods? This seems like more of a client-issue rather than a GLFW one.

abhi-k9:
I think that would pass the torch up the client stack while the core issue remains. For example, client C depends on my library L1 and also another library L2.

If L2 internally initializes GLFW without notifying C, there is no way C can pass on that information to me, i.e. L1. If I then terminate GLFW thinking no one else is using it that would break L2.

This information, whether GLFW is initialized has to be centralized and the best place is GLFW itself following KISS principle. I don't see any downsides to having a `glfwIsInitialized()` function.

Thanks.

israfiel-a:
That's a valid point. It would also be relatively simple to implement. I'd suggest patching the library yourself until this issue can be properly reviewed by a maintainer. For example, you could add a function signature (i.e. `int glfwIsinitialized(void)`) to the GLFW header (`GLFW.h`), and then in `src/init.c`, you could add the following code:

```c
int glfwIsInitialized(void) {
    return _glfw.initialized;
}
```

I've not compiled, but I am 99% sure the following code would work perfectly for your purposes.

abhi-k9:
I have created a pull request for the patch if it helps.
Thanks.

PR #2697, opened 2025-03-09
Add external API to query initialization status of GLFW.

abhi-k9:
Pull request in reference to #2687.

Adds the ability to check if GLFW has been already initialized.


israfiel-a:
I would suggest changing the body of `glfwIsInitialized` to the following;

```c
GLFWAPI int glfwIsInitialized(void)
{
    return _glfw.initialized;
}
```

This both better fits code styling and removes some confusing extra logic.

abhi-k9:
Agreed. Thanks for suggestions.

PR #2714, opened 2025-04-25
Update CMakeLists.txt

shiyou-t1c:
According to the CMake official documentation (as of version 3.28), the valid options for the configure_file command are limited to: @ONLY, ESCAPE_QUOTES, COPYONLY, and NEWLINE_STYLE, while IMMEDIATE is not recognized as a standard option. https://cmake.org/cmake/help/v3.28/command/configure_file.html#configure-file


dougbinks:
There are more changes here than you describe. Could you either break these out into separate PRs or describe what these are?

shiyou-t1c:
https://cmake.org/cmake/help/v3.28/command/configure_file.html#configure-file

-Wl,--nxcompat is a parameter used during linking. CMake's check_c_compiler_flag function is only used during the compilation stage to check whether a specific compiler flag is supported by the current C compiler.

check_c_source_compiles compatible with older CMake versions.
Pre-3.1: The command was part of the deprecated CheckCCompilerFlag module.
Post-3.1: Officially moved to the CheckCSourceCompiles module.
# Pre-3.1 workaround 
include(CheckCCompilerFlag)
# Post-3.1 
include(CheckCSourceCompiles)
```cmake
if(CMAKE_VERSION VERSION_LESS 3.1)
    include(CheckCCompilerFlag)
    # Legacy syntax here 
else()
    include(CheckCSourceCompiles)
    # Modern syntax here 
endif()
```
find_library:
This command is used to find a library. A cache entry, or a normal variable if NO_CACHE is specified, named by <VAR> is created to store the result of this command. If the library is found the result is stored in the variable and the search will not be repeated unless the variable is cleared. If nothing is found, the result will be <VAR>-NOTFOUND.

shiyou-t1c:
I saw the online build passed, so I didn't add include(CheckCSourceCompiles). It was a bit of a gamble.

dougbinks:
`find_library`

The CMake projects own test code uses the same syntax as the current GLFW CMake code:
https://github.com/Kitware/CMake/blob/master/Tests/CMakeOnly/find_library/CMakeLists.txt#L8-L21

Do you have specific examples where this currently fails to work as intended?

shiyou-t1c:
I just learned about the implicit logic of `if()` in CMake: if a variable is `0`, `OFF`, `NO`, `FALSE`, `N`, `IGNORE`, `NOTFOUND`, an empty string, or ends with `-NOTFOUND`, it evaluates to `False`. All other cases are treated as `True`. I believe my version offers better readability and would appreciate your consideration.

dougbinks:
I think we should stick to the CMake standard approach of `if (LIB_VAR_NAME)`, though I agree that CMake syntax and logic is not as understandable as it could be.

shiyou-t1c:
If you have access to the repository, feel free to make the changes as you see fit. I was just expressing my preference—there may be deeper reasons behind your approach that I haven’t fully grasped yet.

PR #2724, opened 2025-06-10
include xkb_common.h in wl_platform.h as well

ashpil:
Fixes this on my machine:

```
/home/andrei/Projects/glfw/src/xkb_unicode.c:939:12: error: use of undeclared identifier 'GLFW_INVALID_CODEPOINT'
    return GLFW_INVALID_CODEPOINT;
````

I imagine I'm running into this (but others/CI is not) because I'm hand-rolling my own build system for this rather than using the provided CMake one. Nonetheless, this seems like a valid change, as Wayland also requires `xkb_common.h`.


dougbinks:
The Wayland platform no longer uses xkb_unicode as of https://github.com/glfw/glfw/commit/21fea01161e0d6b70c0c5c1f52dc8e7a7df14a50

You can see that the include was removed from `wl_platform.h` and the source removed from the build then.

So I would undo this change and remove `xkb_unicode.c` from your hand-rolled build.

ashpil:
Oops, my bad. Should've noticed this.

Thanks for the quick response!

Issue #2727, opened 2025-06-19
Failed assertion when receiving key repeat event with no keyboard focus


jamii:
(This is against https://github.com/tiawl/glfw.zig/commit/a611b31721c05010e7cbfcaea28e8075ff9ed40c, which has a different build setup but should otherwise be vanilla glfw 3.4. That might make repro hard, but I thought it was worth recording this somewhere anyway.)

I'm seeing a few crashes per day in:

```
focus: /home/jamie/focus/deps/glfw/glfw/src/input.c:274: void _glfwInputKey(_GLFWwindow *, int, int, int, int): Assertion `window != NULL' failed.
Set Windows Application User Model ID (AUMID) { AUMID: 'org.whispersystems.signal-desktop' }
NODE_ENV production
NODE_CONFIG_DIR /nix/store/s81a57m4lnr0l32n29v2ax22vqypw2vh-signal-desktop-7.56.1/share/signal-desktop/app.asar/config
NODE_CONFIG {}
ALLOW_CONFIG_MUTATIONS undefined
HOSTNAME vessel
NODE_APP_INSTANCE undefined
SUPPRESS_NO_CONFIG_WARNING undefined
SIGNAL_ENABLE_HTTP undefined
userData: /home/jamie/.config/Signal
config/get: Successfully read user config file
config/get: Successfully read ephemeral config fil
```

From the coredump, it looks like the problem is that glfw sometimes receives a key repeat event while there is no keyboard focus.

```
>>> bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff7c99d43 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:78
#2  0x00007ffff7c413c6 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff7c2893a in __GI_abort () at abort.c:79
#4  0x00007ffff7c2885e in __assert_fail_base (fmt=0x7ffff7dbdf20 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x103c61b "window != NULL",
[...]

PR #2732, opened 2025-06-26
Wayland: Ignore key repeat events when no window has keyboard focus

GamesTrap:
Fix for:

- #2727

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -229,7 +229,7 @@ video tutorials.
  - Brandon Schaefer
  - Sebastian Schuberth
  - Scr3amer
- - Jan Schuerkamp
+ - Jan Schürkamp
  - Christian Sdunek
  - Matt Sealey
  - Steve Sexton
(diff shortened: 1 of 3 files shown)


GamesTrap:
Seems like MSVC is unhappy about deps/nuklear.h :(

>  D:\a\glfw\glfw\deps\nuklear.h(3433,37): error C2220: the following warning is treated as an error [D:\a\glfw\glfw\build-win32-shared-x86\tests\inputlag.vcxproj]
>
> (compiling source file '../../tests/inputlag.c')

Upstream issue: https://github.com/Immediate-Mode-UI/Nuklear/issues/751

dougbinks:
Thanks for this.

You asked about whether `event = GLFW_TRUE;` should be set outside the test for `_glfw.wl.keyboardFocus`, I believe not though the behaviour of event waiting is not currently consistent across platforms.

Checklist for bug fixes:

  - *Change log entry in README.md, describing the incorrect behavior* DONE
  - *Credits entries in CONTRIBUTORS.md for all authors of the bug fix* DONE, you are in the credits already as *Jan Schuerkamp* but we can change it to Jan Schürkamp if you like since it's UTF-8 encoded.

I'll make a local fix for `nuklear.h` and then merge this PR. If you want your name changed to include the accent let me know (if it's after the merge I can make the change in a separate commit).

GamesTrap:
> If you want your name changed to include the accent let me know (if it's after the merge I can make the change in a separate commit).

I've changed it now :)

dougbinks:
Merged, many thanks for the PR!

Issue #2145, opened 2022-07-02
GUID of XInput controllers on Windows


Sentmoraap:
Hello,

when I plug a XInput controller on Windows, it's GUID is 78696e70757401000000000000000000 which is not the format used by SDL2.

I tried with two different controllers : an 8bitdo Lite that exposes itself as 0452 02e0 when in XInput mode, and a Xbox Series X|S controller which is 0452 02ff on Windows and 0452 0b12 on Linux.

On Windows both have the 7869… GUID. On Linux they are respectively 030000005e0400008e02000014010000 and 030000005e040000120b000005050000.

Because of this I can assume that it's a XInput controller but not which one.

I use glfw 3.3.7 (64 bits) on Windows 10 Entreprise N LTSC 21H2 19044.1766 and I tried with 3.3.2-2 and 3.3.7-1 on Debian Bookworm/Linux 5.15.0-2-amd64.

HID controllers have GUID in the correct format, with bcdDevice (bytes 24 to 27) always equals to zero on Windows.

CuriouserThing:
Out of curiosity, are you able to test on SDL2? I know XInput/Win32 is partly to blame here, but I also know that SDL2 frankensteins together a joystick representation using a mix of XInput, DirectInput, and RawInput. I believe SDL2 also has a generic xinput GUID, but I'm unsure the context in which it's used.

Sentmoraap:
I tested it. For some reason they are different on windows, but they use the expected format. The Xbox controller is 0300938d5e040000ff02000000007200. The 8bitdo is 030003f05e0400008e02000000007200.

ixsetf:
I also have run into this GUID issue, and I took a look through the source code to see if I could identify the root cause. 

It looks like the the source of the incorrect GUID is [this line in win32_joystick.c](https://github.com/glfw/glfw/blob/d24ee9953f49c0554ebff09c063e0212a2bfc5bf/src/win32_joystick.c#L526-L527), and to correct it we would need to replicate the output from [this function in SDL_joystick.c](https://github.com/libsdl-org/SDL/blob/0a924b185d2f459a2efbe54484394a17b97f1632/src/joystick/SDL_joystick.c#L2723). If I can find enough time, I may attempt a fix, but if not, I hope this will be helpful to whoever decides to tackle this issue.

PR #2740, opened 2025-07-15
Win32: Update GUID for SDL newer format

lassade:
It should match with `SDL_GameControllerDB` format no longer needed of the GLFW variations ...

By the way whats the purpose of  `_glfwUpdateGamepadGUIDWin32` ? I coundn't find a guid that matches the criteria to be modified


ixsetf:
This seems to incidentally fix [Issue #2145](https://github.com/glfw/glfw/issues/2145), just as an FYI. This should be verified by someone else as I am not very familiar with the library internals and may have missed something.

PR #2745, opened 2025-07-21
update mappings.h

me-it-is:
according to this [documentation](https://github.com/glfw/glfw/blob/master/docs/input.md#gamepad-mappings-gamepad_mapping) the mapping in mappings.h should be updated every release however it seems that the mappings were last updated in 2021.


DaimosMark2:
I have XBox Series pad. I enabled support under Linux by adding:

"050000005e040000130b000021050000,Xbox Wireless Controller,a:b0,b:b1,back:b10,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b12,leftshoulder:b6,leftstick:b13,lefttrigger:a5,leftx:a0,lefty:a1,rightshoulder:b7,rightstick:b14,righttrigger:a4,rightx:a2,righty:a3,start:b11,x:b3,y:b4,platform:Linux,\n"

It is missing in proposed change.

me-it-is:
Thats likely just because i haven't reran the generation in half a year

me-it-is:
It seems its not in the [mappings data base](https://github.com/mdqinc/SDL_GameControllerDB) at all you should open a pr to add it

me-it-is:
This pr should be up to date now

sciencewhiz:
It looks like there's a lot of unrelated .plist files in the PR now.

me-it-is:
Sorry about that i accidentally included the build folder, it should be fixed now.

crocdialer:
I agree with the PR-description but 
while this PR is fixing the outdated mappings, 
the root-cause stays: a manual update + PR are required each time.

I've opened #2887, which automates this instead: a scheduled workflow running the existing
`update_mappings` command and opening a PR whenever the upstream database changes. It ran
end-to-end on my fork — crocdialer/glfw#2 is the PR it opened and merged on its own.

- reason I went that way rather than adding to this PR: this isn't a one-time  problem.
- this PR has been open since July 2025 and already needed re-running once while waiting.
- automating the trigger removes the need for anyone to notice and manually file PR in the future.

PR #2749, opened 2025-07-29
Fix windowProc not working when SetPropW() fails silently

Peru-S:
**Issue**

In some cases, on Windows, `SetPropW` returns `FALSE`.
glfw currently silently ignores the result.
This means that the message pump `windowProc` fails to work (does not receive keyboard events for example). Drawing is fine though as it's a different GLFW mechanism.

I had this failure frequently infrequently and it was frustrating to deal with (just have to rerun the program a few times and hope `SetPropW` doesn't return `FALSE`). So sending this fix in case others are facing the issue on Windows.

**Fix**

On Windows+GLFW, `windowProc` will fallback to looking up the GLFW window list (just like `PollEvents` does).
Also added the error code to `_glfwInputError` message to aid debugging in the future.

My suggestion is to get rid of `SetPropW`/`GetPropW` as it is a string-based Windows API (slow/fragile) - we already have the lightweight linked list, might as well just use it always - also it's usually just the one or two windows we deal with.

- [x] Verified on Windows - message pump is stable with this fix (once I was able to repro this issue).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -232,6 +232,7 @@ video tutorials.
  - Jan Schürkamp
  - Christian Sdunek
  - Matt Sealey
+ - Perumaal Shanmugam
  - Steve Sexton
  - Arkady Shapkin
  - Mingjie Shen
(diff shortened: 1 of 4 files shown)


dougbinks:
This PR has Copilot added as a reviewer. I can't seem to remove it, so I'm guessing you've added it as a reviewer. Could you remove it as it is confusing the PR with AI slop.

Peru-S:
Apparently _I_ have this setting enabled for myself automatically? Sorry about that! 

I have disabled it, but I couldn't actually _remove_ the robot from this review (isn't that like some Asimov rule)?

Hopefully it won't add more comments...

PR #2750, opened 2025-07-30
Win32:  Buffered approach to raw mouse motion, PR fix for #2684

### LINKED ISSUE #2684: `glfwPollEvents` very low performance when using high polling rate mouse
I recently switched to a mouse which has an 8000hz polling rate, and this causes my game to drop to single digits FPS.

While one solution could be to get everything out of the main thread (which I will probably do in the end), this still seems unreasonable.

As a matter of fact it seems that GLFW uses `GetRawInputData` which is the cause of the issue, as it was previously an issue in SDL, [SDL #8756](see https://github.com/libsdl-org/SDL/issues/8756).
The fix would be to use `GetRawInputBuffer` as per the previously mention issue in SDL.

Pannoniae:
Fixes #2684, now without all the commented out debug code. Hope it doesn't break anything.

### DIFF
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -43,6 +43,11 @@
 typedef UINT64 QWORD; // Needed for NEXTRAWINPUTBLOCK()
 #endif
 
+#ifndef RI_MOUSE_HWHEEL
+// MinGW may not have the define for RI_MOUSE_HWHEEL
+#define RI_MOUSE_HWHEEL 0x0800
+#endif
+
 
 // Returns the window style for the specified window
 //
--- a/src/win32_window.c
+++ b/src/win32_window.c
@@ -35,6 +35,14 @@
 #include <assert.h>
 #include <windowsx.h>
 #include <shellapi.h>
+#include <BaseTsd.h>
+#include <stdio.h>
+
+
+#ifdef _WIN64
+typedef UINT64 QWORD; // Needed for NEXTRAWINPUTBLOCK()
+#endif
+
 
 // Returns the window style for the specified window
 //
@@ -265,7 +273,7 @@ static void releaseCursor(void)
 //
 static void enableRawMouseMotion(_GLFWwindow* window)
 {
-    const RAWINPUTDEVICE rid = { 0x01, 0x02, 0, window->win32.handle };
+    const RAWINPUTDEVICE rid = { 0x01, 0x02, RIDEV_NOLEGACY, window->win32.handle };
 
     if (!RegisterRawInputDevices(&rid, 1, sizeof(rid)))
     {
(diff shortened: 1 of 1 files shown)


dougbinks:
It looks like the MinGW builds fail because `RI_MOUSE_HWHEEL` is not defined, so we may need:

```C
#ifndef RI_MOUSE_HWHEEL
// MinGW may not have the define for RI_MOUSE_HWHEEL
#define RI_MOUSE_HWHEEL 0x0800
#endif
```

Pannoniae:
> It looks like the MinGW builds fail because `RI_MOUSE_HWHEEL` is not defined, so we may need:
> 
> ```c
> #ifndef RI_MOUSE_HWHEEL
> // MinGW may not have the define for RI_MOUSE_HWHEEL
> #define RI_MOUSE_HWHEEL 0x0800
> #endif
> ```

ask and ye shall find!

dougbinks:
I've done an initial testing of the code, along with a brief review (one note added above). Given this affects all raw mouse input I'd like a bit more time testing and checking the code. Meanwhile do take a look at the guide on contributing a bug fix and feature. This is somewhat more like a bug fix as the optimization does not add new functionality, simply improves an existing one.

https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-bug-fix

So in addition to the code we need:

  - Change log entry in README.md, describing the incorrect behavior
  - Credits entries in CONTRIBUTORS.md for all authors of the bug fix

I can do these if you don't have the time. You can search previous commits for examples.

dougbinks:
@Pannoniae Could you remove your tag [GLFW "3.5.0"](https://github.com/Pannoniae/glfw/releases/tag/3.5.0)?

Sadly due to an error I made (see https://github.com/glfw/glfw/issues/2758) the tag got added to GLFW until I found out and deleted it. In order to reduce any confusion could you delete the tag (but keep the source code branch)?

Pannoniae:
yeah, I will do it in an hour or two (sorry about that, I've been busy the past week!)

dougbinks:
Do you have time to update the PR with changes related to the comments I've made, in particular the bug I mention in the first review comment?

If not I can look into making the required changes myself when merging.

Pannoniae:
sadly no, some things have come up...

dougbinks:
OK - then if I get time I'll look into making the required changes.

Pannoniae:
much appreciated!

IzumiiKonata:
Thanks for working on this! 

I tested this PR and found that when pressing three mouse buttons simultaneously (e.g., left, right, and a side button) and releasing them at the same time, there is a high probability that some buttons will remain stuck in the pressed state. (This issue only occurs when `GLFW_RAW_MOUSE_MOTION` is set to `GLFW_TRUE`.) 

Let me know if you need any more details on the reproduction steps.

dougbinks:
Thanks @IzumiiKonata that's a great bug report - I can indeed now see in the code that though a set of flags for each button pressed/released is processed, it's done assuming only one key event happens at a time.
https://github.com/glfw/glfw/pull/2750/files#diff-475ae98b1c5241adc1365b9672ee3c565946a3618c8d6f2e409500eb9dba0395R2139-R2192

I can see how to alter this to fix the issue but won't have time for a bit.

dougbinks:
I have pushed several fixes and updates to a branch on my fork:

https://github.com/dougbinks/glfw/tree/pr-2750-win32-buf-raw-m

This should fix the issue @IzumiiKonata raised along with the multi-window bug I mentioned and most of the review comments I made. The one remaining issue is bounding the raw buffer memory allocation to something sensible, and add a loop to cope with larger input buffers.

dougbinks:
I've now added a bounded raw input buffer with looping read, plus fixed a potential crash after an error.

I think the PR is now functionally complete (if testing shows no further issues), but needs some other changes as per the [contributing guide](https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-feature) except documentation since this feature does not change the API:

1. Change log entry in README.md, listing all new symbols
2. News page entry in docs/news.md, briefly describing the feature
3. Credits entries in CONTRIBUTORS.md for all authors of the feature

dougbinks:
I have updated both the `README.md` and `docs/news.md`, so all that remains to be complete is for @Pannoniae to let me know how they want to be credited in the [CONTRIBUTORS.md](https://github.com/glfw/glfw/blob/master/CONTRIBUTORS.md) file.

I've done quite a bit of testing, but if @Pannoniae and @IzumiiKonata could test my branch below that would be very helpful:
https://github.com/dougbinks/glfw/tree/pr-2750-win32-buf-raw-m

Pannoniae:
Just my name as it is on GitHub is fine, thank you! I'll try to see whether any more bugs come up.

dougbinks:
@Pannoniae I've updated the contributors. If you want to merge my branch into yours then I can merge the PR from the Github UI but that's not required.

Pannoniae:
I don't know how to merge it in :\

dougbinks:
No problem, I'll merge the PR from my branch keeping your authorship!

IzumiiKonata:
> I have updated both the `README.md` and `docs/news.md`, so all that remains to be complete is for @Pannoniae to let me know how they want to be credited in the [CONTRIBUTORS.md](https://github.com/glfw/glfw/blob/master/CONTRIBUTORS.md) file.
> 
> I've done quite a bit of testing, but if @Pannoniae and @IzumiiKonata could test my branch below that would be very helpful: https://github.com/dougbinks/glfw/tree/pr-2750-win32-buf-raw-m

I have tested the latest code on your branch and confirmed that the issue has been fully resolved. No abnormalities were observed in multi-window environments either.

Thanks for the quick fix! Great work, and looking forward to the final merge.

dougbinks:
Many thanks @IzumiiKonata.

dougbinks:
We've found a few more issues with the code, so a merge may be a bit of time away. One problem is that an GLFW input state change to the disabled cursor or raw input can occur during looping over the buffer, handling this requires care. My current solution is to continue processing the buffered input but to ignore cursor movement and only process mouse buttons/wheel inputs to ensure the state is consistent.

Leclowndu93150:
by the way this affects minecraft aswell

dougbinks:
I have updated my branch https://github.com/dougbinks/glfw/tree/pr-2750-win32-buf-raw-m by merging the main branch code (Start 3.6), and done some very quick testing.

decce6:
I've identified multiple issues in the PR and those who are interested can check the proposed fixes in https://github.com/dougbinks/glfw/pull/1.

PR #2759, opened 2025-08-15
Add support for creating WebGPU surfaces

Sebastian-Dawid:
This PR adds the option to create a WebGPU surface using GLFW by calling `glfwCreateWindowWGPUSurface`.

The address of `wgpuInstanceCreateSurface` must be provided to the implementation via `glfwSetWGPUInstanceCreateSurfaceAddr`.

Including `<GLFW/glfw3.h>` will also include `<webgpu/webgpu.h>` if `GLFW_INCLUDE_WEBGPU` is defined before the include.

---

Example:
```c
#define GLFW_INCLUDE_WEBGPU
#include <GLFW/glfw3.h>

int main(void)
{
    glfwInit();
    glfwSetWGPUInstanceCreateSurfaceAddr(wgpuInstanceCreateSurface);

    glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
    GLFWwindow* window = glfwCreateWindow(800, 600, "WebGPU", NULL, NULL);
    WGPUInstance instance = wgpuCreateInstance(NULL);

    WGPUSurface surface = glfwCreateWindowWGPUSurface(instance, window);

    wgpuSurfaceRelease(surface);
    wgpuInstanceRelease(instance);
    glfwTerminate();
    return 0;
}
```

### DIFF
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -6531,19 +6531,6 @@ GLFWAPI VkResult glfwCreateWindowSurface(VkInstance instance, GLFWwindow* window
 
 #if defined(WEBGPU_H_)
 
-/*! @brief Provide the address of the `wgpuInstanceCreateSurface` function to GLFW.
- *
- * This function passes the address provided for the `wgpuInstanceCreateSurface` function
- * to GLFW.
- *
- * @param[in] addr The address of the `wgpuInstanceCreateSurface` function.
- *
- * @since Added in version 3.5
- *
- * @ingroup webgpu
- */
-GLFWAPI void glfwSetWGPUInstanceCreateSurfaceAddr(WGPUSurface (*addr)(WGPUInstance, const WGPUSurfaceDescriptor*));
-
 /*! @brief Creates a WebGPU surface for the specified window.
  * 
  * This function creates a WebGPU surface for the specified window.
--- a/include/GLFW/glfw3.h
+++ b/include/GLFW/glfw3.h
@@ -6531,6 +6531,19 @@ GLFWAPI VkResult glfwCreateWindowSurface(VkInstance instance, GLFWwindow* window
 
 #if defined(WEBGPU_H_)
 
+/*! @brief Provide the address of the `wgpuInstanceCreateSurface` function to GLFW.
+ *
+ * This function passes the address provided for the `wgpuInstanceCreateSurface` function
+ * to GLFW.
+ *
+ * @param[in] addr The address of the `wgpuInstanceCreateSurface` function.
+ *
+ * @since Added in version 3.5
+ *
+ * @ingroup webgpu
+ */
+GLFWAPI void glfwSetWGPUInstanceCreateSurfaceAddr(WGPUSurface (*addr)(WGPUInstance, const WGPUSurfaceDescriptor*));
+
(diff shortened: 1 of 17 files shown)


dougbinks:
This seems similar to [#2333](https://github.com/glfw/glfw/pull/2333)

Sebastian-Dawid:
Note that the code to fetch the pre-compiled `wgpu_native` library is heavily based on the way this is implemented in @eliemichel's [WebGPU-distribution](https://github.com/eliemichel/WebGPU-distribution), which I used previously.

The CI pipeline on macOS needs access to both the x64 and arm64 implementation of WebGPU at the same time, which WebGPU-distribution can not provide at the moment, which is why I decided to inline fetching the library.

dougbinks:
Please note that I do not think we would accept a modification to the CMakelists which pulls code from the internet.

Sebastian-Dawid:
> Please note that I do not think we would accept a modification to the CMakelists which pulls code from the internet.

I see. Given that the example needs WebGPU to compile (and I can't include either its source code or binaries in this repo), how should I structure the example?

dougbinks:
The usual way to do this is to use CMake to check for the existence of the module/header dependencies  and then only set the define which builds WebGPU to `ON` if the dependencies exist.

Sebastian-Dawid:
Unfortunately using `find_package` to check if WebGPU is available on the system the example is compiled on is not possible at the moment, since WebGPU is (as far as I know) not distributed on any platform yet (and even if it was it would be unclear if the distributed implementation is `dawn` or `wgpu_native`, which still have differences in the parts of the API that are not standardized yet).

I could still use `find_package` with a locally defined module `FindWebGPU.cmake` that fetches the dependency. This would make it easier to swap this out in the future when and if WebGPU is distributed. This obviously would not solve the issue of fetching the pre-compiled binaries at configure-time of the CMake project though. I could also disable building the WebGPU example by default and add an option to enable building it.

dougbinks:
In addition to `find_package` [CMake can also check for the existence of header files and there are also other compile checks possible](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Write-Platform-Checks).

However disabling WebGPU is also a reasonable start.

dougbinks:
Note that I would favour merging the simpler PR https://github.com/glfw/glfw/pull/2333 over this. Whilst the OpenGL context creation code is somewhat mixed with the platform window code I do not think this should be the case for WebGPU, though I haven't had the time to speak with @elmindreda about their preference.

EDIT: overall the PR looks like good programming work, but the modification of core GLFW files gives me pause.

Sebastian-Dawid:
My rationale for the way I structured the code is that surface creation with WebGPU is very similar to surface creation with Vulkan, so I modeled my implementation for WebGPU on the way this was implemented for Vulkan. While it would have been possible to keep both the platform specific and platform agnostic code in `webgpu.c`, I thought all platform specific code should probably be in the source files relevant only to that platform instead of having `#if defined(...)` blocks for each platform in `webgpu.c`.

For now I will disable building the WebGPU example by default and add an option to enable building it. I will also move the CMake code that fetches the pre-compiled `wgpu_native` library to `CMake/modules/FindWebGPU.cmake` and use `find_package(WebGPU)` in case the option to build the example is set to `ON`.

PR #2763, opened 2025-08-31
Wayland: Add support for dropping files when sandboxed

tautropfli:
Adds support for receiving drops that make use of the [File Transfer][1] portal. This is the case in sandboxed environments such as Flatpak or Snap.

Note that the portal is of course also available outside a sandbox.

[1]: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileTransfer.html

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
             - name: Install dependencies
               run: |
                   sudo apt update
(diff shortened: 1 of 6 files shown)


dougbinks:
I note that you've added DBUS as a required library in the CMake file. The usual approach in GLFW is to load the library and functions needed dynamically. See `_glfwInitWayland` in `wl_init.c` for examples.

tautropfli:
Oh, that's a good point! I will update the PR accordingly :)

PR #2766, opened 2025-09-06
Prevent occasional app crash when waking from an idle display on X11

dweymouth:
This fixes an issue where GLFW apps will occasionally crash on X11 when waking up from system display idle.

The issue was discussed in the Kitty app [here](https://github.com/kovidgoyal/kitty/issues/1006) and this [same fix](https://github.com/kovidgoyal/kitty/commit/ab7f3310c4f7982026a897155eabb097683658b2) was applied to their vendored GLFW code. This is currently affecting [Fyne apps](https://github.com/fyne-io/fyne/issues/5899) and this same fix was proposed to [go-gl/glfw](https://github.com/go-gl/glfw/pull/411) but it was requested that this be fixed upstream.

### DIFF
--- a/src/x11_monitor.c
+++ b/src/x11_monitor.c
@@ -151,6 +151,11 @@ void _glfwPollMonitorsX11(void)
             }
 
             XRRCrtcInfo* ci = XRRGetCrtcInfo(_glfw.x11.display, sr, oi->crtc);
+            if (!ci) {
+                XRRFreeOutputInfo(oi);
+                continue;
+            }
+
             if (ci->rotation == RR_Rotate_90 || ci->rotation == RR_Rotate_270)
             {
                 widthMM  = oi->mm_height;


dougbinks:
I've confirmed that `XRRGetCrtcInfo` can return `NULL`, and the fix looks reasonable. I'll test this but as the crash is occasional I may not be able to confirm if this fix does not result in other issues if `NULL` is returned - from the looks the polling doesn't add the monitor to the current array and thus a disconnect is generated or nothing happens if a disconnect has already occurred, which should be fine and certainly better than a crash.

As per the [Contributing guide](https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-bug-fix) we will also need:

1. Change log entry in `README.md`, describing the incorrect behavior
2. Credits entries in `CONTRIBUTORS.md` for all authors of the bug fix

dougbinks:
Many thanks for the report and PR, I have manually merged this in 4df5129529b1af57d04a51f09145da5079822c3e

PR #2772, opened 2025-09-18
Wayland: Implement support for `glfwSetCursorPos`

IMS212:
This uses the new [pointer_warp_v1](https://wayland.app/protocols/pointer-warp-v1) protocol to implement full support for `glfwSetCursorPos`.

This protocol is still in staging, and support was added to KDE Plasma 6.5 today and Gnome 49 a few days ago. Support is still pending for most other compositors.

Should supersede #2496 for supported compositors.


shrobbyy:
Mojang changed Minecraft in yesterdays snapshot to run on Wayland natively. So this is why the cursor doesn't center anymore when opening a menu or container. Honestly, this makes the game kinda unplayable for me, because it messes with my muscle memory. I hope this is getting merged pretty quickly so the game can work as intended again on Wayland.

IMS212:
This will not affect Minecraft; LWJGL uses a fork of GLFW with this patch already planned to be merged.

huantianad:
The popular patch that previously fixed this behavior using zwp_locked_pointer_v1_set_cursor_position_hint has some interesting behavior to update the position the frame after the request. Is this not necessary with the new protocol? I believe this was originally done to somehow improve the reliability of the method; sometimes I do notice that the pointer lock doesn't work for some reason.

https://github.com/glfw/glfw/pull/2496, https://github.com/BoyOrigin/glfw-wayland/blob/main/patches/0003-Implement-glfwSetCursorPosWayland.patch

PR #2775, opened 2025-09-28
Make dependency include directories public

cqundefine:
This fixes an issue in a rare case where both dependencies and glfw are installed in different non-standard locations. This causes include errors when using glfw with CMake.


dougbinks:
I'm a bit perplexed as to why this would resolve include errors in GLFW since the `private` keyword will populate the `INCLUDE_DIRECTORIES` property of `<target>`, and `public` just means to populate the `INTERFACE_INCLUDE_DIRECTORIES` [according to the CMake target_include_directories documentation](https://cmake.org/cmake/help/latest/command/target_include_directories.html).

What include errors are you seeing?

cqundefine:
The issue is that the GLFW native header includes X11 code on Unixes and in obscure cases like mine when X11 code is in a location that is not searched by default (`/usr/local/...` in my case) and GLFW is installed in a different place, but also not searched by default (in my case it's a subdirectory of my home folder), the CMake config does not pick up the X11 include directories by default, which causes include errors for files like `X11/Xlib.h`. This change causes CMake to add the X11 directories to the CMake config files which resolves the include issues without having to manually add the X11 include path.

dougbinks:
Sorry, I should have been more specific in my request. Could you copy and paste the include errors you are seeing?

Issue #2600, opened 2024-08-04
glfwSetWindowAttrib() for GLFW_MOUSE_PASSTHROUGH causes glfwWaitEvents() to return, as if there's a new event (WINDOWS 11)


MirrasHue:
Hey, I'm experimenting with GLFW and ImGui. What I have now is glfwWaitEvents() on the main thread, and update and rendering in a separate thread, it works great for not blocking the execution while resizing, moving the window or holding the minimize, maximize or close buttons.

I noticed that while rendering ImGui, glfwWaitEvents() never waits, but exits immediately, even when I'm not providing any inputs.

Upon some investigation, I found the problem: ImGui is calling glfwSetWindowAttrib() every frame, and this function calls SetWindowLongW(), which seems to be emitting an unknown event (couldn't catch it in any callback), causing glfwWaitEvents() to not wait for user input.

Taking a look at the implementation of ImGui_ImplGlfw_UpdateMouseData (which calls glfwSetWindowAttrib), there's no check to see if the mouse passthrough attribute is already set to the desired state (enabled / disabled).

I know this check could also be implemented in _glfwSetWindowMousePassthroughWin32() itself, but this event being fired by SetWindowLongW() got me curious.

dougbinks:
I am using `glfwWaitEvents` with Dear ImGui and have not observed this on Windows 11 - although `glfwSetWindowAttrib` is being called every frame it isn't causing any events to be generated, and `glfwWaitEvents` doesn't return.

I am wondering if this is caused by some third party DLL monitoring the window changes - it might be worth inspecting your running exe for attached DLLs, which you can do by adding a breakpoint in Visual Studio and  debug modules window (sorted by path to look for any hooked DLLs).

MirrasHue:
Hi @dougbinks, thanks for the reply.

Did you move your rendering to a separate thread?

I'll try to provide a simple minimal example:

```cpp
#include <stdio.h>
#include <thread>

#include <glad/glad.h>
#include <glfw/glfw3.h>

#include <imgui_impl_glfw.h>
#include <imgui_impl_opengl3.h>
#include <imgui.h>

static GLFWwindow* windowHandle = nullptr;
static bool running = true;

void update()
{
    glfwMakeContextCurrent(windowHandle);

    while(running)
    {
        glClearColor(0.1f, 0.1f, 0.1f, 1.f);
        glClear(GL_COLOR_BUFFER_BIT);
        
        ImGui_ImplOpenGL3_NewFrame();
[...]

Issue #1911, opened 2021-06-09
Wayland glfwWaitEvents() not waiting


FrostKiwi:
On FreeBSD 13, [Sway](https://github.com/swaywm/sway/) with [glew](https://www.freshports.org/graphics/glew/) and [glew-wayland](https://www.freshports.org/graphics/glew-wayland/) packages used, GLFW refuses to wait at glfwWaitEvents(), even though no events are (seemingly) happening.
If I compile the same program with X11 instead, glfwWaitEvents() does wait, even though neither my code, nor the Window manager changed. The only change happening: glew instead of glew-wayland is used and GLFW is used instead of GLFW-Wayland.

FrostKiwi:
Compiling with the latest git master (076bfd55be45e7ba5c887d4b32aa03d26881a1fb) the issue persists.
glfwWaitEvents() is just totally ignored, whilst using non-wayland GLFW via XWayland still honors glfwWaitEvents()

FrostKiwi:
Tested with the new 3.3.5 release again. (Built from 3.3-stable branch, with the latest commit being from the 2021.10.28-22:16:41 ). Same deal, glfwWaitEvents() ignores the wait and blasts through at full speed with Wayland, much to the dismay of my electricity bill.
Sry to sound like a broken record, just want to keep the information up-to date^^

linkmauve:
Hi, I can’t test on FreeBSD but I tried to reproduce on Linux and got no issue, it properly `poll()`s until either the Wayland socket has events, or one of the two `timerfd`s has events.

Would you be able to see which of the three branches is waking up on your system, for instance using `gdb`?

FrostKiwi:
@linkmauve 
Didn't know how to compile GLFW in debug mode, so I added a puts("Event #"); to each branch resulting from poll(), that I think you meant.
See here:
<details><summary>wl_window.c handleEvents function</summary>

```
static void handleEvents(int timeout)
{
    struct wl_display* display = _glfw.wl.display;
    struct pollfd fds[] = {
        { wl_display_get_fd(display), POLLIN },
        { _glfw.wl.timerfd, POLLIN },
        { _glfw.wl.cursorTimerfd, POLLIN },
    };
    ssize_t read_ret;
    uint64_t repeats, i;

    while (wl_display_prepare_read(display) != 0)
        wl_display_dispatch_pending(display);

    // If an error different from EAGAIN happens, we have likely been
    // disconnected from the Wayland session, try to handle that the best we
[...]

Issue #685, opened 2016-01-06
 glfwWaitEvents


waqasm86:
glfwwaitevents() does not seem to be working in the latest version of glfw3 as compare to it's previous versions. 
I have been trying to run or debug the example of rotating triangle with glfwWaitEvents() command, but the triangle keeps rotating along with flickering rather that just supposed to stay stationary with no input via keyboard or mouse.

elmindreda:
Which version of what OS are you using?

waqasm86:
windows 8.1

waqasm86:
windows 8.1 with visual studio 2015

tombsar:
Unable to reproduce using Windows 8.1, MSVC 2015 (update 1), and current glfw master.

elmindreda:
Unable to reproduce.

haikarainen:
Unable to reproduce using Windows 10, MSVC 2015 (update 1) with either current glfw master or 3.1.2 NuGet package

@waqasm86 Post your code

elmindreda:
Still unable to reproduce.

elmindreda:
@dougbinks I'm not opposed to that if people find it sufficiently useful.

tombsar:
Unless this can be reproduced, I suggest this issue be closed. IMO the most likely cause is a third-party piece of software sending unwanted messages to the window (could be anything from a screen recorder to a Windows misfeature). We could maybe create an enhancement issue to discuss designing something to filter these out as dougbinks suggests, probably targeting 3.4.

dougbinks:
I can reproduce this on some machines, and I've been intending to look into this but just haven't had the time yet given that I think it's relatively low priority.

tombsar:
@dougbinks Are you able to reproduce with any of the glfw test/example programs, or maybe a very simple program that I can copy? I will try again to reproduce and debug now that I'm on Windows 10.

dougbinks:
This happens on any program using glfwWaitEvents() - any message sent to a window will cause the glfwWaitEvents() to return since the underlying message procedure handles the message.

However outside of user input this only happens on some machines. I don't have the ones which do this in front of me right now and don't recall the message being processed, but you could create a program which sends a message to your glfw program using SendNotifyMessage  to test if you need.

dougbinks:
I have a potential prototype implementation which only exits a wait if a GLFW relevant event is received. This is on my branch: https://github.com/dougbinks/glfw/tree/wait-events

[...]

PR #2792, opened 2025-11-22
glfwWaitEvents (and glfwWaitEventsTimeout) which only return if GLFW events are encountered (or timeout)

dougbinks:
This is a draft PR with a fix for issues #685, #1911 and #2600 which works by only returning from the wait if either a GLFW relevant event is encountered or the timeout is reached if set.

## Current status of backends

- Win32: implemented, tested as working as before but currently I do not have a repro of  #685
- Wayland: implemented and tested as resolving an issue with the same characteristics as #1911 by @memononen encountered on Fedora KDE (private communication).
- X11: implemented and tested as working as before
- Cocoa: not yet implemented

 I do not have a Mac machine at the moment as mine has died, but I will see about using an online remote desktop service to implement the Cocoa branch if this PR architecture is deemed suitable.

## Design

This PR adds a boolean `newEventsRcvd` to the `_glfw` global struct. This is set to `GLFW_FALSE` at the start of a poll/wait event call, and set to `GLFW_TRUE` if a GLFW relevant event is received, including one from a `glfwPostEmptyEvent`. The majority of the code for `newEventsRcvd` state management is in cross platform code.

Each platform adds a loop to the 2 wait event calls, reducing the timeout by the passage of time where required (*note to self: POSIX implementation may need timeout adjustment*). This code could potentially be moved to the cross platform code.

Each platform needs to handle the `glfwPostEmptyEvent` on the same thread as the `glfwWaitEvents*` call (the main thread by design).
- Wayland adds a handler for the `wl_display_sync` display sync done
- X11 adds code to the `drainEmptyEvents` pipe function
- Win32 adds a `WM_NULL` case to the helper window procedure
- Cocoa will need to handle the `NSApplicationDefined` `NSEvent` for this.

## Questions

1. Should we add an init hint to control this feature, and if so should it default to ON or OFF?
2. Is a boolean a reasonable choice, or should we move to a `uint32_t` or `uint64_t` event counter?
    2a. Should this be a global event counter or reset every event loop?
    2b. With atomics it might actually be possible to move the entire implementation to cross platform code. Whilst C99 does not have atomic integers I am well versed in their cross platform implementation and can add these if needed.
4. Do we expose this to applications, so they can know if the timeout was reached, or an event was encountered (or how many events if we move to a counter)?
5. Have I implemented all event types?

## TODO

See [Contributing a bug fix documentation](https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-bug-fix) for general requirements for a bug fix contribution, and if we add functions to control/query this then also the feature list.

- Cocoa implementation
- Answer questions above
- Add changelog entry
- If we extend the API then we need news page entry, documentation, reference documentation etc. as per the guide.


sirnacnud:
@dougbinks I went ahead and did the cocoa implementation based on how you implemented the other platforms. You can find it here on my fork: https://github.com/sirnacnud/glfw/commit/0eaf92b220ecb73aabdca6eca2ab6e7902b93ad7. Feel free to bring it into your PR.

It would be good to get some progress on this PR. Let me know if you need further help with testing, as I have tested the cocoa implementation.

dougbinks:
@sirnacnud Many thanks, I'll take a look at this as soon as I get time.

vursc:
Hello, I ran into #1911. Two things:
- In "current status of backends", #1911 is the Wayland issue, #685 is the Win32 one, and they're flipped. Also, I can't find the "memononen Fedora KDE" reference.
- `_glfwInputJoystickAxis` in `src/input.c`  should also set `_glfw.newEventsRcvd`. AFAICT all other event types are implemented. (Separate issue: should we add callbacks for joystick movement and/or `_glfwInputMonitorWindow`?)

My suggestions:
- This is a bugfix, so it should be always enabled. If we missed some event types we better get a report and fix it sooner than later.
- We should not expose `newEventsRcvd` to users, as it's not useful: users probably don't care whether an event was received if they don't set a callback and don't read the value (using glfwGetKey etc). A bitmask of received event types would be useful though.

dougbinks:
> In "current status of backends", https://github.com/glfw/glfw/issues/1911 is the Wayland issue, https://github.com/glfw/glfw/issues/685 is the Win32 one, and they're flipped. Also, I can't find the "memononen Fedora KDE" reference.

I've fixed that, thanks for noting (the conversation with Mikko was on Mastodon but I can't find it anymore).

> _glfwInputJoystickAxis in src/input.c should also set _glfw.newEventsRcvd. AFAICT all other event types are implemented. (Separate issue: should we add callbacks for joystick movement and/or _glfwInputMonitorWindow?)

I'll add the `_glfwInputJoystickAxis` logic when I get time. Separate issues should be separate (I think there already is an issue for this).

dougbinks:
I've merged @sirnacnud's cocoa implementation (I do not have access to a Mac to test this).

I plan to fix the conflicts and update this PR as soon as possible, but am somewhat overloaded at the moment.

dougbinks:
I have merged main, fixed the conflicts and done some basic testing on Windows and Linux, but cannot on macOS as I don't have access to a working Mac.

PR #2805, opened 2025-12-25
Set dock icon on MacOS

annes12345678910:
Currently on MacOS, GLFW throws the error _"Regular windows do not have icons on macOS",_ and I use raylib so I wanted to fix that. Now GLFW sets the dock icon instead of throwing an error.

Here's an example with raylib, compiled with my changes:
Screenshot 2025-12-25 at 8 13 08 PM

## Summary of commits
* Added nullptr alias
* included more libraries: Cocoa and the objc runtime
* Added private function __CocoaSetDockIcon
* Changed the function _glfwSetWindowIconCocoa

skibidi

### DIFF
--- a/src/cocoa_window.m
+++ b/src/cocoa_window.m
@@ -36,6 +36,8 @@
 #include <string.h>
 #include <assert.h>
 
+#define nullptr ((void*)0) // i like c++
+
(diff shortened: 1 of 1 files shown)


ws909:
- #2293

annes12345678910:
shii

PR #2810, opened 2026-01-04
Round halfway away from zero per fractional-scale-v1 protocol spec

road2react:
When using glfw with Minecraft (compiling glfw from source and using Prism Launcher's Workarounds -> "Use system installation of GLFW") for native Wayland support, the framebuffer is slightly smaller than the actual resolution for certain sizes and scale levels (for example, 4K at 132.5% scale, which is 159/120 for `wp_fractional_scale_v1::preferred_scale`), causing the compositor (tested with KDE Plasma) to slightly upscale the window, making it blurry.

This PR fixes the framebuffer size calculation to match [the spec](https://wayland.app/protocols/fractional-scale-v1) which states "For toplevel surfaces, the size is rounded halfway away from zero."

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -461,8 +461,9 @@ static void resizeFramebuffer(_GLFWwindow* window)
 {
     if (window->wl.fractionalScale)
     {
-        window->wl.fbWidth = (window->wl.width * window->wl.scalingNumerator) / 120;
-        window->wl.fbHeight = (window->wl.height * window->wl.scalingNumerator) / 120;
+        // Round halfway away from zero per fractional-scale-v1 protocol spec
+        window->wl.fbWidth = (window->wl.width * window->wl.scalingNumerator + 60) / 120;
+        window->wl.fbHeight = (window->wl.height * window->wl.scalingNumerator + 60) / 120;
     }
     else
     {


GalaxySnail:
FWIW: <https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/309>

umbrageodotus:
> While this [formula] is correct for non-negative w/h, this does not work for the stable algorithm where the coordinates can be negative. There the formulas should be `W = (w*s + sign(w)*60) / 120`; `H = (h*s + sign(h)*60) / 120.

\- By one of the maintainers - text adapted for clarity via insertions (marked with []).

The behavior introduced by this PR is still non-spec compliant.

umbrageodotus:
[glfw-bias-fix.patch](https://github.com/user-attachments/files/26456162/glfw-bias-fix.patch)
This version of the patch works.

zeux:
Just noting I ran into the same problem with a slightly more conventional scale and a slightly less conventional resolution; the same fix (either the PR or the patch) resolves the problem correctly. At 5120x2160 and 150% DPI scale when using `niri` as a compositor, a fullscreen window is `3413.33...` wide, which gets rounded to `3413` and then passed to GLFW; `3413 * 1.5` = `5119.5` which currently is truncated to 5119. Rounding to nearest results in correct width of `5120`. Would be great to see this merged!

(I don't understand why negative width/height is an important consideration, and it seems like GLFW assumes non-negative values and `resizeWindow` in `wl_window.c` clamps the inputs)

umbrageodotus:
I didn't go looking, but if someone is sure clamping is done, then the old formula can be used.

PR #2824, opened 2026-02-16
Added Emscripten external platform (comment)

ypujante:
This is the PR that was mentioned in this [discussion thread](https://discourse.glfw.org/t/glfw-3-4-external-platform/2572/1).

This PR "reserves" the Emscripten external platform in a comment to make sure this constant is not used in the future.


dougbinks:
Thanks.

For future reference this was discussed here: https://discourse.glfw.org/t/glfw-3-4-external-platform/2572

PR #2829, opened 2026-02-21
Wayland: Don't segfault

a-usr:
Since the compositor may send key modifiers before the keymap event, we may end up making libxkbcommon dereference a nullpointer for a state, which is... inconvenient.


a-usr:
.. Or maybe the segfault occurs due to some other reason after all?

Issue #359, opened 2014-09-28
Function to return key state array


emanuelpalm:
When dealing with keyboard key combinations, I've found it very convenient to be able to look in a keyboard state table rather than having to get each relevant key. This is [something you can do with SDL2](http://hg.libsdl.org/SDL/file/9ecf775ead1b/src/events/SDL_keyboard.c#l843), and is something I believe could be of benefit to the GLFW3 API.

Using it could look something like this:

    vector2_t velocity = {.x = 0.0, .y = 0.0};
    const char *keyboard = glfwGetKeyboardState(window);
    if (keyboard[GLFW_KEY_UP]    || keyboard[GLFW_KEY_W]) velocity.y -= 1.0;
    if (keyboard[GLFW_KEY_DOWN]  || keyboard[GLFW_KEY_S]) velocity.y += 1.0;
    if (keyboard[GLFW_KEY_LEFT]  || keyboard[GLFW_KEY_D]) velocity.x -= 1.0;
    if (keyboard[GLFW_KEY_RIGHT] || keyboard[GLFW_KEY_A]) velocity.x += 1.0;
    return vector2_normalize(velocity);

I read through the GLFW source code and started hacking on a possible implementation, but then I realized that the whole sticky key deal makes things more complicated. As `_GLFW_STICK` keys may not be reported to the GLFW user, those key states have to be changed to either `GLFW_PRESS` or `GLFW_RELEASE` when calling the function.

The solutions I've thought of are to either (1) prevent the user from using the function if sticky keys are enabled, (2) arbitrarily set all `_GLFW_STICK` keys to `GLFW_PRESS` when calling the function, or (3) somehow mimic the behavior of the `glfwGetKey()` function.

Is this a function people would use, and whats thought do you have regarding its implementation in relation to sticky keys?

elmindreda:
```c
#define keyboard(key) glfwGetKey(window, key)
vector2_t velocity = {.x = 0.0, .y = 0.0};
if (keyboard(GLFW_KEY_UP)    || keyboard(GLFW_KEY_W)) velocity.y -= 1.0;
if (keyboard(GLFW_KEY_DOWN)  || keyboard(GLFW_KEY_S)) velocity.y += 1.0;
if (keyboard(GLFW_KEY_LEFT)  || keyboard(GLFW_KEY_D)) velocity.x -= 1.0;
[...]

PR #2833, opened 2026-03-03
Wayland: Add #ifndef guard for _GNU_SOURCE in wl_window.c

tomas-mraz:
Hi, in the GLFW Go/Golang binding project, there is [issue #359](https://github.com/go-gl/glfw/issues/359), which is trying to solve the warning message "wl_window.c:29: **_GNU_SOURCE" redefined**" during building for Wayland.

```
../../go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20250301202403-da16c1255728/glfw/src/wl_window.c:29: warning: "_GNU_SOURCE" redefined
   29 | #define _GNU_SOURCE
```

It happens because go-gl/glfw includes like ...

```
  #include "glfw/src/wl_init.c"     // 1. #include "internal.h" → standard headers → features.h ✓                                                                                                                                                                                                               
  #include "glfw/src/wl_monitor.c"  // 2. #include "internal.h" → features.h already in cache
  #include "glfw/src/wl_window.c"   // 3. #define _GNU_SOURCE ← too late!
```
... the “features.h” file expects “_GNU_SOURCE” (to internally activate “__USE_GNU”). Then the variable was added to the compiler's CFLAGS. And it leads to the warning about redefined _GNU_SOURCE in the file /src/wl_window.c during compilation.

This PR adds a guard for the variable.

```
  #ifndef _GNU_SOURCE                                                                                                                                                                                                                                                                                           
  #define _GNU_SOURCE
  #endif
```

I believe it means no harm for you, and it helps on our side.
Thank you very much

This PR allows editing by maintainers.

### DIFF
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -24,7 +24,9 @@
 //
 //========================================================================
 
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
 
 #include "internal.h"
 


tomas-mraz:
Hi, I dug some more ...
It's cgo's generated export prologue — it needs <stddef.h> for the ptrdiff_t type used in the _GoString_ struct, and it internally loads <features.h> ... and it is a place where _GNU_SOURCE is needed. This behaviour is hardcoded in Go itself ... for the build of Go GLFW, the variable must be in CFLAGS. I don't see any other way to fix this without increasing the maintenance burden in the build flow.

Thank you for your time.

elmindreda:
That makes sense, but wouldn't this needs fixing in `posix_poll.c` as well?

tomas-mraz:
> That makes sense, but wouldn't this needs fixing in `posix_poll.c` as well?

No, it isn't. I fork it, test it, use it just like is in the PR. So it is OK as is. Thanks

PR #2834, opened 2026-03-07
Add glfwStartInteractiveMove for compositor-managed window drag

Rounak-Paul:
Adds glfwStartInteractiveMove(GLFWwindow*) — a new API that initiates a compositor-managed interactive window move, triggered by a client-side title bar press.


dougbinks:
I think I understand why this is a good idea for an enhancement, but could you explain your reasoning behind this and how it helps (briefly)?

Also please note the [contributing a feature guide requirements](https://github.com/glfw/glfw/blob/master/docs/CONTRIBUTING.md#contributing-a-feature). To start with it would be worth adding the `news.md` entry as this would be a good way to explain the feature,

PR #2836, opened 2026-03-10
Fix memory leak of un-cleaned up callbacks.

crazydays:
Copying the logic from libSDL fix.
https://github.com/libsdl-org/SDL/pull/8060/files


elmindreda:
Thank you for the bug report! This should be fixed now with the re-implementation in bb8048122a8ef5a12e60d2ab3c8586961ae2366c.

PR #2846, opened 2026-04-10
Change MACOSX_DEPLOYMENT_TARGET version

mosfet80:
Updated MACOSX_DEPLOYMENT_TARGET to 14.6 for build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
         timeout-minutes: 4
         env:
             CFLAGS: -Werror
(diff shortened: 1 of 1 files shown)


dougbinks:
What is your rational for this change? `MACOSX_DEPLOYMENT_TARGET` is the **minimum** version of macOS on which the target binaries are to be deployed, and GLFW supports macOS 10.11 and later as per the version set.

mosfet80:
@dougbinks : OS X 10.11 (El Capitan) support ended 10 years ago. 14(sonoma) is the oldest suported

PR #2855, opened 2026-05-01
Wayland support

yokljo:
I changed a lot of the Wayland backend to fix a lot of problems:

-     DPI is now calculated correctly, with fractional scaling
-     Undecorated windows are draggable via the title bar
-     Undecorated windows are resizable via their edges
-     Resizing edges are disabled for maximised windows
-     The mouse cursor appears as a resize cursor while hovering the edge of the window
-     There is only one function that decides what shape the cursor should be, avoiding glitchiness
-     Native cursor shapes are preferred

And more!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ tests/triangle-vulkan
 tests/window
 tests/windows
 
+# Hazel-specific
+bin
+bin-int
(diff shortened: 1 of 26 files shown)


yokljo:
Sorry, I accidentally selected the wrong target

PR #2857, opened 2026-05-14
Fix building GLFW on TCC and others GCC compatible compiler

GermanAizek:
@elmindreda,

Clang, GCC, TCC, TDM-GCC successful building.

My experiments on 2 CPUs Xeon E5-2699v3 (72 threads): I am surprised by the compilation speed on TCC, as well as equivalent resource consumption on monitoring

    TCC 0.3s compilation (difference ~14x)
    Clang 4.1s compilation
    GCC 5.2s compialtion

I think Gentoo Linux users will idolize Fabrice Bellard.

https://en.wikipedia.org/wiki/Gentoo_Linux

Install tcc compiler easy on Debian/Ubuntu `apt install tcc`

Wiki: https://en.wikipedia.org/wiki/Tiny_C_Compiler
Official site: https://bellard.org/tcc/

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -37,10 +37,20 @@ add_executable(triangle-opengles WIN32 MACOSX_BUNDLE triangle-opengles.c ${ICON}
 add_executable(wave WIN32 MACOSX_BUNDLE wave.c ${ICON} ${GLAD_GL})
 add_executable(windows WIN32 MACOSX_BUNDLE windows.c ${ICON} ${GLAD_GL})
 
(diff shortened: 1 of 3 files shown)


GermanAizek:
Actions Windows failed by reason:
```

Run cmake -B build-win32-shared-x86 -G "Visual Studio 17 2022" -A Win32 -D BUILD_SHARED_LIBS=ON
CMake Error at CMakeLists.txt:3 (project):
  Generator

    Visual Studio 17 2022

  could not find any instance of Visual Studio.
```

PR #2867, opened 2026-06-16
Add a glfwGetKeyNameModifiers function

### LINKED ISSUE #2685: Obtaining Key Names with Modifiers
Branching off from #1502,

I'm trying to build a virtual on-screen keyboard. Naturally, I would like to display the characters that will actually be input when a virtual key is pressed. To do this I currently use `glfwGetKeyName` for the corresponding key, and visualise the key being pressed when it's input on the actual keyboard. This works well.

What doesn't work well is when modifiers come into play. For instance, if I press Shift, the alphabetic keys should become upper case. This is not hard to do myself, of course, but different layouts have vastly different translations for shifted keys that aren't alphabetics, not to mention additional modifiers like AltGr or Shift+AltGr.

Currently I see no way to obtain a representation of what I will get for a text callback given a key and a set of modifiers. The OS should provide a way to retrieve this translation, but GLFW does not expose a similar API. Perhaps a `glfwGetKeyNameWithModifiers` would be appropriate here.

Shinmera:
This is work to resolve #2685. As it stands the current PR is not yet done and should not be considered ready for merge. 

~~I thought I'd put it out there since I'm already running into issues that others might have insight into, namely on Linux/Wayland and Win32:~~

~~On Wayland, xkb_keymap_key_get_syms_by_level is used, which takes a "shift level" to signify which "level" of the pressed keycode to translate to a specific emitted symbol. What I can't figure out is this: I don't see any way whatsoever to translate a modifier set to a level. The only function in the entire libxkbcommon API that returns a level set is only used to determine the level of a specific key in the current keyboard state, which isn't useful for this.~~
Resolved using a potentially slow level scanning method.

~~On Win32 neither `MapVirtualKeyW` nor `MapVirtualKeyExW` allow passing on a modifier set, and I haven't been able to find out an API that does allow one to do this. Will dig further, though.~~ Resolved by expanding the keyname table.

On X11 there's `XkbLookupKeySym` which promises to do what we need, and on Cocoa the standard function `UCKeyTranslate` already accepts a modifiers flag, so these should be fine. ~~For the null backend I'll see about adding the standard US-ANSI keyboard layout levels.~~

Need to just test and document this now and it should be good to merge pending review

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -313,6 +313,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Yukari Hafner (Shinmera)
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -310,6 +310,7 @@ video tutorials.
  - Jonas Ådahl
  - Lasse Öörni
  - Leonard König
+ - Yukari Hafner (Shinmera)
  - All the unmentioned and anonymous contributors in the GLFW community, for bug
    reports, patches, feedback, testing and encouragement
 
--- a/README.md
+++ b/README.md
@@ -120,6 +120,9 @@ information on what to include when reporting a bug.
 
 ## Changelog since 3.4
 
+ - Added `glfwGetKeyNameModifiers` to obtain the native key translation including
+   active modifier sets in order to accurately display key state and handle key
+   chords in the presence of shift and other modifiers (#2685)
  - Added `GLFW_UNLIMITED_MOUSE_BUTTONS` input mode that allows mouse buttons beyond
    the limit of the mouse button tokens to be reported (#2423)
  - Added `glfwGetEGLConfig` function to query the `EGLConfig` of a window (#2045)
--- a/README.md
+++ b/README.md
@@ -120,8 +120,9 @@ information on what to include when reporting a bug.
 
 ## Changelog since 3.5
 
-None.
-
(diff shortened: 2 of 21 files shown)


Shinmera:
I've implemented a solution that should work for Wayland, and updated the Null backend to allow for shifted keys based on the en-US ANSI layout.

The only platform remaining now is Win32, and then of course a bunch of testing of each platform. I'd especially appreciate help testing MacOS, as I don't currently have access to a Mac machine.

Shinmera:
Okey, added documentation and attribution. Also changed the events test to use the modifier keys. Tested on Wayland to work, will test X11 and Windows soon. Don't have a mac handy atm to test.

I think this should be ready for review in the meantime.

Shinmera:
Okey, tested on X11, Wayland, and Win32.

X11 works flawlessly, even translating things like Ctrl+Alt+a, which on my keyboard produces ä. On Wayland I'm having some mixed results, using a standard keymap layout it seems to translate things correctly, but using my personal, custom map, it fails to translate some of the keys. I figure the issue there is Wayland/xkb rather than GLFW though.

Win32 seems to work fine for basic shifted keys, but mappings with Ctrl+Alt appear bugged:


I have no clue why that would be, it seems like the unicode translation breaks or something?

Shinmera:
Actually it seems to produce the same broken output as the character callback, so I think it's working right, just not rendering right.

EeeRika-1005HA:
Tested on macOS Golden Gate 27.0 with both DE and US layouts. All keypresses behaved as expected with any number of modifiers, as far as I can tell.

Shinmera:
Great, thanks a lot! I think that covers all the platforms, then :)

@elmindreda I'd very much appreciate a review of the changes!

Shinmera:
Rebased on top of master

Issue #2523, opened 2024-03-29
Turning on GLFW_CURSOR_DISABLED when moving the mouse will produce huge mouse position offsets


Intro1997:
# Problem
Hello, I'm using the glfw 3.4 version by building with CMake. I found that when you set `GLFW_CURSOR_DISABLED` by `glfwSetInputMode()`in moving mouse, sometimes you will find that there is a big offset between the cursor position obtained from `glfwGetCursorPos` before and after `glfwPollEvents` is called. I test it on macOS 14.4, Windows 11 and Ubuntu 22.04 LTS, only ubuntu doesn't have this issue. 

I found this issus may as same as #1523 and #2411, and when I remove ` _glfw.platform.setCursorPos(window, width / 2.0, height / 2.0);` in `_glfwCenterCursorInContentArea` at `input.c` file, both macOS and Windows 11 are correct.

# Reproduce Problem
I push [a test project](https://github.com/Intro1997/ICVp_test) to reproduce this issue. You can play with tag [v0.0.1-cursor_bug](https://github.com/Intro1997/ICVp_test/releases/tag/v0.0.1-cursor_bug) to check this issue.

And I also record a video to show what the issue looks like (using Bongo cat Mver to show my mouse and keyboard):

In this video, I print my cursor position by `glfwGetCursorPos` before and after `glfwPollEvents` is called when cursor is in `GLFW_CURSOR_DISABLED` state:
```cpp
    if (glfwGetInputMode(window, GLFW_CURSOR) == GLFW_CURSOR_DISABLED) {
      glfwGetCursorPos(window, &cursor_x, &cursor_y);
      std::cout << "before poll cursor pos (" << cursor_x << ", " << cursor_y << ")\n";
    }
    glfwPollEvents();
    if (glfwGetInputMode(window, GLFW_CURSOR) == GLFW_CURSOR_DISABLED) {
      glfwGetCursorPos(window, &cursor_x, &cursor_y);
      std::cout << "after poll cursor pos (" << cursor_x << ", " << cursor_y << ")\n";
    }
```

https://github.com/glfw/glfw/assets/29827323/4242f9e7-6f76-4aaa-ac0c-42326ba899a9

Let me describe what happens and when does this happen in video:

1. Show the basic interaction. (0:04~0:11)
[...]

PR #2871, opened 2026-07-02
Fix random jump when cursor is disabled

cpichard:
Hi,

Last night I have been looking again at bug #2523 where moving camera with cursor disabled had some random jumps, which was really annoying. Patching my app as a solution didn't feel great nor moving to SDL.
I used Fable this time, it made a repro, found the root cause and a simple patch. 
I have limited knowledge of that part of the code, but to my understanding the patch sounds correct, so I am proposing it here. virtualCursorPos is reset/set before the platform enter/exit warp which creates a discontinuity that shows up when there is a transition of state. The fix is to avoid setting virtualCursorPos since its absolute value is not needed anyway, and reset the cursor position on output. Here is the patch explanation from Fable:

```
 usdtweak patch (glfw/glfw#2523): upstream re-reads the OS 
 cursor position into virtualCursorPos here — before the 
 platform performs its enter/exit warps, and without firing 
 the cursor position callback — so apps tracking the 
 position from callbacks (e.g. ImGui) desync and see the 
 discontinuity replayed as one giant delta on the next 
 motion event. Instead: 
  - Entering GLFW_CURSOR_DISABLED: do nothing. The virtual 
    cursor continues from the last reported position (its 
    absolute base is meaningless by definition) and the 
    platform's warp-to-center is already hidden from the 
    delta stream. No rebase, no event, no discontinuity. 
  - Leaving GLFW_CURSOR_DISABLED: the platform warps the 
    cursor back to the pre-capture position — a real 
    teleport. Apply the mode first, then report the settled 
    position through _glfwInputCursorPos so the callback 
    fires and consumers resync. 
 - Transitions not involving DISABLED don't move the 
    cursor; nothing to do. 
```

I have been testing it yesterday night in my application https://github.com/cpichard/usdtweak/tree/jumpingcursorv2 and it works fine so far, I'll test it in depth this week-end when I have more time.

[repro.tar.gz](https://github.com/user-attachments/files/29584900/repro.tar.gz)

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -608,12 +608,17 @@ GLFWAPI void glfwSetInputMode(GLFWwindow* handle, int mode, int value)
             if (window->cursorMode == value)
                 return;
 
+            const int oldMode = window->cursorMode;
             window->cursorMode = value;
 
-            _glfw.platform.getCursorPos(window,
-                                        &window->virtualCursorPosX,
-                                        &window->virtualCursorPosY);
             _glfw.platform.setCursorMode(window, value);
+
+            if (value != GLFW_CURSOR_DISABLED && oldMode == GLFW_CURSOR_DISABLED)
+            {
+                double xpos, ypos;
+                _glfw.platform.getCursorPos(window, &xpos, &ypos);
+                _glfwInputCursorPos(window, xpos, ypos);
+            }
             return;
         }
 


cpichard:
Hi, I have been testing this patch and the jumping cursor can still happen with the mouse, so this patch doesn't correct it. Closing this PR.

Issue #2519, opened 2024-03-25
Dereference of NULL error according to GCC's -fanalyzer


Nostress767:
OS and version: Linux archlinux 6.8.1-arch1-1 <span>#</span>1 SMP PREEMPT_DYNAMIC Sat, 16 Mar 2024 17:15:35 +0000 x86_64 GNU/Linux
Compiler version: gcc (GCC) 13.2.1 20230801
Release or commit: bf945f1213728a98f7647380616f9cff9f6b3611

I run a small project using the flags "-fanalyzer -Werror -Wall -Wpedantic". When I tried adding GLFW to it, GCC's static analyzer flared up and gave me some errors. I'm not sure what they affect or how they interact with other things, but I tried replicating the issue from the latest commit with a minimal code example just to make sure and they still popped up.

I created the Makefile and main.c in the root directory of this project and ran `make`

Makefile:
```makefile
# Disable built-in rules and variables
MAKEFLAGS += --no-builtin-rules
MAKEFLAGS += --no-builtin-variables

CC       := gcc
CPPFLAGS := -Iinclude -D_GLFW_X11
CFLAGS   := -fanalyzer -c
LDFLAGS  := -lm
EXE      := Example

OBJS := main.o
OBJS += src/context.o src/init.o src/input.o src/monitor.o src/platform.o src/window.o
OBJS += src/posix_module.o src/posix_thread.o src/posix_time.o src/posix_poll.o
OBJS += src/egl_context.o src/glx_context.o src/osmesa_context.o
OBJS += src/linux_joystick.o src/xkb_unicode.o
OBJS += src/vulkan.o
OBJS += src/x11_window.o src/x11_init.o src/x11_monitor.o
OBJS += src/null_init.o src/null_joystick.o src/null_monitor.o src/null_window.o

all: clean $(EXE)

$(OBJS) : %.o : %.c
	$(CC) $(CPPFLAGS) $(CFLAGS) $< -o $@

$(EXE): $(OBJS)
	$(CC) -o $@ $(OBJS) $(LDFLAGS)

clean:
	rm -f src/*.o $(EXE)
```

main.c:
```c
#include "GLFW/glfw3.h"

#include <stdbool.h>

int main(int argc, char* argv[])
{
	glfwInit();

	GLFWwindow *window = glfwCreateWindow(800, 600, "Test", NULL, NULL);

	glfwMakeContextCurrent(window);

	while(!glfwWindowShouldClose(window)){
		glfwPollEvents();

[...]

PR #2883, opened 2026-07-29
fix null pointer derefence in x11_monitor.c

Gilqamesh:
Associated issue(s):
  - https://github.com/glfw/glfw/issues/2519

To reproduce a null dereference on X11, my setup is as follows: Fedora, DisplayLink using two monitors. The issue occurs when I turn off one of the monitors and then turn it back on while calling glfwGetMonitorWorkarea().

### DIFF
--- a/src/x11_monitor.c
+++ b/src/x11_monitor.c
@@ -40,7 +40,7 @@
 //
 static GLFWbool modeIsGood(const XRRModeInfo* mi)
 {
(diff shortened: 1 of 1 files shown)


Gilqamesh:
Not sure if `modeIsGood` should be modified.
Alternatively could just check for non-null `mi` before call sites for full backward compatibility, and do best-effort return (report error with default return values)

PR #2889, opened 2026-08-19
Add ANGLE DirectComposition surface hint

### LINKED ISSUE #2876: Would a GLFW init/window hint exposing EGL_DIRECT_COMPOSITION_ANGLE be acceptable?
ANGLE already exposes EGL_ANGLE_direct_composition.
Enabling this attribute allows ANGLE's D3D11 backend to use
the modern DirectComposition / flip-model presentation path.

Without it, GLFW users cannot opt into this ANGLE feature
without patching GLFW locally.

From my testing it permit to get Hardware composed flip on render (Intel presentmoon) 
I got this result by patching angle directly, 

Image

Without this, i obtain a "composed flip" in presentmoon

acutally we always go in 
mDirectComposition = false 
In  https://github.com/google/angle/blob/main/src/libANGLE/renderer/d3d/d3d11/win32/NativeWindow11Win32.cpp

HRESULT NativeWindow11Win32::createSwapChain(ID3D11Device *device,
                                             IDXGIFactory *factory,
                                             DXGI_FORMAT format,
                                             UINT width,
                                             UINT height,
                                             UINT samples,
                                             IDXGISwapChain **swapChain)

Benefice : 
- Flip model swap chains
- Hardware Composed Flip
- Independent Flip (when the system allows it)

if you accept i can work, on a new hint like GLFW_ANGLE_SURFACE_DIRECT_COMPOSITION

fabiitch:
Fix #2876
Add a new ANGLE-specific window hint that enables EGL_DIRECT_COMPOSITION_ANGLE during EGL window surface creation on Windows.

When used with ANGLE on Windows, this allows ANGLE to create DirectComposition-backed swap chains.

The EGL_ANGLE_direct_composition extension is checked during EGL initialization. If the extension is not supported by the platform, the hint is safely ignored.

Default behavior remains unchanged.

### Benefits
Enabling this hint allows applications using the ANGLE backend on Windows to opt into modern presentation paths instead of standard composed blits. Specifically, it unlocks:

Hardware Composed Flip :  reducing presentation latency


BlindBison:
Anyway for me to help you get this in sooner? This would be great to have and I'd like to use it in my own Libgdx project once it gets merged in. Thanks for pushing this up mate.

